PR #11, opened 2026-04-09
fix: centralize repl lambda hot reload tracking

mattneel:


tsotchke:
Superseded by 79f1a1b — architectural REPL hot reload with PLT/GOT-style slot indirection.

Matt's cleanup (removing the dead `registerLambdaVar` calls) is included as 74e1020. The full hot reload commit on top of it goes much further: versioned function names, heap-allocated variable slots, R7RS-correct dynamic name resolution in closures, mutual exclusion between function/variable namespaces, and comprehensive test coverage for every redefinition form (function, plain var, lambda var, type changes, namespace switching, mutual recursion).

Thanks Matt — your earlier CI matrix work (PR #10) is what made cross-platform verification possible.

PR #12, opened 2026-04-09
fix(pkg): reject path-like dependency names

Gabriel-Kahen:
## Summary
- reject dependency names that contain path-like characters before loading or adding them to a manifest
- prevent `eshkol-pkg install` from treating dependency keys as raw filesystem paths outside the package cache and `eshkol_deps`
- add a regression test that reproduces the `../outside_link` escape and verifies it is rejected

## Testing
- `g++ -std=c++17 -O0 -g tools/pkg/eshkol_pkg.cpp -o /tmp/eshkol-pkg-test`
- `g++ -std=c++17 -O0 -g tests/pkg/path_validation_test.cpp -o /tmp/eshkol-pkg-path-validation-test`
- `/tmp/eshkol-pkg-path-validation-test /tmp/eshkol-pkg-test`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,6 +90,10 @@ option(ESHKOL_BUILD_INTEGRATION_TESTS "Build integration tests" OFF)
 option(ESHKOL_ENABLE_ASAN "Enable Address Sanitizer" OFF)
 option(ESHKOL_ENABLE_UBSAN "Enable Undefined Behavior Sanitizer" OFF)
 
+if(ESHKOL_BUILD_TESTS)
+    enable_testing()
+endif()
+
 # XLA Backend Support (optional, default OFF)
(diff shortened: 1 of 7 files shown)


tsotchke:
Merged via cherry-pick as c6a628b, then hardened in 3896128.

Gabriel's original validation logic is intact. The follow-up commit adds:
- 64-character length cap (matches cargo)
- Explicit rejection of "." and ".."
- Case-insensitive rejection of Windows reserved device names (CON, NUL, AUX, PRN, COM1-9, LPT1-9) — checked against the stem before any extension dot, so "CON.txt" is also rejected
- Shared subprocess launcher (`inc/eshkol/pkg/subprocess.h`) so the production code and both regression tests use one implementation

The path validation test passes on the clean CI-mimic build. Thank you Gabriel — solid security fix.

PR #13, opened 2026-04-09
fix(pkg): stop shelling out through std::system

Gabriel-Kahen:
## Summary
- replace `std::system()`-based package manager invocations with direct child-process execution on both POSIX and Windows
- pass compiler, runner, and `git clone` arguments as structured argv vectors so manifest data is treated literally instead of by a shell
- add a regression test that proves package metadata containing shell metacharacters no longer executes during `eshkol-pkg build`

## Testing
- `g++ -std=c++17 -O0 -g tools/pkg/eshkol_pkg.cpp -o /tmp/eshkol-pkg-test`
- `g++ -std=c++17 -O0 -g tests/pkg/command_injection_test.cpp -o /tmp/eshkol-pkg-command-injection-test`
- `/tmp/eshkol-pkg-command-injection-test /tmp/eshkol-pkg-test`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,6 +90,10 @@ option(ESHKOL_BUILD_INTEGRATION_TESTS "Build integration tests" OFF)
 option(ESHKOL_ENABLE_ASAN "Enable Address Sanitizer" OFF)
 option(ESHKOL_ENABLE_UBSAN "Enable Undefined Behavior Sanitizer" OFF)
 
+if(ESHKOL_BUILD_TESTS)
+    enable_testing()
+endif()
+
 # XLA Backend Support (optional, default OFF)
 # When enabled, provides accelerated tensor operations via XLA/StableHLO
 # Requires MLIR and StableHLO libraries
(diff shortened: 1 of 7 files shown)


tsotchke:
Merged via cherry-pick as f3249cc, then hardened in 3896128.

Gabriel's fork+execvp / CreateProcessW architecture is kept as-is with one critical Windows fix applied during the merge: `CreateProcessW` is called with `lpApplicationName = nullptr` so it searches %PATH% for tools like `git.exe` (an explicit application name disables PATH search per MSDN, which would have broken `eshkol-pkg install` on Windows).

The follow-up commit (3896128) also adds:
- Shared `inc/eshkol/pkg/subprocess.h` header — single implementation used by production + both regression tests, with proper msvcrt argv quoting (handles trailing-backslash-before-quote edge case)
- `is_safe_git_url()` defense-in-depth allowlist (https:// and git@host:path only) applied before `git clone`, since git itself supports URL schemes that can execute commands on clone
- Dropped the over-defensive `__has_include(<filesystem>)` / `<experimental/filesystem>` shim — eshkol-pkg already requires C++17

Both regression tests pass on the clean CI-mimic build. Thank you Gabriel — excellent security work.

PR #15, opened 2026-04-10
fix(server): validate Content-Length headers

Gabriel-Kahen:
## Summary
- validate `Content-Length` header values before converting them in `eshkol-server`
- return `400 Bad Request` for malformed or overflowing values instead of letting `std::stoul` terminate the process
- add a focused parser regression test plus a web test that checks malformed raw requests are rejected without crashing the server

## Testing
- `g++ -std=c++17 -Iinc tests/web/content_length_parse_test.cpp -o /tmp/content_length_parse_test && /tmp/content_length_parse_test`
- `bash -n scripts/run_web_tests.sh`


tsotchke:
Merged via cherry-pick as c813f4b, then hardened in 8d89930 with 9 additional vulnerability fixes. See commit message for full details. Thank you Gabriel — another strong security contribution.

PR #18, opened 2026-04-13
fix(repl): load stdlib assoc in eval fallback

Gabriel-Kahen:
## Summary
- prevent nested `core.*` requires from re-entering the precompiled stdlib path while `stdlib.esk` is being source-loaded in eval mode
- add an end-to-end regression test for `eshkol-run -e` that exercises both precompiled stdlib loading and source fallback for `assoc`
- keep `--no-stdlib` behavior unchanged as a negative control in the new test

## Testing
- cmake -S . -B build
- cmake --build build --target assoc_eval_test stdlib
- ctest --test-dir build --output-on-failure -R assoc_eval_test

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1253,6 +1253,19 @@ if(ESHKOL_BUILD_TESTS AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/tests/web/content_
     add_test(NAME content_length_parse_test COMMAND content_length_parse_test)
 endif()
 
+if(ESHKOL_BUILD_TESTS AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/tests/repl/assoc_eval_test.cpp")
+    add_executable(assoc_eval_test tests/repl/assoc_eval_test.cpp)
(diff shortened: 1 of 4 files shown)


Gabriel-Kahen:
Any idea why windows-x64-lite is hanging?

tsotchke:
The windows-x64-lite hang is not related to your PR — it's a pre-existing issue in the test runner. `Invoke-ProcessCapture` in `run_all_tests.ps1` had `TimeoutSec=0` (no timeout) as default, so any test that hangs on a specific platform blocks the entire CI job forever. One of the .esk tests is hanging on Windows x64 (but not arm64).

Fix committed to master: default timeout changed to 120 seconds per test. Your PR changes are correct — tested and passing locally on macOS, and 15/16 CI lanes pass. The only failures are 2 VM-specific tests (`vm_kb_tensor_test.esk`, `vm_system_test.esk`) that test VM-only builtins without LLVM codegen wiring yet.

Nice work on the re-entry guard — the save/restore pattern for `loading_stdlib_from_source_` is exactly the right approach. The deferred private symbol registration is also clean.

PR #19, opened 2026-04-14
feat(core): add model and tensor checkpoint serialization

Gabriel-Kahen:
## Summary
- add runtime/backend tensor and model checkpoint serialization with the `ESKM` binary format, version checks, CRC32 validation, and arena-backed tensor loading
- expose `model-save`, `model-load`, `tensor-save`, and `tensor-load` through the bytecode VM (`800-803`) and LLVM/JIT codegen so the feature works across both execution paths
- add round-trip and corruption tests for empty tensors, 0-d scalars, large tensors, and multi-tensor checkpoints, and document the new builtins plus updated native ID ranges

## Testing
- cmake -S . -B build -DLLVM_CONFIG_EXECUTABLE=/opt/homebrew/opt/llvm@21/bin/llvm-config
- cmake --build build --target model_io_test eshkol-run eshkol-pkg-path-validation-test eshkol-pkg-command-injection-test content_length_parse_test eshkol-pkg -- -j4
- ctest --test-dir build --output-on-failure
- ./build/eshkol-run -e '(begin (define entries (list (cons "W1" #(1 2)) (cons "b1" #(3)))) (display (model-save "/tmp/eshkol-pr-check.em" entries)) (newline) (display (model-load "/tmp/eshkol-pr-check.em")) (newline))'


tsotchke:
## Review

The model serialization core is solid — ESKM binary format with CRC32, arena-backed loading, both VM and LLVM paths, thorough tests. Good work.

**Two issues to fix before merge:**

### 1. Reverting PR #18 fixes (BLOCKER)

`lib/repl/repl_jit.cpp` and `lib/repl/repl_jit.h` revert the re-entry guard (`loading_stdlib_from_source_`) and deferred private symbol registration from PR #18. These are production fixes for the stdlib source-fallback path — reverting them reintroduces the nested require bug.

**Fix:** Rebase on current master (which includes PR #18) and drop the repl_jit revert hunks. Keep only the `#include <eshkol/model_io.h>` addition.

### 2. Deleting assoc_eval_test.cpp

`tests/repl/assoc_eval_test.cpp` is deleted and its CMake block replaced. This test is already merged to master from PR #18 and should remain. The model_io_test should be added alongside it, not replacing it.

**Fix:** Keep the assoc_eval_test CMake block and add the model_io_test block after it. Don't delete the test file.

### Otherwise

The implementation itself looks correct:
- ESKM format with magic + version + CRC32 — production-grade
- Uses all-pointer calling convention for LLVM codegen (ABI-safe)
- Native IDs 800-803 with geometric manifold range shifted to 804+
- Tests cover round-trip, corruption detection, edge cases
- Documentation updated

Fix the two revert issues and this is ready to merge.

tsotchke:
**Clarification on the repl_jit revert:** Your branch appears to have been based before PR #18 merged. The `loading_stdlib_from_source_` guard and deferred private symbol registration are now on master — please rebase on current master (`git fetch origin && git rebase origin/master`) and your repl_jit.h/cpp changes should resolve to just adding the `#include <eshkol/model_io.h>` line. Same for the assoc_eval_test — it's on master now, so after rebase the deletion should disappear from your diff.

tsotchke (approved):
LGTM. Model serialization with ESKM binary format, CRC32 validation, arena-backed tensor loading. 15/16 CI green (windows-x64-lite REPL failure is pre-existing). Covers both VM (native IDs 800-803) and LLVM codegen paths. Clean implementation.

PR #75, opened 2026-06-26
fix(ad): true nested/higher-order AD via perturbation tagging (ESH-0067)

tsotchke:
Replaces the **central-difference numerical differentiation** that all higher-order/nested AD secretly used — the source of perturbation confusion + the silent finite-difference hessian (found in the deep edge-case campaign).

**Mechanism:** a 4-component jet `{primal, e1, e2, e1·e2}` with perturbation tagging by nesting depth. Each derivative level (and each hessian diff-var) gets a distinct perturbation e1/e2 — the standard cure for perturbation confusion. 2 simultaneous perturbations ⇒ **2 levels of forward-over-forward nesting are exact**; deeper nesting is flagged explicitly (no silent fallback). hessian/jacobian/laplacian now use exact tagged AD.

**Before → after (verified):**
| expr | before | after |
|---|---|---|
| `d/dx(d/dy x·y)` @5 | 0 | **1** |
| `d²/dx²(x³)` @2 | 0 | **12** |
| `d/dx(x·d/dy(y²)@x)` @1 | garbage 5e-314 | **4** |
| hessian off-diagonal (separable) | 4.44e-8 | **exactly 0** |

Single-level AD unchanged. **Regression:** `tests/ad/nested_ad_test` 10/10, `stateful_tape` 22/22, `free_energy_ad` 2/2, `ad_input2` 4/4. (Started by a worktree agent cut off by a session limit; verified + completed + tested here.)

### DIFF
--- a/inc/eshkol/backend/autodiff_codegen.h
+++ b/inc/eshkol/backend/autodiff_codegen.h
@@ -90,6 +90,15 @@ public:
     // stays in the current block.
     llvm::Value* safeUnpackDualFromTagged(llvm::Value* tagged);
 
+    // Nested forward-mode AD (perturbation tagging by nesting depth).
+    // seedDerivativeInput: build the dual argument for a `derivative` at the
+    //   given nesting depth, preserving any perturbation the point already
+    //   carries (so an inner derivative does not strip the outer one).
+    // extractDerivativeResult: pull the derivative w.r.t. this level's slot —
+    //   a scalar at depth 0, a dual slice when nested.
+    llvm::Value* seedDerivativeInput(llvm::Value* point_tagged, int depth);
+    llvm::Value* extractDerivativeResult(llvm::Value* result_tagged, int depth);
+
     /**
      * Add two dual numbers: (a + b, a' + b')
      * @param left Left dual number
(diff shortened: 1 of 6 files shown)


tsotchke:
**Hold — do not merge yet.** Deep bisection (full writeup in `AD_NESTED_REUSE_ROOTCAUSE_2026-06-26.md`) found this branch's 4-jet rework has a 2nd-order correctness bug: binding a *computed* dual to a local/let/loop variable and reading it more than once corrupts the second-order (cross-perturbation) components. Minimal repro: `(gradient (lambda(y)(gradient (lambda(x)(let ((t (* x x))) (* t t))) y)) 7.0)` returns 196, must be 588; `(* (+ x 0.0)(+ x 0.0))` nested returns 0, must be 2. First-order AD is unaffected. This is the same defect behind the Noesis named-let blow-up (loop var reused per iteration → 3.48e36). Root: the `*` path builds reverse-mode tape nodes (arena_allocate_ad_node/tape_add_node) that alias the forward dual heap structs during a pure-forward `gradient`; this is also the per-call tape leak (20k-iter 2nd-order gradient SIGILLs). Fix in progress on ESH-0070.

PR #77, opened 2026-06-26
fix(jit): portable sincos shim — unbreak windows-arm64 (MSVC) build

tsotchke:
## Summary

- fixes the original windows-arm64 MSVC compile failure by replacing the nonstandard CRT `::sincos` dependency with an Eshkol-owned portable shim
- handles the second windows-arm64 AOT layer by avoiding AArch64-COFF Large code model on Windows, where LLVM 21 emits mismatched SEH unwind prologues
- keeps windows-arm64 in the small code model and shrinks generated AOT executables with COFF function/data sections plus `/OPT:REF,/OPT:ICF`
- fixes the follow-up compile regression from the layer-3 patch: `eb20ba7b` now derives `executable_triple` in the executable link path before applying windows-arm64 link flags
- bounds the hosted `windows-arm64-lite` smoke surface in `dfbb7885` / `28f98976` to representative portable tests; the previous broad run reached the end of the suite and only failed probes requiring unlinked lite-path surfaces (`csv_comprehensive_test.esk`, AD tape/consciousness, and memory crypto/random)
- fixes the hosted lite typesystem slice in `3a89fece` to use the existing contract-aware `Invoke-TypesystemSuite`, so tests with `EXPECT-MODE` / `EXPECT-STDERR` are checked as type-system probes instead of being run as ordinary executables

## Verification

- Previous run `28274609083`: all non-windows-arm64 lanes passed; windows-arm64 failed compiling `lib/backend/llvm_codegen.cpp` because `target_triple` was undeclared in the executable link path.
- Run `28340423219` on `eb20ba7b`: all non-Windows lanes passed; windows-arm64-xla and windows-arm64-cuda passed; windows-arm64-lite configure/build passed and then failed only in Windows-lite smoke tests with four known non-lite probes.
- `dfbb7885` excluded the failing non-lite probes; `28f98976` further tightened `windows-lite` to the bounded hosted-runner smoke suite documented for Windows validation. Run `28344145912` on `28f98976`: every lane passed except `windows-arm64-lite`; the lite log showed 41/44 passing with the only failures in `typesystem` because the generic runner ignored existing `EXPECT-*` contracts.
- `3a89fece` switches the hosted lite typesystem subset to `Invoke-TypesystemSuite` with include filtering.
- Current CI is rerunning on `3a89fece`.
- Local static check: `git diff --check`

CI is the verifier for the MSVC/windows-arm64 branch; there is no local arm64-Windows machine in this workspace.

### DIFF
--- a/lib/backend/llvm_codegen.cpp
+++ b/lib/backend/llvm_codegen.cpp
@@ -35689,26 +35689,34 @@ int eshkol_compile_llvm_ir_to_object(LLVMModuleRef module_ref, const char* filen
         //   - x86_64 / others: leave at default (nullopt → small/medium
         //     per ABI). Range overflow doesn't bite there.
         std::optional<llvm::CodeModel::Model> code_model;
-#if LLVM_VERSION_MAJOR >= 21
-        // LLVM 21's AArch64 COFF backend supports the Large code model
-        // (movk/movz + COFF base relocations for absolute 64-bit addresses).
-        // windows-arm64 needs it for the SAME reason Linux/macOS ARM64 do:
-        // libeshkol-static + user code exceeds the ±128 MiB BL/ADRP reach, and
-        // on COFF the small model makes lld-link insert range-extension thunks
-        // that fail to converge ("adding thunks hasn't converged after 10
-        // passes"). Large eliminates the range-limited branches entirely.
-        const bool use_large_aarch64_code_model = object_triple.getArch() == Triple::aarch64;
-#else
-        // Pre-21 toolchains: keep Windows on the default model — the AArch64
-        // COFF Large code model is not reliably supported there.
-        const bool use_large_aarch64_code_model = object_triple.getArch() == Triple::aarch64 &&
-                                                  !object_triple.isOSWindows();
-#endif
+        // AArch64 Linux/macOS use the Large code model to escape the ±128 MiB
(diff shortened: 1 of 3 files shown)


tsotchke:
**Progress: 2 of 3 windows-arm64 layers fixed.** This branch now has (1) the portable `sincos` shim [fixes the compile error] and (2) the AArch64 Large code model for aarch64-windows on LLVM≥21 [fixes `lld-link: adding thunks hasn't converged after 10 passes`]. CI confirms the thunk error is gone.

**Layer 3 now exposed:** object emission fails with `error: Incorrect size for <fn> prologue: 28 bytes of instructions in range, but .seh directives corresponding to 24 bytes` — the Large code model's longer prologue desyncs the SEH (.seh_*) unwind directives (an LLVM AArch64-COFF backend interaction). Tracked as ESH-0073.

Next candidate (likely cleaner): instead of the Large code model, shrink the AOT binary with dead-strip (`/OPT:REF,/OPT:ICF` + function-sections/`--gc-sections`) so the ~58MB stdlib prunes below the thunk-convergence threshold — avoiding the SEH issue entirely. Keeping #77 open as the windows-arm64 unbreak track. Note: windows-arm64 is a non-required, long-baseline-red lane; the 10 baseline-green lanes (macOS/linux x64+arm64) are unaffected.

PR #168, opened 2026-07-06
feat(ad): P11 tower-based user numerics -- Taylor ODE/root/inversion (ESH-0196)

tsotchke:
## Summary

Implements Phase 11 (ESH-0196) of the Taylor-tower AD design (`docs/design/AD_TAYLOR_TOWER.md` sec 13): exposes the univariate Taylor tower (`taylor`/`derivative-n`, Phases P1/P2/P6) as general-purpose numerical methods rather than an internal-only AD representation. Pure new-file Scheme module, zero codegen/runtime changes, following `core.ad.guw`'s (P4) precedent of orchestrating entirely on top of the existing `taylor` primitive.

**Branch base note:** this branch is stacked on `feat/ad-taylor-p6-exact-coefficients` (#163) as instructed, but that branch does not include P4's `lib/core/ad/guw.esk` (PR #162, also stacked on P2, a sibling of P6 rather than an ancestor). P11's ODE/root/inversion methods don't need GUW's multivariate machinery directly, but the task's stated substrate and regression gate ("guw 35/35") assume it's present, so **this branch merges `feat/ad-taylor-p4-guw-multivariate` in** (clean merge, no conflicts). That means this diff currently shows P4's files too; it will shrink to just the P11 files once #162 merges into the P6/P11 stack ahead of this one.

### API (`lib/core/ad/taylor_numerics.esk`, `require core.ad.taylor_numerics`)

- `(taylor-eval coeffs dt)` -- Horner evaluation of a `taylor`-style coefficient list, shared by everything below.
- `(taylor-ode-solve f y0 t0 t1 k n)` -- classic recursive/automatic Taylor-series integrator for `y'=f(t,y)`. Builds the solution's order-`k` coefficients one order at a time (a fresh `taylor` call of growing order per step, exploiting that formal power-series composition is lower-triangular), then advances `n` fixed steps via `taylor-eval`.
- `(taylor-root f x0 k)` -- Householder-family root refinement via series reversion of f's own tower: `k=1` is Newton (quadratic), `k=2` is Halley (cubic), general `k` gives order-`(k+1)` convergence.
- `(taylor-inverse-series f x0 k)` -- local functional inverse via Lagrange inversion / series reversion: given f analytic at `x0` with `f'(x0)!=0`, returns the Taylor coefficients of `f^-1` around `f(x0)`. Exact (P6) coefficients in, exact rationals out.

**Argument style deviation:** the design doc's API sketch uses `#:order`/`#:steps` keyword args. All arguments here are plain **positional** instead -- keyword-arg formals turn out to be unusable in any file that also contains a dotted `(require ...)` (a newly discovered, pre-existing compiler bug, filed as ESH-0220, not fixed here).

### Error-vs-order table (`y'=y, y(0)=1 -> e`, fixed steps=4)

| order | error |
|---|---|
| 2 | 2.34e-2 |
| 4 | 7.19e-5 |
| 6 | 1.06e-7 |
| 8 | 9.13e-11 |
| 10 | 5.15e-14 |
| 12 | 4.44e-16 (double-precision floor) |
| 14 | 4.44e-16 |

### Exact-inversion case (Phase P6 compositional)

`f(x)=x+x^2` at exact integer `x0=0`: reversion coefficients come back **bit-exact** and match the hand-derived alternating-sign Catalan series `0, 1, -1, 2, -5, 14, -42, 132, -429` (verified independently via undetermined coefficients, not by re-deriving the library's own algorithm). A second exact case at rational `x0=1/2` matches the hand-derived `sqrt(1+w)-1` binomial series `1/2, 1/2, -1/8, 1/16, -5/128, 7/256, -21/1024`, also bit-exact.

### Root finding

`x^2-2` from `x0=1`: order 1 (Newton), 2 (Halley), 4 all converge to `sqrt(2)` to `<1e-14`. An independent hand-written classical Halley-formula cross-check (different code path, 4 manual iterations) shows textbook cubic convergence: errors `1.41e-2, 3.64e-7, 2.22e-16, 0`. A transcendental root `cos(x)=x` (Dottie number) also converges.

### Bugs discovered and worked around (not fixed -- separately filed)

- **ESH-0220**: keyword-arg formals (`#:name`) fail to resolve (`Unknown function: __keyword-arg`) in *any* file that also contains a dotted `(require ...)` statement, regardless of order. Root cause suspected in stdlib-symbol visibility after module resolution; same underlying issue as the pre-existing "no `list-ref` in core modules" workaround in `guw_multivariate_test.esk`.
- **ESH-0221** (higher severity): a `taylor` call whose lambda argument closes over a tail-recursive (TCO'd/PHI-loop) function's own loop-carried parameter reads a stale/corrupted value once the function actually recurses -- silently wrong for a captured scalar, a crash or segfault for a captured list. Worked around in `tn-ode-coeffs-loop` by rebinding the loop-carried parameters into fresh locals immediately before constructing the closure, verified correct with a minimal, independently reproduced repro.

## Test plan

- [x] `tests/ad/taylor_numerics_test.esk` (41 checks): PASS under `-r` (JIT) and AOT, exit 0. Registered as CTest `taylor_numerics_runtime_smoke`.
- [x] Regressions (both `-r` and AOT): `taylor_tower_test.esk` 52/52, `taylor_tower_mono_test.esk` 441/441, `exact_taylor_test.esk` 68/68, `guw_multivariate_test.esk` 35/35.
- [x] `scripts/run_ad_oracle.sh`: 56/56.
- [x] `scripts/run_sicp_smoke.sh`: 88/88.
- [x] `cmake --build build --target eshkol-run stdlib -j8` clean.


tsotchke:
Rebased: this branch is now a single clean commit stacked directly on `feat/ad-taylor-p6-exact-coefficients` (PR #163), not on master, because #163 is not yet merged and P11 depends on its P6 content. **Merge after #163.** Once #163 lands on master, this will fast-forward cleanly. Gate: `tests/ad/taylor_numerics_test.esk` 41/41 under both -r and AOT.

PR #238, opened 2026-07-10
perf(codegen): optimize generated artifacts by default (O2)

tsotchke:
## Summary
The LLVM backend defaulted generated code to O0, so a CMake Release build of the compiler still emitted unoptimized Eshkol binaries -- `eshkol-run file.esk -o bin` produced O0 output regardless of how the compiler was built (ADR 0007, "Generated code has an independent optimization plane").

This defaults the artifact-producing paths (`-o` AOT binary, `-c` object, `--shared-lib`) to O2. Ephemeral/interactive paths (plain run, `-r`, `-e`, REPL) and `-g` debug builds stay at O0 for fast turnaround. An explicit `-O<n>` always wins (including `-O0` to opt out, `-O3` for a performance artifact).

## Measured speedup (AOT, best of 3, O0 vs new O2 default)
| program | O0 | O2 | speedup |
|---|---|---|---|
| scalar numeric loop | 2.12s | 0.94s | 2.26x |
| pure-Scheme matmul | 2.45s | 0.96s | 2.55x |

## Correctness
- AD gradients are bit-identical between O0 and O2 (the default pipeline enables no fast-math/reassociation).
- Autodiff suite: 54/54.
- Full `run_all_tests.sh`: green (one example, `eagle_train.esk`, flaked once under load and passes on isolated re-run; it uses the O0 plain-run path, unchanged by this PR).

## Tests / PGO Phase-0 groundwork
- Adds `scripts/run_codegen_optlevel_tests.sh` (wired into `run_all_tests.sh`): asserts the default opt level per path and that explicit `-O` wins -- a regression guard for the sleeper bug.
- Checks in a first PGO training corpus under `bench/pgo_corpus/`.
- ADR 0007 Phase-0 status note records what landed and what is deferred (JSON measurement runner + full PGO train/merge/use workflow).

### DIFF
--- a/bench/pgo_corpus/arithmetic.esk
+++ b/bench/pgo_corpus/arithmetic.esk
@@ -1,0 +1,39 @@
+;; PGO training corpus — arithmetic / numeric hot paths.
+;;
+;; Exercises the integer + floating-point arithmetic codegen, recursion,
+;; tail calls, and the bignum dispatch path. Deterministic and offline:
+;; the program prints checksums so the workload driver can confirm the
+;; instrumented binary actually executed (not just linked).
+
+;; Tight integer loop — TCO + add/mul/compare dispatch.
+(define (sum-to n acc)
+  (if (= n 0) acc (sum-to (- n 1) (+ acc n))))
+
+;; Floating-point accumulation — packDouble / unpackDouble round-trips.
+(define (fsum i n acc)
+  (if (>= i n) acc (fsum (+ i 1) n (+ acc (* (exact->inexact i) 1.5)))))
+
+;; Recursive factorial — drives the bignum promotion path past 2^63.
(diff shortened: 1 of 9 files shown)


tsotchke:
Superseded by #237 (already merged). The O2-default codegen change, opt-level assertion test, PGO corpus, and ADR 0007 Phase-0 note in this PR are byte-identical to what landed on master via #237 (verified: git diff origin/master against this branch shows the PGO files SAME; the only delta is site/press updates this branch predates). This PR and #237 were two parallel efforts on the same fix. Closing to avoid a duplicate merge; the work is on master and independently verified (autodiff 54/54, bit-identical gradients O0 vs O2, ~2.3-2.5x AOT speedup).

PR #263, opened 2026-07-11
fix(memory): deep-walk bignum rationals in region evacuation (heap-use-after-free)

tsotchke:
A latent heap-use-after-free in region evacuation, found by an ASan audit of every `HEAP_SUBTYPE_*` against `evac_kind_for`'s coverage.

## The bug
`HEAP_SUBTYPE_RATIONAL` was classified `EVAC_LEAF`. That is correct for the int64 fast path, but wrong for the bignum path (`is_big == 1`, used when a GCD-reduced exact rational's numerator/denominator overflows int64): `big_num` / `big_den` are raw `eshkol_bignum_t*` pointers to independently arena-allocated objects. As a leaf, they were left unwalked, so after the owning region popped, a rational that survived the region carried dangling pointers into the freed arena. This is the same residual-subtype class as ESH-0214d/e (#226/#230), in a subtype neither existing memory test constructed.

## Evidence (ASan, before the fix)
```
ERROR: AddressSanitizer: heap-use-after-free ... READ of size 4
  #0 eshkol_bignum_compare bignum.cpp:690
  #1 eshkol_rational_equal rational.cpp:456
  #2 eshkol_deep_equal runtime_deep_equal.cpp:198
freed by:  arena_destroy runtime_arena_core.cpp:186  <- region_destroy runtime_regions.cpp:435
allocated: arena_create  runtime_arena_core.cpp:94   <- region_create  runtime_regions.cpp:370
```

## Fix
`lib/core/runtime_regions.cpp`: add `EVAC_RATIONAL` to `evac_kind_for` and a worklist case that deep-copies `big_num` / `big_den` via the existing `evac_object_ptr` raw-pointer path (the same mechanism used for a knowledge base's `facts[]`) whenever `is_big` is set.

## Verification
- New fixture `tests/memory/region_evac_rational_bignum_test.esk`: before fix, 3/3 deterministic ASan crash and 5/5 poison-build failures; after fix, clean.
- Full memory suite stable at 15/15 across 3 runs (14 prior + the new fixture) on a normal build; ASan+poison runs of the fixture clean.

Found during a broader region-evacuation audit; an additional gap in `make-parameter`'s dynamic value stack (`HEAP_SUBTYPE_PARAMETER`) was identified and is tracked separately (it needs write-barrier call-site changes, not an `evac_kind_for` entry).

### DIFF
--- a/lib/core/runtime_regions.cpp
+++ b/lib/core/runtime_regions.cpp
@@ -15,6 +15,7 @@
 #include "../../inc/eshkol/core/logic.h"       // eshkol_substitution_t / _fact_t / _knowledge_base_t
 #include "../../inc/eshkol/core/inference.h"   // eshkol_factor_graph_t / _factor_t
 #include "../../inc/eshkol/core/workspace.h"   // eshkol_workspace_t / _workspace_module_t
+#include "../../inc/eshkol/core/rational.h"    // eshkol_rational_t (bignum-backed exact rationals)
 
 #include <cstring>
 #include <cstdlib>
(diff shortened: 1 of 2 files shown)


tsotchke:
Included in #265 (the rational-bignum evac UAF fix is cherry-picked into the green-CI integration).

PR #264, opened 2026-07-11
fix(codegen): pin library-mode stdlib exports against O2 GlobalDCE (fixes red lite/asan lanes)

tsotchke:
Fixes the RED `lite` and `asan-ubsan` CI lanes (all platforms). Since PR #256 they have failed to link AOT stdlib test binaries with undefined references to stdlib functions (`atomic-write-file`, `csv-parse-line`, `fold-left`, `string-trim`, ...).

## Root cause
PR #256 gated the homoiconic display registry behind `homoiconicRegistryEnabled() = !freestanding && !wasm` so that standalone wasm objects could dead-strip unused stdlib. But the native precompiled `stdlib.o` is *also* a freestanding object, so it lost the registry's `ptrtoint` pinning as well. Its `LinkOnceODR` exports that nothing else in the translation unit calls were then deleted by the standard -O2 GlobalDCE before the object was written, so any downstream AOT program using one failed to link. The `xla`/`cuda` lanes stayed green only because they don't run the stdlib AOT test suite.

## Fix
Restore the pre-#256 contract for a precompiled library object — every public definition is part of its exported surface regardless of in-module callers — by marking library-mode weak/linkonce **function** definitions `@llvm.compiler.used`. That is LLVM's documented mechanism for "treat as reachable" without changing linkage, so duplicate weak definitions across separately linked objects still merge. Only functions are pinned, never data globals (the old registry only address-took functions; force-keeping weak globals corrupts the freestanding object's memory — an ASan-confirmed dead-end during development). Native non-library codegen and the `--wasm` path are unaffected: wasm never sets `library_mode`, so its Internalize+GlobalDCE dead-strip (the #256 win, guarded by `scripts/check_wasm_size.py`) is preserved.

## Verification (lite config, Release, XLA/GPU off)
- stdlib AOT suite: **18/18** (was failing to link). `nm build/stdlib.o` confirms `fold-left` / `atomic-write-file` / `csv-parse-line` / `string-trim` are now defined (`T`).
- memory suite: **14/14 stable across 20 consecutive runs** (no crashes, deterministic).
- The `--wasm` dead-strip path is untouched (guarded separately by `library_mode`).

Note: this audit also surfaced two pre-existing region-evacuation memory bugs (a bignum-rational heap-use-after-free, fixed in a separate PR; and a `make-parameter` write-barrier gap, tracked as follow-up) — found because enlarging `stdlib.o` perturbed memory layout. They are independent of this linkage fix.

### DIFF
--- a/lib/backend/llvm_codegen.cpp
+++ b/lib/backend/llvm_codegen.cpp
@@ -3357,6 +3357,8 @@ public:
 
             pruneUnusedFreestandingDeclarations();
 
+            preserveLibraryModeDefinitions();
+
             // Coerce mismatched integer types before verification so wasm32
             // does not trip on i32 size_t / i64 byte-count mismatches in
             // generated calls or binary operators.  On native this is usually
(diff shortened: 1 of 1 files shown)


tsotchke:
Superseded by #265: the tail-call registration root fix (llvm_codegen.cpp) is the true cause of the stdlib DCE stripping; this blanket compiler.used pin is no longer needed.

PR #295, opened 2026-07-17
build: guard DOWNLOAD_EXTRACT_TIMESTAMP for CMake < 3.24

LJGz:
DOWNLOAD_EXTRACT_TIMESTAMP was added in CMake 3.24. On older CMake (e.g. 3.22 still common on Ubuntu 22.04 / Linux Mint 21) the unknown keyword corrupts the URL_HASH argument and aborts configure.

Guard the option so the project continues to build on CMake 3.22+. Behaviour on CMake ≥ 3.24 is unchanged.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3750,10 +3750,15 @@ if(ESHKOL_BUILD_AGENT_FFI)
             message(FATAL_ERROR "Pinned PCRE2 did not define pcre2-8-static")
         endif()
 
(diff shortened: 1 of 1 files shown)


tsotchke:
Merged — thank you. Verified independently before landing: a fresh Ubuntu 22.04 container (CMake 3.22) reproduces the configure failure on master exactly as described, and this branch configures clean; the guard also passed the full 14-lane CI matrix. This unblocks the default build on Mint 21.x and Ubuntu 22.04, which our new scripts/build-linux.sh targets, so it was well timed.

Issue #296, opened 2026-07-17
AD: custom-VJP gradient (e.g. vqe-energy) silently returns zeros inside a nested define


RichardHoekstra:
## Summary

`gradient` through a function with a custom vector-Jacobian-product (e.g. `vqe-energy` from `agent.quantum`) **silently returns a zero vector** when the `gradient` call sits inside a nested `define`d helper, but returns the correct gradient when the same call is written in the enclosing `lambda`/scope.

It fails **silently** (zeros, no error), which is the dangerous part: a perfectly ordinary training loop written with helper functions compiles, runs, and simply never learns — the energy/loss stays frozen with no diagnostic.

Pure-Eshkol-function gradients are **not** affected in the same nested scope, so this is specific to the custom-VJP / external-adjoint path, not nested-scope AD in general.

## Environment

- Eshkol Compiler **v1.3.3-evolve** (commit `831894f`)
- Built with `-DESHKOL_QUANTUM_ENABLED=ON`, LLVM 21.1.8, Ubuntu 22.04
- Run via `eshkol-run -r`

## Minimal repro

```scheme
(require agent.quantum)

;; A pure Eshkol function, differentiated inside a nested define (works fine).
(define (pure-sq v) (let ((a (vector-ref v 0))) (* a a)))
(define (nested-pure-grad v) (gradient pure-sq v))

(with-hamiltonian (make-h2-hamiltonian 0.7414)
  (lambda (ham)
    (define p (vector 0.1 0.2 0.3 0.4))

    ;; (A) custom-VJP gradient in the enclosing lambda scope
    (display "A. vqe grad, enclosing scope : ")
    (display (gradient (lambda (q) (vqe-energy ham q)) p)) (newline)

    ;; (B) the SAME call inside a nested define
    (define (nested-vqe-grad v) (gradient (lambda (q) (vqe-energy ham q)) v))
    (display "B. vqe grad, nested define   : ")
    (display (nested-vqe-grad p)) (newline)

    ;; (C) control: a PURE-function gradient inside a nested define
    (display "C. pure grad, nested define  : ")
    (display (nested-pure-grad p)) (newline)))
```

## Actual output

```
A. vqe grad, enclosing scope : #(0.254119 -0.00358856 0.113678 -2.11092e-18)
B. vqe grad, nested define   : #(0 0 0 0)
C. pure grad, nested define  : #(0.2 0 0 0)
```

[...]

PR #306, opened 2026-07-17
perf(dispatch): self-calibrate the BLAS peak from real GEMMs

RichardHoekstra:
## Summary

The matmul dispatcher's cost model uses a compiled-in `blas_peak_gflops = 1100`
constant to choose BLAS vs SIMD vs GPU. That number is an Apple M-series
measurement; on other silicon it is simply wrong (a GB10's cBLAS ceiling is
~405 GFLOP/s), so every backend-vs-backend estimate in `estimate_matmul_cost` is
skewed. This wires up the runtime calibration the struct *already declares but
never performs*: time each real `cblas_dgemm` and use the measured peak.

## Motivation

`CostModelParams` already states the intent, and even carries the state fields —
they are just never written:

```c
// Cost model parameters are calibrated at runtime based on actual
// measurements, adapting to the specific hardware.
...
bool calibrated  = false;   // <- never set
int  sample_count = 0;      // <- never set
```

So the model always used the hardcoded 1100. This PR honours that existing
design rather than inventing a new one.

The update rule matters. Achieved FLOP/s is a lower bound on the true peak that
is *tight* when the GEMM is compute-bound, so the **high-water of achieved
GFLOP/s is the estimate, not a floor** — which lets a measured 405 correctly
*replace* an over-stated 1100. (A naïve "raise the estimate toward achieved"
scheme could only push it up, never correct a too-high default down, which is
exactly the failure here.)

## What changed

`lib/backend/blas_backend.cpp` — ~20 lines, no new API, no config:

- **`CostModelParams`**: one field, `double blas_peak_measured = 0.0`.
- **`eshkol::blas::matmul`**: time the `cblas_dgemm` with `steady_clock`, compute
  achieved GFLOP/s (`2*M*K*N / ns`), and update the high-water
  `blas_peak_measured`, set `calibrated = true`, bump `sample_count`.
- **`estimate_matmul_cost`**: use
  `p.calibrated ? p.blas_peak_measured : p.blas_peak_gflops`, so the compiled-in
  default is used only until the first real GEMM has run.

The `ESHKOL_BLAS_PEAK_GFLOPS` env override is preserved (it seeds
`blas_peak_gflops`, still the pre-calibration value).

## Validation

- Compiles and links cleanly on the aarch64 LLVM-21 toolchain; no new warnings
  from the change.
- The high-water rule was checked against measured GB10 matmul timings: the
  learned peak converges to ~405 GFLOP/s within a handful of GEMMs, whereas the
  compiled-in 1100 mispredicts large-GEMM time by ~2.7× at the ceiling. Correcting
  the peak downward is precisely what stops the model from over-preferring BLAS
  where it shouldn't.

## Notes / scope

- **Thread-safety.** The high-water update is a benign racy write of a `double`
  from `matmul` — a cost-model heuristic, not a correctness path — matching the
  existing non-atomic reads of `g_cost_model` in `estimate_matmul_cost`. Happy to
  wrap it in a `std::atomic` if you'd prefer.
- Only the BLAS peak self-calibrates here; the GPU/SIMD peaks are untouched. The
  same pattern would extend to them, but I kept this PR to the one hot constant.


tsotchke:
With the precision wave merged (#312/#314/#315/#316/#317/#318) this conflicts — could you rebase onto current master? Post-rebase this becomes stage 1 of the #311 selector plan (the calibrated cost model), so it's now on the critical path rather than a standalone nicety. Review will be prompt once it's rebased.

RichardHoekstra:
Rebased onto current master (fa8b8c7). Clean — the branch is now a single commit touching only \`lib/backend/blas_backend.cpp\` (+21/−2); the earlier "conflict" was 5 of your own PRs (#296/#300, #301, #303, #304, #305) that had merged locally at fork time and have since landed upstream, so dropping them *is* the rebase. Nothing upstream has touched `blas_backend.cpp` since, so it replants cleanly.

Also verified the calibration on real GB10 silicon (the hardware the M-series `1100` default mis-serves), driving `cblas_dgemm` through the exact patched path:

```
Before calibration: estimate(2048³).blas_ns = 15,623 µs   (assumes peak 1100)
N=256    high-water =  165.6 GFLOP/s
N=512    high-water =  222.3
N=1024   high-water =  291.6
N=2048   high-water =  370.5
N=4096   high-water =  395.2   ← real GB10 FP64 peak
After calibration:  estimate(2048³).blas_ns = 43,473 µs   (peak measured 395.2)  →  2.78x correction
```

So the compiled-in default was telling the dispatcher a GEMM is ~2.8× cheaper on BLAS than it actually is on this silicon — exactly the skew this removes. Ready for review as stage 1 of the #311 selector plan (the calibrated cost model).

tsotchke:
Heads-up: with #346/#347 progressing the selector plan, could you rebase this onto current master when you get a chance? The measured-peak calibration is still wanted as the selector's stage-1 input. CI needs a fresh approval run after rebase either way.

RichardHoekstra:
Rebased onto current master (6407886, post-#346). Clean — still the single-commit `blas_backend.cpp` change (+21/−2); nothing upstream touched that file. Ready for a fresh CI run.

tsotchke:
Integration is proceeding through #448 because this fork branch currently renders against master as 1,716 commits and 3,287 changed files. #448 ports the intended 2d138758 BLAS calibration contribution onto current master, hardens its concurrency and override semantics, and preserves Richard Hoekstra as co-author. I will close this PR as integrated only after #448 merges.

tsotchke:
Integrated on current master through #448. The original fork branch had 1,716 unrelated commits and conflicted with current master, so it was not safe to merge literally. #448 preserves the intended self-calibrating BLAS peak behavior, adds thread-safe atomic high-water tracking, preserves explicit override semantics, filters undersized GEMMs, adds focused provider/concurrency coverage, and retains Richard Hoekstra as co-author. Merged commit: 74ae5b265656d939524f1de81bc576021d5c27ec.

PR #315, opened 2026-07-18
feat(numerics): mixed-precision iterative-refinement linear solver

tsotchke:
## Summary

Adds `(linear-solve A b)`, a dense `Ax=b` solver with a full-f64 accuracy guarantee and an opportunistic mixed-precision fast path.

- Apple/Accelerate factorizes in fp32 and uses f64 iterative refinement; if it cannot certify the residual, it falls back to f64 `dgesv`.
- Non-Apple builds use direct f64 LU with partial pivoting.
- Singular, dimension-mismatch, wrong-type, and non-2D inputs raise catchable conditions instead of crashing.
- The builtin is wired through native JIT/AOT codegen and the bytecode VM, with docs, parity metadata, an ICC completion oracle, and a smoke probe.

## Verified test evidence

Verified on the exact rebased head against current `master` on an M2 Ultra. JIT, AOT, and VM each exit 0 with:

```
solve-n3 rel-residual = 0
PASS: solve-n3
solve-n4 rel-residual = 0
PASS: solve-n4
solve-n8 rel-residual = 2.24535e-16
PASS: solve-n8
identity-exact max|x-b| = 0
PASS: identity-exact
PASS: singular-raises
PASS: dim-mismatch-raises
PASS: wrong-type-raises
PASS: non-2d-raises
=== linear-solve: 8 passed, 0 failed ===
```

The regression suite computes the relative backward residual independently and requires `<= 1e-12`. The two added invalid-input cases specifically verify that a native integer operand and a non-2D tensor produce catchable errors on JIT, AOT, and VM.

ICC is indexed at the exact PR head, source drift is zero, and the ICC pre-commit gate reports zero blockers.

## Measured Apple speedup

Same-function comparison against `ESHKOL_LINSOLVE_FORCE_DGESV=1`, best of three interleaved runs:

| N | dgesv | iterative refinement | residual | speedup |
|---:|---:|---:|---:|---:|
| 2048 | 0.03393 s | 0.02821 s | 1.47e-15 | 1.20x |
| 4096 | 0.18093 s | 0.13437 s | 2.22e-15 | 1.35x |

Both paths include the row-to-column-major transpose, so the comparison is like-for-like.

## Implementation map

- Runtime core: `lib/core/linear_solve.cpp`
- Native codegen: `inc/eshkol/backend/tensor_codegen.h`, `lib/backend/tensor_linalg_codegen.cpp`, `lib/backend/llvm_codegen.cpp`
- VM: `lib/backend/eshkol_vm.c`, `lib/backend/vm_native.c`
- Regression: `tests/features/linear_solve_test.esk`
- Docs/oracle/smoke: tensor references, `.icc/completion-oracles.yaml`, and `scripts/run_icc_smoke.sh`

### DIFF
--- a/.icc/completion-oracles.yaml
+++ b/.icc/completion-oracles.yaml
@@ -172,6 +172,13 @@ oracles:
         severity: high
         label: 'AD nesting: jacobian/hessian differentiating through an inner forward-mode derivative is exact, not silent-zero (ESH-0120/0121)'
         action: ./scripts/run_icc_smoke.sh
+      - runtime_event:
+          event_kinds: [eshkol_smoke]
+          event_names: ["linear_solve_full_f64_oracle"]
+          event_values: ["PASS"]
+        severity: high
+        label: 'linear-solve: mixed-precision iterative-refinement dense solver reaches full-f64 backward residual (<=1e-12, computed independently) on well-conditioned/identity systems and raises catchably on singular/dimension-mismatch — verified on JIT, AOT, and the VM'
+        action: ./scripts/run_icc_smoke.sh
 
   # ─────────────────────────────────────────────────────────────────
   # TOTAL-LANGUAGE completion — unlike the compiler-readiness ratchet above,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   boxed) so both compute bit-identical results. Docs:
   `docs/reference/language/i128.md`; tests: `tests/types/i128_test.esk`
   (native + VM parity) and `tests/types/i128_error_test.esk`.
+- **`linear-solve` — full-f64 dense linear solver with a mixed-precision fast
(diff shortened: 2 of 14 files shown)


tsotchke:
Blocking verification finding: the native path crashes on a wrong-type matrix argument instead of raising a catchable condition.

Exact head `5eaf8c845fff8d73f66a629cca2a2e6063488c83`, repro:
```scheme
(guard (e (#t #t))
  (linear-solve 1 (make-tensor (list 1) 1.0)))
```
Run with `build-irs/eshkol-run -e ...`; it exits 139 with SIGSEGV at address `0x1`. `TensorCodegen::tensorLinearSolve` unpacks/dereferences both tagged operands as tensor pointers and reads `A` dimensions/allocates the result before `eshkol_linear_solve` can validate shapes, so the runtime error contract is bypassed.

Please add native tagged-value/type validation before any tensor dereference (and ensure shape validation precedes result allocation), then cover wrong-type plus non-2D inputs on JIT, AOT, and VM. The existing 6/6 JIT/AOT/VM behavior and VM parity audit otherwise pass independently.

PR #316, opened 2026-07-18
feat(gpu): opt-in reduced-precision fully-GPU Ozaki-II fast tier (Metal)

tsotchke:
## Summary

Adds an opt-in reduced-precision, fully GPU-resident Ozaki-II tier for Metal. Select it with `ESHKOL_SF64_KERNEL=ozaki-fast` or `ESHKOL_OZAKI_FAST=1`; default DGEMM and fixed-N=16 exact Ozaki remain unchanged.

The fast tier performs cap-limited prime-power residue splitting, MPS f32 GEMMs, and df32 fractional-CRT reconstruction on GPU. The host uploads A/B once, performs one O(N²) exponent pass, and downloads C.

## Measured M2 Ultra results

| tier | N=4096 | N=8192 | vs AMX dgemm at N=8192 |
|---|---:|---:|---:|
| fast, 10 moduli | ~1080 GF | ~1448 GF | 1.32x; fails pi/e gate |
| **fast, 11 (default)** | **~1081 GF** | **~1384 GF** | **1.26x** |
| fast, 12 moduli | ~1022 GF | ~1278 GF | 1.16x |
| exact, 16 moduli | ~140 GF | ~143 GF | 0.13x |
| AMX f64 dgemm | ~1080 GF | ~1099 GF | 1.00x |

The default 11-modulus tier's worst measured relative error across integer, fractional, pi/e, and wide-range regimes at K<=4096 is about `2.4e-8`, within its `1e-7` contract.

## Hardware gates and hardening

Verified on the exact rebased head on Atlas (M2 Ultra), for JIT and AOT:

- Fixed-N=16 correctness: literal PASS; exact and clamp summaries have zero failures.
- Adaptive policy: every `MAXREL < 1e-6`.
- Clamp: zero failures and the required warning.
- Fast tier: literal PASS at `n_mod=11`, zero failures, no fallback.
- Every exact/adaptive/clamp/fast run has exactly 16 dispatch-specific GPU markers and one accepted init marker.

The shell gates exclude initialization lines from dispatch and poison assertions. Synthetic probes prove that each gate accepts exactly 16 dispatch markers plus init, but rejects zero markers, 17 markers, and 16 markers plus poison.

There is no Darwin/runtime-presence post-run skip: missing Metal dispatch evidence fails on macOS. Non-Darwin and missing-binary preflight skips remain unchanged.

## Verification

- Fresh RelWithDebInfo Ninja Metal build: PASS.
- Real Atlas JIT correctness and fast gates: PASS.
- Exact-head AOT exact/adaptive/clamp/fast assertions: PASS.
- ICC indexed with both runtime traces; source drift zero; pre-commit zero blockers.
- Accepted hardening patch appears exactly once; rejected Darwin-skip patch is absent.

## Files

- `lib/backend/gpu/gpu_memory.mm`
- `lib/backend/gpu/metal_softfloat.h`
- `tests/gpu/ozaki_correctness_gate.sh`
- `tests/gpu/ozaki_fast_gate.sh`
- `tests/gpu/ozaki_fast_test.esk`
- `.icc/completion-oracles.yaml`
- `CHANGELOG.md`
- `docs/breakdown/GPU_ACCELERATION.md`

### DIFF
--- a/.icc/completion-oracles.yaml
+++ b/.icc/completion-oracles.yaml
@@ -2052,3 +2052,24 @@ oracles:
         severity: high
         label: INT8 tensor-core Ozaki f64 GEMM (CUDA, opt-in ESHKOL_CUDA_F64_KERNEL=ozaki-int8) recovers full f64 (normwise < 1e-9 at T=6) vs an independent CPU f64 reference across integer/fractional/pi-e/wide-magnitude regimes at K<=4096, with the T=4 fast tier correct and the out-of-range accuracy knob clamped loudly
         action: ./tests/gpu/cuda_ozaki_correctness_gate.sh
+  # Ozaki-II REDUCED-PRECISION fast tier (opt-in). ESHKOL_SF64_KERNEL=
+  # ozaki-fast selects a fully-GPU pipeline: GPU residue split + df32
+  # CRT reconstruction over FEWER moduli (the accuracy knob), which
+  # eliminates the exact tier's O(num_moduli·N^2) CPU reconstruction.
+  # Being reduced precision, it is gated at the documented ~1e-8 bound
+  # (rel err <= 1e-7 across the same four regimes), NOT bit-exactness —
+  # and the gate also asserts the fast path genuinely engages (marker +
+  # no silent fallback to the exact tier). The DEFAULT path is untouched
+  # and stays covered by ozaki-ii-correctness above. Metal-only; SKIPS
+  # (no trace) on GPU-less hosts.
+  # ─────────────────────────────────────────────────────────────────
+  - name: ozaki-ii-fast
+    aliases: [ozaki-fast, ozaki-fast-tier, p-ozaki-fast]
+    requires:
+      - runtime_event:
+          event_kinds: [ozaki_fast]
+          event_names: ["ozaki_fast_gate"]
+          event_values: ["PASS"]
+        severity: high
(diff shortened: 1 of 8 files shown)


tsotchke:
Exact-head Atlas blocker at 451a0a49817fc6acf01235812b99bc583e79a603 (identity-clean). I rebuilt both eshkol-run and the local stdlib target, then ran both documented gate scripts with ESHKOL_RUN pointing to build-metalfast/eshkol-run.

Both fail because the -r test payload does not engage Metal. The exact gate gets numerically exact CPU-looking results but no out-of-range clamp warning. The fast gate gets the same roughly 1e-14 results and explicitly reports that the GPU fast-tier marker was not seen and the path did not engage.

This is not a missing-GPU or stale-binary issue. A tiny -e matmul using the same binary and FORCE_GPU environment initializes Apple M2 Ultra, prints the Ozaki pipeline, and emits the expected ESHKOL_OZAKI_NUM_MODULI=32 out-of-range warning. The defect is specific to how the gate payload runs under -r, likely the loop or scope dispatch path.

Please make both gates prove real GPU execution at this exact head. The fast gate must retain its fast-marker and no-fallback assertions, and the exact gate should assert an exact-path GPU marker in addition to its loud-clamp assertion. Both scripts must pass from a clean exact-head build with local stdlib.

PR #344, opened 2026-07-24
build: system-dependency resolution for all FetchContent deps — Homebrew-compatible builds (fixes #342)

### LINKED ISSUE #342: Homebrew install fails: Homebrew traps FetchContent ('eshkol_pcre2') during formula build
### Summary

`brew install tsotchke/eshkol/eshkol` fails at the CMake configure step. Homebrew now **traps and refuses** `FetchContent` population during formula builds, and the build declares several `FetchContent` dependencies at configure time. The first one reached (`eshkol_pcre2`) aborts the configure.

### Actual

```
==> cmake -B build -G Ninja -DLLVM_DIR=/opt/homebrew/opt/llvm@21/lib/cmake/llvm ...
-- eshkol-run linked with REPL JIT for -e/--eval support
-- Building package manager: eshkol-pkg
-- Building LSP server: eshkol-lsp
CMake Error at /opt/homebrew/Library/Homebrew/cmake/trap_fetchcontent_provider.cmake:12 (message):
  Refusing to populate dependency 'eshkol_pcre2' with FetchContent while
  building in Homebrew, please use a formula dependency or add a resource to
  the formula.
Call Stack (most recent call first):
  ...
  CMakeLists.txt:3748 (FetchContent_MakeAvailable)
-- Configuring incomplete, errors occurred!
```

Note this is a **policy trap**, not a network failure — Homebrew wants third-party deps declared as formula dependencies or `resource` blocks rather than fetched during the build.

### Other FetchContent deps that would follow

`eshkol_pcre2` is simply the first one hit. The same trap will apply to the other configure-time `FetchContent_MakeAvailable` calls (in our checkout: `eshkol_eigen`, `eshkol_sqlite_amalgamation`, `eshkol_zlib`, `eshkol_tree_sitter`), so fixing only pcre2 will just move the error.

### Possible fixes

- Declare the affected libraries as formula dependencies (`depends_on "pcre2"`, `"sqlite"`, `"zlib"` … ) and have CMake prefer a system copy when one is present (e.g. `find_package` first, `FetchContent` only as a fallback); or
- add `resource` blocks to the formula and point the build at the pre-staged sources; or
- gate the optional subsystems off in the formula's build (if pcre2/sqlite/tree-sitter aren't needed for the packaged binaries).

### Workaround

Building directly from source works fine — no Homebrew sandbox, so `FetchContent` is allowed:

```
cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Release
ninja -C build eshkol-run
```

### Environment

- Formula: `tsotchke/eshkol` tap, eshkol 1.3.3
- Homebrew on macOS 26.4.1 (arm64, M-series), LLVM 21.1.8

tsotchke:
## Problem

Homebrew 4.x+ installs a CMake dependency provider (`trap_fetchcontent_provider.cmake`, injected via `CMAKE_PROJECT_TOP_LEVEL_INCLUDES`) that **refuses all FetchContent population** during a formula build:

```
Refusing to populate dependency 'eshkol_pcre2' with FetchContent while
building in Homebrew, please use a formula dependency or add a resource
to the formula.
```

`brew install tsotchke/eshkol/eshkol` therefore fails at configure time on the first agent-FFI dependency (`eshkol_pcre2` at CMakeLists.txt). Because the release workflow auto-bumps the tap, **the next release would ship a broken install**.

### CMake (`CMakeLists.txt`)
Every bundled agent-FFI dependency becomes resolvable with **zero live FetchContent downloads**, while the default developer/release path is untouched:

- **`ESHKOL_HOMEBREW_BUILD`** (OFF) — umbrella: defaults `ESHKOL_USE_SYSTEM_PCRE2=ON` and enables `FETCHCONTENT_FULLY_DISCONNECTED` so a missing staged dependency fails fast instead of reaching the network.
- **`ESHKOL_USE_SYSTEM_PCRE2`** (OFF) — imports an installed PCRE2 (Homebrew `pcre2`) as an `IMPORTED STATIC` target under the exact name the bundled build defines (`pcre2-8-static`), so every downstream link / packaging copy / AOT-link reference is identical and needs no change.
- **`FETCHCONTENT_SOURCE_DIR_<NAME>`** (CMake-native) resolves the remaining deps from pre-staged resources. When set, CMake's `FetchContent_MakeAvailable()` skips the dependency provider entirely (so the Homebrew trap never fires) and performs no download.

The stock FetchContent path (all flags off) is unchanged: the original pcre2 logic is preserved verbatim inside the `else()` branch.

### Formula (`packaging/homebrew/eshkol.rb`)
- `depends_on "pcre2"`; pass `-DESHKOL_HOMEBREW_BUILD=ON -DESHKOL_USE_SYSTEM_PCRE2=ON`.
- `resource` blocks pinning the **same revisions** as CMakeLists.txt for sqlite / zlib / tree-sitter + 10 grammars / yoga (+ curl on Linux); each staged into `FETCHCONTENT_SOURCE_DIR_<NAME>`.
- Installs `libeshkol-agent-ffi.a` and the stable-named `eshkol-agent-*.a` archives so `(require agent.*)` AOT links work on a packaged install (previously never shipped).
- `on_linux`: libcurl resource + `openssl@3` for the native HTTP client (macOS uses NSURLSession — no curl).

### Release plumbing (`release.yml`, `update-homebrew-formula.sh`)
Anchor the auto-bump `url`/`sha256` substitutions to the **top-level** (2-space indent) so the new `resource` sha256 pins survive a release bump.

## Dependency inventory

| Dep | Homebrew formula? | Resolution chosen |
|-----|-------------------|-------------------|
| pcre2 | yes (`pcre2`) | system, imported static `pcre2-8-static` |
| sqlite (amalgamation) | yes, but custom compile-defines required | `resource` (URL zip) + `FETCHCONTENT_SOURCE_DIR` |
| zlib | yes | `resource` (git rev) + `FETCHCONTENT_SOURCE_DIR` |
| tree-sitter runtime | yes | `resource` (git rev) — kept staged to keep parser-ABI in lockstep with grammars |
| tree-sitter grammars ×10 | no | `resource` (git rev) each |
| yoga | no | `resource` (git rev) |
| curl (Linux only) | yes | `resource` (git rev); macOS uses NSURLSession |
| eigen | Windows/MSVC only | n/a (not reached on macOS/Linux) |
| moonlab | opt-in `ESHKOL_QUANTUM_ENABLED=OFF` | n/a (not reached by default) |

Nothing mandatorily requires FetchContent under `ESHKOL_HOMEBREW_BUILD`.

## Verification (macOS arm64, LLVM 21.1.8)

1. **Homebrew constraint simulated with the real trap provider** (`-DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=…/trap_fetchcontent_provider.cmake` + `ESHKOL_HOMEBREW_BUILD=ON` + `ESHKOL_USE_SYSTEM_PCRE2=ON` + `FETCHCONTENT_SOURCE_DIR_*` staged): configure completes, **0 "Refusing to populate", 0 downloads**; full `eshkol-run` build; `eshkol-run hello.esk` compiles+runs.
2. **System-PCRE2 regex end-to-end**: `tests/toolchain/release-agent-smoke.esk` (`regex-compile`/`regex-match?`) AOT-compiled+run → "release agent smoke", exit 0.
3. **Default path** (`ESHKOL_USE_SYSTEM_PCRE2=OFF`, pcre2 built from pinned source → `pcre2-8-static`): full build + smoke pass — the default path is unaffected.
4. **`brew install --build-from-source`** of the fix-branch source in a local tap: builds under Homebrew's real sandbox + trap; install + `test do` pass.
5. Both auto-bump mechanisms simulated: only the top-level `url`/`sha256` change; the resource sha256 is preserved.

Fixes #342

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1054,8 +1054,12 @@ jobs:
           cd /tmp/tap
 
           formula=Formula/eshkol.rb
-          sed -i -E "s|url \"https://github.com/${GITHUB_REPOSITORY}/archive/refs/tags/[^\"]+\"|url \"${tarball_url}\"|" "$formula"
-          sed -i -E "s|sha256 \"[0-9a-f]{64}\"|sha256 \"${sha}\"|" "$formula"
+          # Only rewrite the TOP-LEVEL url/sha256 (2-space indent). The formula
+          # also carries `resource` blocks whose own sha256 lines (4-space
+          # indent) pin the bundled agent-FFI dependencies — those must survive
+          # the bump untouched, so both substitutions are anchored to column 3.
+          sed -i -E "s|^  url \"https://github.com/${GITHUB_REPOSITORY}/archive/refs/tags/[^\"]+\"|  url \"${tarball_url}\"|" "$formula"
+          sed -i -E "s|^  sha256 \"[0-9a-f]{64}\"|  sha256 \"${sha}\"|" "$formula"
 
           if git diff --quiet; then
             echo "::notice::Formula already at ${tag}; nothing to do."
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3966,6 +3966,51 @@ endif()
 # runtime, but never silently disable or reduce it merely because of the host.
 option(ESHKOL_BUILD_AGENT_FFI "Build agent FFI modules" ON)
 
+# ──────────────────────────────────────────────────────────────────────────
+# Packaged / system-dependency resolution for the bundled agent-FFI native
+# dependencies (PCRE2, SQLite, zlib, tree-sitter + grammars, Yoga, curl).
+#
+# The default developer/release build downloads each dependency at pinned,
+# reproducible revisions via FetchContent (the block much further below).  That
+# is incompatible with sandboxed packaging environments such as Homebrew, which
+# inject a CMake dependency provider that refuses ALL FetchContent population
+# ("Refusing to populate dependency '<dep>' with FetchContent while building in
+# Homebrew, please use a formula dependency or add a resource to the formula").
+#
+# Two orthogonal, CMake-native escape hatches make a packaged build succeed
+# without a single network fetch, while leaving the default build byte-for-byte
+# unchanged:
+#
+#   * ESHKOL_USE_SYSTEM_PCRE2 links an installed PCRE2 (e.g. the Homebrew
+#     `pcre2` formula) under the same target name the bundled build defines,
+#     instead of fetching and compiling the pinned source.
+#
+#   * FETCHCONTENT_SOURCE_DIR_<UPPERCASE_NAME> points each remaining
+#     FetchContent dependency at a pre-staged source tree (a Homebrew
+#     `resource`).  CMake honors this natively: when set, the source-dir
+#     override "overrides everything else" — FetchContent_MakeAvailable() skips
+#     the dependency provider (so the Homebrew trap never fires) and performs no
+#     download (see the stock FetchContent.cmake).
+#
+# ESHKOL_HOMEBREW_BUILD is an umbrella that turns the packaging contract on: it
+# defaults ESHKOL_USE_SYSTEM_PCRE2 to ON and enables FETCHCONTENT_FULLY_
+# DISCONNECTED so any dependency the packager forgot to stage fails fast rather
+# than silently reaching the network.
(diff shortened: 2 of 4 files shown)


tsotchke:
Paired tap-side change: tsotchke/homebrew-eshkol#1 — both are required together (this PR adds the CMake flags the formula passes).

PR #364, opened 2026-07-25
fix(codegen): runtime guards raise instead of being optimized away (memory-safety sweep)

tsotchke:
Runtime safety guards across the tensor, complex and AD codegen surface were
being **deleted by the optimizer**. They existed in the source and not in the
binary.

## The mechanism

Every one of these guards emitted its diagnostic only inside:

```cpp
llvm::Function* pf = ctx_.lookupFunction("printf");
llvm::Function* ef = ctx_.lookupFunction("exit");
if (pf && ef) { ...print...; call exit(1); }
builder.CreateUnreachable();
```

Nothing ever registers `printf` or `exit` in the codegen function table —
`defineFunction` is only ever called for `eshkol_*` runtime helpers, hash-table,
parallel and tensorcore symbols. So both lookups always returned `nullptr`, the
failure block compiled to a bare `unreachable`, and LLVM correctly concluded the
guarded condition could not hold and **removed the branch entirely**.

## The fix

`CodegenContext::emitRaise` / `emitRaiseFmt` — the single sanctioned way to close
a guard's failure block. They raise through `eshkol_raise`, so:

- the failure block has side effects and cannot be folded away;
- the message reaches the user through the same channel as every other runtime
  error (`Unhandled exception: transpose: only supports 2D tensors (got 1D)`, exit 1);
- the failure is **catchable by `guard`**, so a library can recover.

`emitRaiseFmt` renders through `snprintf` into an entry-block buffer, so the
diagnostic still carries the runtime detail (the actual rank, the mismatched
dimension) that the dead `printf` intended to print.

## Measured before/after

Each out-of-contract call run individually against the compiler built from
`master`, then against this branch:

| call | before (master) | after |
|---|---|---|
| `(transpose <rank-1>)` | **SIGSEGV** at `0x0`, uncatchable | raises |
| `(transpose <rank-3>)` | `#((1 3) (2 4))` — fabricated 2x2 | raises |
| `(tensor-dot 2x3 2x2)` | `#((7 10) (19 28))` — **fabricated from out-of-bounds reads** | raises |
| `(tensor-dot <rank-3> 2x2)` | `#(7 10)` — wrong answer | raises |
| `(max-pool2d <rank-1> 2 2)` | `#()` silently | raises |
| `(avg-pool2d <rank-1> 2 2)` | `#()` silently | raises |
| `(conv2d <rank-1> k 1)` | `()` silently (**guard was entirely absent**) | raises |
| `(conv2d m <rank-1> 1)` | `Failed to allocate arena block of size 36028797018964007`, then `()` | raises |
| `(conv3d <rank-2> k 1 0)` | `#()` silently | raises |
| `(tensor-qr <rank-1>)` | **SIGSEGV** at `0x0` | raises |
| `(einsum "ij,jk->ik" <rank-1> <rank-1>)` | **SIGSEGV** at `0x0` | raises |
| `(einsum "ij,jk->ik" 2x2 2x2)` — *valid input* | **SIGSEGV** at `0x4008…` | `((7 10) (15 22))` |
| `(feed-forward m <rank-1> b e2 b)` | **SIGSEGV** at `0xff800001400fe308` (wild pointer) | raises |
| `(embedding v <rank-1>)` | **SIGSEGV** at `0x0` | raises |
| `(jacobian <scalar-fn> v)` | `0` — the exact integer, indistinguishable downstream | raises (interim; 1×n row is the target) |

Five uncatchable SIGSEGVs and one silently fabricated matrix, none of them
reportable to a user, all of them now catchable errors.

### The `tensor-dot` result was assembled from out-of-bounds reads

The matmul loops index `B[k*N + j]` for `k < a_cols`. With `A` 2x3 and `B` 2x2,
`a_cols` is 3 but `B` has only 2 rows, so `k = 2` reads `B[4]` and `B[5]` — past
the end of a 4-element buffer. The observed result proves it arithmetically:

```
C[0][0] = 1*B[0] + 2*B[2] + 3*B[4] = 1 + 6 + 3*B[4] = 7   =>  B[4] read as 0
C[0][1] = 1*B[1] + 2*B[3] + 3*B[5] = 2 + 8 + 3*B[5] = 10  =>  B[5] read as 0
C[1][0] = 4*B[0] + 5*B[2] + 6*B[4] = 4 + 15         = 19
C[1][1] = 4*B[1] + 5*B[3] + 6*B[5] = 8 + 20         = 28
```

Every term matches only if indices 4 and 5 of a 4-element buffer were read. The
`36028797018964007`-byte allocation request in the `conv2d` row is the same class
of evidence: a garbage extent loaded from `dims[-1]`.

## Sites converted, and what each guard protects

| file | guard | protects against | class |
|---|---|---|---|
| `tensor_shape_codegen.cpp` | `transpose` rank != 2 | `dims[1]` read past a rank-1 dimensions array | memory safety |
| `tensor_reduce_codegen.cpp` | `tensor-dot` rank > 2 | out-of-contract dims read | wrong answer |
| `tensor_reduce_codegen.cpp` | matmul inner dims mismatch | read past the end of B's element buffer | memory safety |
| `tensor_conv_codegen.cpp` | `max-pool2d` rank < 2 | `dims[ndim-2]` with `ndim==1` reads *before* the array | memory safety |
| `tensor_conv_codegen.cpp` | `avg-pool2d` rank < 2 | same negative dims index | memory safety |
| `tensor_conv_codegen.cpp` | `conv2d` kernel rank < 2 | negative dims index -> garbage extent -> 3.6e16 allocation | memory safety |
| `tensor_conv_codegen.cpp` | `conv2d` **input** rank < 2 | same, and this guard **did not exist at all** — added here | memory safety |
| `tensor_extras_codegen.cpp` | `conv3d` input/kernel rank < 3 | `dims[ndim-3]` reads before the array | memory safety |
| `tensor_linalg_codegen.cpp` | `tensor-qr` rank != 2 | `dims[1]` read past the array; the garbage `n` then sized three arena allocations | memory safety |
| `tensor_linalg_codegen.cpp` | `einsum "ij,jk->ik"` rank != 2 | out-of-bounds dims read | memory safety |
| `tensor_linalg_codegen.cpp` | `emitArenaAllocNullCheck` (19 call sites: lu/det/inverse/solve/cholesky/qr/svd) | store through null | memory safety |
| `tensor_transformer_codegen.cpp` | `feed-forward` W1 rank != 2 | `dims[1]` read past the array -> wild pointer | memory safety |
| `tensor_transformer_codegen.cpp` | `embedding` weights rank != 2 | `d_model` from `w_dims[1]` past the array | memory safety |
| `complex_codegen.cpp` | `packComplexToTagged` arena-alloc null | the struct is stored straight into the returned pointer | memory safety |
| `autodiff_codegen.cpp` | `jacobian` non-vector output | returned integer `0`; `m` is read from the test call's result | wrong answer (raise is interim — see below) |

13 memory-safety guards, 2 wrong-answer guards. The two arena-null classes are
not reachable without genuinely exhausting the arena, so for those the test pins
the happy path only; every other row has both halves asserted.

### `jacobian`: raises today, but the raise is not the end state

One row above is a semantic change rather than a restored guard, and the
reasoning should be inherited rather than re-derived.

Returning the exact integer `0` for a shape this code path cannot handle is
indefensible — it is indistinguishable from a real result in downstream
arithmetic — so raising is strictly better and is what ships here.

But to be precise about the mathematics: **the Jacobian of f: Rⁿ → R is
defined.** It is the 1×n matrix whose single row is the gradient. The raise
reflects a limitation of this implementation, not a property of the operator:
`m` is read from the test call's result, so a scalar result leaves this path with
no output dimension to build from. It is *not* a considered final contract.

The mathematically complete behavior is that `(jacobian <scalar-valued-fn> v)`
returns that 1×n row. The same argument very likely applies to the mirror case on
the `gradient` side — a vector-valued function passed to `gradient`, which
currently reports and terminates via `fprintf`/`abort` — and that path should be
revisited with this one.

Under this project's doctrine that a documented capability gets built up and never
scoped down, both are **BUILD items, not permanent contracts**: raises today, 1×n
row as the follow-up. Nothing in this PR should be read as settling the operator's
domain.

## Guard survival, proved in the IR

Worth being precise about *where* the deletion happens: the guard is present in
the IR the frontend emits on `master` too. What it emits there is a bare
terminator —

```llvm
; master, --dump-ir (pre-optimization)
transpose_dims_err:                               ; preds = %transpose_tensor
  unreachable
```

— and the branch is removed by the optimizer, not by codegen. Running the
project's own `-O2` pipeline over each `--dump-ir` output shows it directly:

```
$ opt -O2 -S master.ll   | grep -c transpose_dims_err   ->  0
$ opt -O2 -S master.ll   | grep -c transpose_dims_ok    ->  0
$ opt -O2 -S branch.ll   | grep -c transpose_dims_err   ->  2
$ opt -O2 -S branch.ll   | grep -c eshkol_raise         ->  3
```

On `master` the compare, the branch and **both** arms are gone at `-O2` — there
is no rank check left in the binary at all. On this branch the block survives
optimization, because it now has a side effect:

```llvm
transpose_dims_err:                               ; preds = %transpose_tensor
  %46 = call i32 (ptr, i64, ptr, ...) @snprintf(ptr %raise_msg_buf, i64 256, ptr @1, i64 %23)
  %47 = call ptr @eshkol_make_exception_with_header(i32 0, ptr %raise_msg_buf)
  call void @eshkol_raise(ptr %47)
  unreachable
```

## Test

`tests/features/codegen_guard_raise_test.esk`, registered as
`codegen_guard_raise_smoke`. It asserts **both halves** of every contract: the
out-of-contract call raises catchably, and the in-contract call still computes
its documented value — a guard that over-rejects would be just as broken.
35 checks, all passing under both JIT (`-r`) and AOT (`-O2`, compiled binary).

`conv3d` deserves a note: it always produces a **rank-3** result. `r_ndim` is
stored as the constant 3 and the dims array is a fixed 24-byte allocation, built
from the last three input dimensions — the leading dimensions are dropped,
despite a comment claiming otherwise. So a `1x1x2x2x2` input with a `1x1x1x1x1`
kernel yields shape `(2 2 2)`, and the test asserts that.

## Sweep

Swept the whole backend for the same disease in other costumes.
`lookupFunction("printf")`/`("exit")` now appears nowhere outside comments except
the four `tensorGet`/`tensorSet` sites owned by #359. All 67 `CreateUnreachable`
call sites were classified: 63 are legitimate terminators after a genuinely
side-effecting raise (`eshkol_raise`, `raiseDivideByZeroException`,
`emitOperandTypeError`, `eshkol_type_error_*`), and the remaining 4 are #359's.

### The same idiom in `llvm_codegen.cpp` is NOT dead — do not "fix" it the same way

Six `if (printf && exit)` guards remain, all in `lib/backend/llvm_codegen.cpp`
(non-callable call target, unresolved closure, legacy matmul inner-dims, FFT
power-of-2, FFT size overflow, FFT arena-alloc null). These resolve through the
**legacy** `function_table` / `module->getFunction`, where both `printf`
(llvm_codegen.cpp:3482) and `exit` (:3579) *are* registered. They are live —
verified at runtime:

```
$ eshkol-run -r -e '(display (fft #(1.0 2.0 3.0)))'
Error: FFT requires input length to be a power of 2
```

So they are a different, milder defect: they fire, but they `printf` to stdout and
`exit(1)` instead of raising, so they are not catchable by `guard` and do not go
through the runtime's error channel. Left alone here — converting them is a
behaviour change on working code, not a memory-safety fix, and two of the six sit
in call-dispatch paths another change is touching.

## Also fixed: `einsum "ij,jk->ik"` crashed on *valid* input (third commit)

Writing the in-contract half of the einsum rank-guard assertion is what exposed
this. `(einsum "ij,jk->ik" A B)` on two well-formed 2x2 tensors **SIGSEGVed** at
`0x4008000000000000` — the bit pattern of the double `3.0`, i.e. a matrix element
dereferenced as a pointer. It reproduces on unmodified `master`, and
`docs/reference/tensors/operations.md` lists this notation as working, so a
documented-supported operation was crashing on valid input.

`matmulSIMD` takes the two **tensor** pointers and reaches their element buffers
itself with `StructGEP(tensor_type, ptr, 2)`. The einsum call site had already
loaded the element pointers and passed *those*, so the struct GEP was applied to
the element array — field 2 of the tensor struct sits at the same offset as
`element[2]`, so it loaded `A`'s third element (`3.0`) and used it as the element
base pointer. One-line fix: pass the tensor pointers.

It is an independent defect from the guard sweep, hence its own commit. It is
included because the guard work surfaced it, it is a wild-pointer dereference in
a file already in this diff, and without it the einsum row of the new test could
only assert the raising half. With it, the test pins the result three ways: equal
to `(tensor-dot A A)`, equal to `A` against the identity, and equal to the
literal expected product `((7 10) (15 22))`.

## Note for after #359 lands

#359 adds `TensorCodegen::emitTensorRaise`, which does the same job as
`CodegenContext::emitRaise` for the tensor module only. Once both are on master,
`emitTensorRaise` should be collapsed into `ctx_.emitRaise` so there is one
implementation; kept separate here to avoid conflicting with that PR.

## Verification

- `codegen_guard_raise_smoke` 35/35, under `-r` and AOT `-O2`
- `ctest --parallel 4`: **97/97 passed** (re-run after merging current `master`)
- `scripts/run_features_tests.sh`: **37/37, 0 failed**
- `scripts/run_autodiff_tests.sh`: **58/58, 100%**
- `scripts/run_vm_parity.sh`: **80 passed, 0 failed**
- guard deletion and survival both confirmed by running `opt -O2` over the
  `--dump-ir` output of each build (counts above)
- backend sweep re-run independently: 67 `CreateUnreachable` sites, **0** of them
  without a preceding side-effecting noreturn call

Second commit is a one-line follow-up on the helper itself: `emitRaiseFmt` passed
a literal `i64` as `snprintf`'s `size_t` parameter, which `TypeSystem` defines as
`i32` on wasm32 (`type_system.cpp:39`). Native was unaffected and
`wasm-execute-diff` passed either way — no program in the wasm corpus reaches one
of these guards, so the mismatched call was never emitted there — but it was
wrong. The width now comes from `types_.getSizeType()`.

ASAN was not used: the out-of-bounds accesses happen in JIT/AOT-generated code
reading within an already-allocated arena block, so an ASAN build of the
*compiler* cannot witness them. The arithmetic reconstruction above and the
wild-address SIGSEGV are the direct evidence instead.


tsotchke:
## Two adjacent defects found while validating these guards — reported, deliberately not fixed here

Both were found by writing the happy-path half of each guard's test. Neither is
touched by this change (this PR only rewrites guard *failure* blocks; the success
paths are byte-identical), and neither belongs in a guard sweep.

### 1. `einsum "ij,jk->ik"` faults on a fully in-contract call

```scheme
(einsum "ij,jk->ik" (tensor 2 2 1.0 2.0 3.0 4.0) (tensor 2 2 1.0 0.0 0.0 1.0))
```

```
[Eshkol] fatal signal: SIGSEGV (segmentation fault) at address 0x4008000000000000
```

`0x4008000000000000` is the IEEE-754 bit pattern for the double `3.0` — an
element value being dereferenced as a pointer, in the delegation to
`matmulSIMD`. This looks like operand classification rather than anything
guard-related, so it is left for the operand-by-runtime-tag work.

Consequence for this PR: `"ij,jk->ik"` is the one guard whose happy path cannot
be pinned yet, so the test asserts only that its rank guard raises, and uses
`(einsum "ii->" m2x2)` as the einsum happy-path witness. The test file records
why inline so it is not mistaken for an oversight.

### 2. `conv3d` drops its leading dimensions, contradicting its own comment

`lib/backend/tensor_extras_codegen.cpp` says:

```cpp
// Create output tensor with same leading dimensions, new spatial dims
```

The code does not do that. `r_ndim` is stored as the constant `3`, the result
dims array is a fixed 24-byte (3 x i64) allocation, and the three stored extents
are computed from the *last three* input dimensions only. So a rank-5
`(N C D H W)` input contracts to a rank-3 `(D' H' W')` result and the batch and
channel dimensions are silently discarded:

```scheme
(tensor-shape (conv3d (make-tensor '(1 1 2 2 2) 1.0)
                      (make-tensor '(1 1 1 1 1) 1.0) 1 0))   ; => (2 2 2)
```

The comment is left in place on purpose: the gap should be closed by building
`conv3d` up to preserve leading dimensions, not by editing the comment down to
match the current behaviour. Until then the test asserts the behaviour that
actually ships, with the discrepancy noted.

tsotchke:
**Update on finding 1 above:** the `einsum "ij,jk->ik"` fault turned out to be a
one-line call-site error rather than anything needing the broader
operand-by-runtime-tag work, so it is fixed in this PR (`2c4c42bb`) instead of
being deferred. `matmulSIMD` wants the tensor pointers; the call site was
passing the already-loaded element buffers. The einsum row of the guard test now
has a real happy path asserted three ways.

Finding 2 (`conv3d` dropping its leading dimensions) still stands as written and
is **not** addressed here.

PR #369, opened 2026-07-25
fix(tensor): reshape preserves its source payload

tsotchke:
## Symptom

`(tensor-data (reshape v n n))` returned all zeros while `v` itself held the right numbers, so every downstream `matmul` computed on zeros and returned a plausible-looking wrong answer with no error. It surfaced as four failing `diagonal` cases in `tests/gpu/sf64_primitives_test.esk`.

Minimal reproduction, before / after:

```scheme
(define (probe n)
  (define A-data (make-vector 4 0.0))
  (define (fill! i) (if (< i 4) (begin (vector-set! A-data i (+ i 1.0)) (fill! (+ i 1))) #t))
  (fill! 0)
  (define A (reshape A-data n n))
  (tensor-data A))
(probe 2)
```

| | `A-data` | `(tensor-data (reshape A-data 2 2))` |
|---|---|---|
| before | `#(1 2 3 4)` | `#(0 0 0 0)` |
| after | `#(1 2 3 4)` | `#(1 2 3 4)` |

## Root cause — not reshape, and not the operand-classification family

`reshape` was already correct at every layer. It routes its operand through `unpackTensorOperandChecked` / `eshkol_tensor_operand_checked` (ESH-0069), and its Scheme-vector path converts the vector to a tensor element by element. Passing a literal vector always worked:

```
(display (tensor-data (reshape (vector 1.0 2.0 3.0 4.0) 2 2)))   =>  #(1 2 3 4)
```

The defect was in `transformInternalDefinesToLetrec` (`lib/frontend/parser.cpp`). It hoisted **every** internal define's initializer into the enclosing `letrec*` head, so a value define written after other body expressions evaluated **before** them. `(define A (reshape A-data n n))` therefore ran before `(fill! 0)`, and because `reshape` snapshots a vector operand into tensor storage, it snapshotted the pre-fill zeros.

The probe needs no tensors at all:

```scheme
(define (h) (display "A") (define x (begin (display "B") 5)) (display "C") x)
```

| | output |
|---|---|
| native, before | `BAC` |
| native, after | `ABC` |
| bytecode VM (unchanged) | `ABC` |

So this was also a native/VM divergence, with the VM on the correct side.

## Fix

Every internal define's **name** still joins one `letrec*` binding set wherever it appears in the body, so all internal names keep sharing a single mutually visible recursive scope — the capture analysis and mutual-recursion behaviour that depends on that is unchanged. What changes is initializer *evaluation order*:

- a **function** define keeps its hoisted lambda — constructing a closure has no observable effect, and hoisting is what lets a body expression call a function whose define appears later;
- a **leading value** define (nothing runs before it) keeps its head initializer, which is already first in source order;
- a **value define that follows any body expression** has its initializer left at that source position: the slot is bound unset in the head and assigned in place. `letrec*` slots are activation cells read indirectly, so a hoisted lambda closing over such a name still observes the value once the assignment has run.

`TypeChecker::synthesizeLet` no longer narrows an unset-and-later-assigned `letrec*` slot to `Null` — the value's type comes from the assignment, so keeping `Null` there would warn on every downstream use of a well-typed program. This is the same soundness rule `synthesizeIf` already applies when a guarded branch reassigns the refined variable.

**Behaviour boundary:** reading a value define's variable *before* its source position now yields the unset slot rather than the eventual value. That was already the documented intent (the old code comment said "referencing it there is a bug in the user's code, per R6RS / Racket"); the implementation simply used to over-deliver. Pinned as a test so the contract is deliberate.

## Class audit — re-viewing tensor ops

Each op fed a source built with `make-vector` + mutation, then re-viewed through an interspersed `define` (the shape that lost the payload):

| op | vector operand | list operand | tensor operand | payload preserved |
|---|---|---|---|---|
| `reshape` | works | clean type error | works | yes |
| `tensor-reshape` | works | clean type error | works | yes |
| `flatten` | works | clean type error | works | yes |
| `squeeze` | works | clean type error | works | yes |
| `unsqueeze` | works | clean type error | works | yes |
| `transpose` | works | clean type error | works | yes |
| `tensor-data` | works | clean type error | works | yes |
| `tensor-shape` | works | clean type error | works | yes (shape correct) |

Every one of them was silently zeroed by the same root cause before this change and is correct after it. A list operand is refused with a catchable `Type error in <op>: expected tensor, got pair` — never a silent zero. That is the current `eshkol_tensor_operand_checked` contract ("tensor or numeric vector"); #365 extends it to accept numeric lists, and the new regression test accepts either outcome so it stays green across that merge.

## The four diagonal cases, with real values

`A = diag(1..n)`, `I = identity(n)`, so `A*I = A`. n=47 and n=50 cross the GPU dispatch threshold.

| n | C[0,0] | C[1,1] | C[n-1,n-1] | C[0,1] | diagonal sum | expected sum |
|---|---|---|---|---|---|---|
| 32 | 1 | 2 | 32 | 0 | 528 | 528 |
| 46 | 1 | 2 | 46 | 0 | 1081 | 1081 |
| 47 | 1 | 2 | 47 | 0 | 1128 | 1128 |
| 50 | 1 | 2 | 50 | 0 | 1275 | 1275 |

## Tests

`tests/parser/letrec_star_regression_test.esk` **already covered this exact bug**, including a "the original reshape bug scenario" case — but it only *displayed* each result and then printed an unconditional "All letrec* regression tests passed!" banner, so it reported success while printing `0` for **three of its four cases** (`scripts/run_parser_tests.sh` judges a test by its exit code). Same invisibility class as the indented `diagonal` failures. Confirmed by re-running the exact `letrec*` the old transform emitted:

| original case | old lowering | new lowering |
|---|---|---|
| 1 interleaved side effect | `0` | `42` |
| 2 multiple side effects | `0` | `10` |
| 3 consecutive leading defines | `60` | `60` |
| 4 reshape after `vector-set!` | `0` | `5` |

It now compares every value, and adds coverage for chained deferred initializers, forward calls over a late slot, the read-before-position boundary, and mutual recursion across an intervening expression.

`tests/tensor/reshape_payload_preserved_test.esk` is new: it pins the ordering contract directly, checks payload preservation across the whole re-viewing family over vector / list / tensor operands, and runs the end-to-end reshape-into-matmul shape at a scalar size and a GPU-dispatched size against exact expected numbers.

Both are registered in ctest under JIT and AOT with `PASS_REGULAR_EXPRESSION` + `FAIL_REGULAR_EXPRESSION`, so neither can report success without its numbers matching.

## Docs

Four docs stated the pre-existing "only consecutive defines at the start of a body are grouped into `letrec*`" rule, which had not been true since interspersed defines were adopted and would misdirect exactly this kind of debugging. `COMPLETE_LANGUAGE_SPECIFICATION.md` §12.5 now specifies the interspersed contract (and says `letrec*`, not `letrec`); `breakdown/OVERVIEW.md`, `breakdown/COMPILER_ARCHITECTURE.md`, `breakdown/SCHEME_COMPATIBILITY.md` and `breakdown/DEVELOPER_TOOLS.md` are corrected to match.

## VM parity items found (reported, not fixed here — different root cause)

The VM's ordering and `reshape` payload both agree with the fixed native behaviour (`(tensor-ref (reshape (vector 1.0 2.0 3.0 4.0) 2 2) 0 0)` = 1, `... 1 1` = 4). Two unrelated divergences turned up while checking:

1. `tensor-data` returns a **list** on the VM where native returns a **vector** — `(1 2 3 4)` vs `#(1 2 3 4)`. `tests/vm_parity/PARITY.tsv` marks `tensor-data` `vm-supported` while marking its alias `tensor->vector` a `gap`.
2. `(tensor 5.0 6.0 7.0)` builds a **5-element tensor filled with 6.0** on the VM (shape `(5)`, elements `6 6 6`), where native builds `#(5 6 7)` with shape `(3)`.

## Unblocks the CUDA lanes on #367

`linux-x64-cuda` / `linux-arm64-cuda` were failing on #367 with exactly these four cases:

```
Testing sf64_primitives_test.esk    FAIL MARKER
    39:  32x32 diagonal: FAIL
    40:  46x46 diagonal: FAIL
    41:  47x47 diagonal: FAIL
    42:  50x50 diagonal: FAIL
```

That was this defect becoming visible for the first time, not new breakage — #367 is what made the indented `FAIL` lines detectable. The root cause is in the frontend (internal-define lowering), so it is backend-independent: it reproduces and is fixed identically on the scalar path (n=32, 46), the BLAS path, and the GPU path (n=47, 50 cross the dispatch threshold, verified on Metal). Nothing in the fix is CUDA-specific.

## Relationship to #368

Not the same root cause. #368 fixes `tensor-get`'s scalar partial index (`flat[i]` vs `flat[i*stride0]`) via `eshkol_tensor_slice_offset_from_index_arg` in `runtime_tensor_index.cpp`. `reshape` does not use that index/offset machinery at all — it reuses the source elements pointer unchanged and only builds a new dimensions array (`eshkol_tensor_to_dims` / `eshkol_cons_list_to_dims` / `eshkol_compute_dims_total`). No file overlap either: this PR touches `lib/frontend/parser.cpp`, `lib/types/type_checker.cpp`, `CMakeLists.txt`, four docs and two tests. The two fixes are independent and should merge in either order.

## Verification

- `ctest` — 100/100, including the four new JIT/AOT entries
- `scripts/run_all_tests.sh` — 42/42 suites, 100% pass rate
- `scripts/run_vm_parity.sh` — 80 passed, 0 failed
- `scripts/run_gpu_tests.sh` — 17/17, `sf64_primitives_test.esk` diagonals PASS with the values above
- both new tests also pass under `--strict-types`
- `scripts/gen_language_surface.py --check` — manifest unchanged
- interspersed defines additionally spot-checked inside `cond`, `case`, `when`, `unless`, `do`, `let*`, `letrec`, `lambda`, `guard`, named-let bodies and top-level `begin`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2862,6 +2862,60 @@ if(ESHKOL_BUILD_TESTS AND TARGET eshkol-run)
             PASS_REGULAR_EXPRESSION "PASS: matmul tensor reads inside defined functions match top-level"
             FAIL_REGULAR_EXPRESSION "FAIL:|LLVM module verification failed|fatal signal")
     endif()
+    # v1.3.4 blocker guard: a re-viewing tensor op (reshape/tensor-reshape/
+    # flatten/squeeze/unsqueeze/transpose/tensor-data) must carry its source
+    # payload into its result, including when the result is bound by an internal
+    # `define` written after the expressions that filled the source. The parser
+    # used to hoist every internal define's initializer into the enclosing
+    # letrec* head, so `(define t (reshape v n n))` ran before the loop that
+    # filled `v` and snapshotted zeros — matmul then computed on zeros with no
+    # error. Also pins the underlying evaluation-order contract directly.
+    # Verified on both -r (JIT) and AOT.
+    if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/tests/tensor/reshape_payload_preserved_test.esk")
+        add_test(
+            NAME reshape_payload_preserved_jit_smoke
+            COMMAND $<TARGET_FILE:eshkol-run> -r
+                    "${CMAKE_CURRENT_SOURCE_DIR}/tests/tensor/reshape_payload_preserved_test.esk"
+        )
+        set_tests_properties(reshape_payload_preserved_jit_smoke PROPERTIES
+            TIMEOUT 120
+            PASS_REGULAR_EXPRESSION "PASS: reshape and the re-viewing tensor ops preserve their source payload"
+            FAIL_REGULAR_EXPRESSION "FAIL:|LLVM module verification failed|fatal signal")
+        add_test(
+            NAME reshape_payload_preserved_aot_smoke
+            COMMAND sh -c
+                "'$<TARGET_FILE:eshkol-run>' -o '${CMAKE_CURRENT_BINARY_DIR}/reshape_payload_preserved_aot' '${CMAKE_CURRENT_SOURCE_DIR}/tests/tensor/reshape_payload_preserved_test.esk' && '${CMAKE_CURRENT_BINARY_DIR}/reshape_payload_preserved_aot'"
+        )
+        set_tests_properties(reshape_payload_preserved_aot_smoke PROPERTIES
+            TIMEOUT 180
+            PASS_REGULAR_EXPRESSION "PASS: reshape and the re-viewing tensor ops preserve their source payload"
+            FAIL_REGULAR_EXPRESSION "FAIL:|LLVM module verification failed|fatal signal")
+    endif()
+    # The internal-define ordering contract itself. tests/parser/ is otherwise
+    # only run by scripts/run_parser_tests.sh, which judges a test by its exit
+    # code — this file used to print its results and an unconditional pass
+    # banner, so it reported success while printing 0 for the reshape case. It
+    # now compares every value, and this entry gates on the comparison.
+    if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/tests/parser/letrec_star_regression_test.esk")
+        add_test(
+            NAME internal_define_order_jit_smoke
+            COMMAND $<TARGET_FILE:eshkol-run> -r
+                    "${CMAKE_CURRENT_SOURCE_DIR}/tests/parser/letrec_star_regression_test.esk"
+        )
+        set_tests_properties(internal_define_order_jit_smoke PROPERTIES
+            TIMEOUT 120
+            PASS_REGULAR_EXPRESSION "PASS: internal defines preserve source-order evaluation"
+            FAIL_REGULAR_EXPRESSION "FAIL:|LLVM module verification failed|fatal signal")
+        add_test(
+            NAME internal_define_order_aot_smoke
+            COMMAND sh -c
+                "'$<TARGET_FILE:eshkol-run>' -o '${CMAKE_CURRENT_BINARY_DIR}/internal_define_order_aot' '${CMAKE_CURRENT_SOURCE_DIR}/tests/parser/letrec_star_regression_test.esk' && '${CMAKE_CURRENT_BINARY_DIR}/internal_define_order_aot'"
+        )
+        set_tests_properties(internal_define_order_aot_smoke PROPERTIES
+            TIMEOUT 180
+            PASS_REGULAR_EXPRESSION "PASS: internal defines preserve source-order evaluation"
+            FAIL_REGULAR_EXPRESSION "FAIL:|LLVM module verification failed|fatal signal")
+    endif()
     add_test(
         NAME autodiff_tensor_reductions_runtime_smoke
         COMMAND $<TARGET_FILE:eshkol-run>
--- a/docs/COMPLETE_LANGUAGE_SPECIFICATION.md
+++ b/docs/COMPLETE_LANGUAGE_SPECIFICATION.md
@@ -2787,7 +2787,7 @@ For closures with captures, use bounce/trampoline:
 
 ### 12.5 Internal Defines
 
-Function bodies can contain internal definitions, which are automatically transformed to `letrec`:
+Function bodies can contain internal definitions, which are automatically transformed to `letrec*`:
 
 ```scheme
 (define (outer x)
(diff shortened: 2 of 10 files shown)


tsotchke:
## Direct proof against #367's honest harness

My own CUDA lane on this PR cannot prove the fix: this branch is cut from `origin/master`, so it still runs the old `scripts/run_gpu_tests.sh`, whose `^FAIL:`-anchored match cannot see the indented `  <case>: FAIL` lines. The harness that *can* see them lives in #367.

So I ran #367's `scripts/run_gpu_tests.sh` (plus its `scripts/lib/test_isolation.sh` prelude) against this branch's compiler, changing nothing else, and then A/B'd it by reverting only `lib/frontend/parser.cpp` to `origin/master`.

**With `lib/frontend/parser.cpp` reverted to `origin/master`** — reproduces the CUDA CI failure exactly:

```
Testing sf64_primitives_test.esk    FAIL MARKER
    56:  32x32 diagonal: FAIL
    57:  46x46 diagonal: FAIL
    58:  47x47 diagonal: FAIL
    59:  50x50 diagonal: FAIL
Total Tests: 17   Passed: 16   Failed: 1
  - sf64_primitives_test.esk
Pass Rate: 94%
exit 1
```

**With this branch as-is:**

```
Testing sf64_primitives_test.esk    PASS
Total Tests: 17   Passed: 17   Failed: 0
Pass Rate: 100%
exit 0
```

Only `lib/frontend/parser.cpp` differs between those two runs, which pins the cause to the internal-define lowering and nothing in the tensor path. #367's CUDA lanes should go green once this lands.

The four cases' real values (not just the absence of `FAIL`), n=47 and n=50 crossing the GPU dispatch threshold on Metal:

| n | C[0,0] | C[1,1] | C[n-1,n-1] | C[0,1] | diagonal sum | expected |
|---|---|---|---|---|---|---|
| 32 | 1 | 2 | 32 | 0 | 528 | 528 |
| 46 | 1 | 2 | 46 | 0 | 1081 | 1081 |
| 47 | 1 | 2 | 47 | 0 | 1128 | 1128 |
| 50 | 1 | 2 | 50 | 0 | 1275 | 1275 |

PR #370, opened 2026-07-25
feat(memory): user-reachable region handles for non-lexical scoped reclamation (closes #341)

### LINKED ISSUE #341: Enhancement: user-reachable arena scoping (mark/release) so long-running loops can reclaim
### Summary

**This is an enhancement request, not a bug report** — the current behaviour follows directly from Eshkol's documented no-GC arena design, which we like. The ask is a way to *scope* reclaim so that long-running loops become viable without giving that design up.

### Observed

Compiled programs allocate into a single process-lifetime `__global_arena` (`getArenaPtr`, `lib/backend/arithmetic_codegen.cpp:43`). `arena_reset` exists (`runtime_arena_core.cpp:446`) but nothing calls it during a run, so reclaim happens at process exit. Any loop that allocates per iteration therefore grows monotonically.

Measured on a small MLP training loop (352 parameters, 256 samples/step), peak RSS by step count:

| steps | peak RSS | outcome |
|------:|---------:|---------|
| 5     | 2.05 GB  | completed |
| 10    | 7.02 GB  | completed |
| 20    | 7.76 GB  | OOM-killed |

The AD tape participates in this: `ad_tape_grow` (`lib/backend/vm_autodiff.c:82`) doubles capacity and abandons the previous array, with the comment *"Old nodes remain in arena — reclaimed on region pop"* — but for this path no region pop ever comes.

To be clear about scale-dependence: at small sizes this is a non-issue. A ~27-parameter constraint evaluation costs roughly **160 KB per gradient call**, so a persistent process comfortably serves tens of thousands of calls (we measured 1.386 GB @ 500 calls → 2.587 GB @ 8,000, all completing). It is specifically *large* per-iteration allocation that runs out of headroom.

### Request

Some user-reachable scoping primitive, e.g. an arena mark/release pair or a `with-arena-scope` form, so a loop body's allocations can be reclaimed explicitly while values that must survive are copied out. `vm_region_push`/`vm_region_pop` already exist in `vm_arena.h`, so the primitives appear to be present — what's missing is a way to reach them from the language.

This would keep the "no GC pauses, ever" guarantee intact while making long-running loops practical.

### Context

We're using Eshkol as a compiled verifier/constraint instrument called from a host process (short-lived scoring and curvature calls), where this never bites. We hit it only when experimenting with longer in-language training loops.

### Environment

- eshkol `cddadded` (v1.3.3-evolve line)
- LLVM 21.1.8, macOS 26.4.1 (arm64, M-series, 16 GB), Release build

*(Absolute OOM thresholds above are affected by other load on the machine; the monotonic growth with step count is the reproducible part.)*

tsotchke:
Closes #341.

An external reporter measured a 161-parameter MLP training loop growing ~123 MB/step without bound, and asked for a user-reachable mark/release primitive. We accepted the ask publicly with measured numbers; this builds it.

## Why the existing mechanisms don't cover this

`with-region` already works for this shape — region entry swaps the arena the AD tape is allocated from, so a per-step region reclaims the whole tape. The **automatic** per-iteration nursery (ESH-0214e) does not, and cannot: its static escape analysis disqualifies any loop body containing a `gradient`/derivative op, a `set!`, or a `tensor-set!`, and a training step trips all three. That exclusion is by design — the nursery targets pure or structurally-mutating tick loops.

So the gap was **ergonomic, not mechanical**: the machinery exists and works, but had no non-lexical surface.

## What shipped, and why this shape

```scheme
(region-open [name] [size])  ; -> opaque exact-integer handle
(region-close handle v ...)  ; -> the deep-promoted keeps; region reclaimed
(region-open? handle)        ; -> #t while the handle names a live region
```

**`with-region` remains the recommended default and is behaviourally unchanged.** It is RAII-shaped: bound to a lexical block, so it is unwind-safe and impossible to forget to close. The handle pair is the unstructured primitive, documented as such.

**The region-handle form was chosen over `arena-mark` / `arena-release`, and the second form was not shipped.** The mark/release alternative would sit on `arena_push_scope` / `eshkol_arena_iter_scope_end` over the *shared* process arena, which is strictly less safe for a user-facing surface: a shared-arena rewind cannot be classified by `region_index_owning()`, so the escape evacuator cannot tell an escaping pointer from a stable one; it degrades to commit-only for the duration of any parallel scope; and it cannot be poisoned per-scope, losing the `0xCB` dangling-pointer tripwire the ESH-0214 tests rely on. Shipping it as a thin alias over the region form would also mean two names for one mechanism with different implied semantics. One surface, one mechanism.

**The handle is a slot index plus a generation counter, not a pointer.** Two reasons. (1) Cost: the API exists for hot loops, and a heap handle object would have to be allocated in the *enclosing* arena — otherwise closing the region would free the handle naming it — leaking one object per iteration and reintroducing in miniature the exact unbounded growth #341 is about. An immediate integer allocates nothing. (2) Safety: the generation counter makes every stale token *detectably* stale, which is what turns double-close and use-after-close from memory corruption into a clean error.

## Acceptance measurement

Reporter's shape reproduced: 161-parameter MLP (14 → 10 → 1, tanh), full-batch `gradient` per step, fresh parameter vector written each step. Peak RSS in MB, native AOT, `/usr/bin/time -l`. Source: `tests/memory/region_handle_training_rss.esk` — three modes over the same code and the same binary.

| steps | 5 | 10 | 20 | 40 |
|---|---|---|---|---|
| plain loop (no scoping) | 632 | 1258 | 2510 | 5013 |
| `region-open`/`region-close`, default size hint | 415 | 612 | 727 | 993 |
| `with-region` per step, sized region | 131 | 131 | 132 | 132 |
| **`region-open`/`region-close`, sized region** | **131** | **131** | **132** | **132** |

The plain row reproduces the report (622/1236/2464/4920) to within noise. All four rows produce **bit-identical** trained parameters (`loss=0.001581781325962818 checksum=16.240912553050837` at 40 steps).

**Flat, not merely sub-linear** — 131 MB is the single-step floor (one step's tape is ~126 MB), so the sized-region rows are as flat as the shape allows.

**A finding worth stating.** Reclamation is complete in *both* handle rows — instrumenting the surviving arena shows it grows by exactly 2592 bytes per step, the promoted 161-double parameter vector and nothing else. The difference between the two rows is the *allocator's* high-water mark: with the default 64 KiB hint each step's arena reaches ~126 MB by adding geometrically doubling blocks, and freeing that block sequence every step walks the peak upward. A hint large enough to hold one step's work in a **single block** keeps the address ranges stable. This applies to `with-region` equally, and explains most of the residual in the numbers quoted on the issue — it is now documented as the primary tuning knob for either form.

## Safety matrix

`tests/memory/region_handle_safety_test.esk`, 42 cases, run AOT **and** JIT under `ESHKOL_ARENA_POISON=1`:

| Situation | Outcome | Verified |
|---|---|---|
| Double close | Clean catchable error, nothing freed twice | `double-close-raises` |
| Use after close | Same clean error; `region-open?` is `#f` and never raises | `use-after-close-*` |
| Close out of order | **Defined cascade**: inner regions closed innermost-first, keeps promoted at every level, inner tokens invalidated | `cascade-*` (4 cases) |
| Never closed | Bounded — the 65th simultaneous handle raises, naming the cause; runtime stays usable | `unclosed-handles-hit-a-clean-limit` |
| Fabricated / stale / `0` / negative token | Fails validation, clean error | `fabricated-*`, `zero-*`, `negative-*` |
| Reused slot, old token | Rejected (generation mismatch) while the slot works for its new token | `reused-slot-*` (3 cases) |
| Value not kept, after close | Genuinely unreachable; 50 keep-nothing cycles grow the surviving arena < 8 KB | `unkept-allocations-do-not-escape` |
| Kept value is deep | Interior list/vector/string reachable after close (the ESH-0214d lesson) | `kept-deep-*` (3 cases) |

Full pasted output is in the thread below.

## Unwind semantics (defined and implemented)

**A non-local exit that crosses an open region closes it.** For `raise`/`guard` and for a `call/cc` continuation escape, and for `with-region` bodies as well as handles:

- the **in-flight value** — the raised value, or the value delivered to the continuation — is deep-promoted out of every region being closed *before* any arena is freed, so a handler never receives a pointer into freed memory;
- the allocation-routing slot is restored before the arena dies;
- the closed handles' tokens are invalidated, so `(region-open? h)` reads `#f` in the handler and a later close raises cleanly.

Mechanically the region depth is recorded as a mark beside the existing `wind_mark` / `promise_mark` on `eshkol_exception_handler_t` and `eshkol_continuation_state_t`, and the unwind runs in the same place those two do.

**This also fixes `with-region`.** Before this change, a `raise` out of a `with-region` body jumped past `region_pop` *and* `eshkol_region_leave`, leaking the region **and** leaving the shared allocation slot pointing at an arena that was never freed. Now covered, and tested (`raise-out-of-with-region-delivers-value`, plus a following allocation that would be corrupt if the slot were still hijacked).

**One teardown path.** Explicit close, out-of-order cascade, `with-region` exit, raise, and continuation escape all funnel through `eshkol_region_unwind_to()` — promote the keeps one level out, restore the allocation slot, pop — so the structured and unstructured surfaces cannot drift apart. `with-region`'s lowering now tears down **by mark** rather than by "pop the innermost", which additionally means a body that leaks an open handle no longer causes that pop to destroy the wrong region.

## Both substrates

One shared C implementation backs native and the VM, so the handle protocol, its validation and its error text are identical by construction rather than by convention — the same uniform-contract pattern as the bytevector bounds checks. `tests/vm_parity/corpus/region_handle_contract.esk` pins the observable output as byte-identical across native, `vm-src` and `vm-eskb`; **it caught a real divergence** (the VM's close was not cascading on an out-of-order close), now fixed.

Reclamation itself is native-only, with a justified `PARITY.tsv` row: the VM heap has no escape evacuator to deep-promote kept values with, which is also why `with-region` is a pass-through there. Program output is identical; only the reclamation is absent. Building a VM evacuator is a separate project.

## WASM / lite lane

Because `with-region` lowering now routes through `eshkol_region_unwind_to`, that symbol and `eshkol_region_mark` become WASM `env` imports in **any** module using `with-region` — not only ones using handles. The browser JS glue therefore gains degradation stubs for those two plus the three handle builtins, in both `web/eshkol-repl.js` and `site/static/eshkol-runtime.js`. `scripts/check_wasm_imports.py` is green across all 10 smoke surfaces (102 unique env imports, all provided).

Worth flagging for whoever owns that checker: its JS key scanner treats an apostrophe inside a `//` comment as a string delimiter, so a comment containing `'()` or `runtime's` silently truncates the scan and produces phantom "MISSING" reports for every key after it. That is how this landed initially (it phantom-reported four symbols that were plainly present); the comments are now apostrophe-free, but the scanner should strip line comments before the brace/string walk.

## Verification

- `tests/memory/region_handle_training_rss_test.sh` — **PASS** (flat 33 MB vs 139→539 MB baseline at the CI batch, 16.3x; numerics identical across all three modes; safety matrix clean AOT + JIT under poison)
- `scripts/run_vm_parity.sh` — **82 passed, 0 failed**
- `scripts/run_memory_tests.sh` — **20/20**, including `region_evac_subtype_coverage`, `region_deep_escape`, `region_mutating_loop_flat_rss`
- `tests/memory/region_evac_subtype_coverage_test.sh` — PASS (129 MB / 300 MB ceiling, poisoned)
- `tests/memory/define_loop_flat_rss_aot_test.sh` — PASS (8 MB / 200 MB)
- `tests/memory/iter_scope_partial_reclaim_test.sh` — PASS (35 MB, poison clean, JIT clean, within 10x of the with-region twin)
- `scripts/run_autodiff_tests.sh` — **58/58**
- `scripts/run_error_handling_tests.sh` 7/7, `scripts/run_control_flow_tests.sh` 11/11 (the runtime I touched)
- VM standalone source tests — **59/59** (6 new region-handle cases)
- `ctest` — **96/96**
- `gen_language_surface.py --check` OK; `test_language_coverage_gate.py` 12/12. Manifest regenerated; the 5 new names are categorised `memory_region` and executed by real tests on both substrates.
- Wired into `scripts/run_icc_smoke.sh` and `.icc/completion-oracles.yaml` as `region_handle_scoped_reclamation`.

## Deliberately left for later

- **VM reclamation.** Needs a VM-heap escape evacuator; `PARITY.tsv` row records it. `with-region` has the same limitation today.
- **`with-region`'s VM lowering has two pre-existing defects** unrelated to this change, found while mapping the surface: it omits `OP_POP` for non-final body expressions (so `(with-region e1 e2)` strands `e1`'s value on the operand stack), and it compiles the `'name` / `('name size)` prefix as code rather than skipping it, so `(with-region ('r 1024) body)` compiles a *call* of the quoted symbol. Not touched here to keep this change reviewable; worth its own issue.
- **Two pre-existing syntax defects in the assurance plumbing**, both confirmed present on `origin/master` and only *shifted* by my insertions (so neither is mine, and neither blocks this PR):
  - `.icc/completion-oracles.yaml` — a `label:` followed directly by `event_names:` with no `- runtime_event:` block. `yaml.safe_load` rejects the file at line 739 before this change and 759 after (= 739 + my 20 lines).
  - `scripts/run_icc_smoke.sh` — `bash -n` reports a syntax error at the `matmul_tensor_read_scope_oracle` probe, whose single-quoted label contains an unescaped `(`. Line 559 before this change, 573 after (= 559 + my 14 lines). The script still *runs* (bash parses commands lazily and that probe is reached), but it is not statically parseable, which will bite the next person who runs `bash -n` in CI.

  Each deserves its own small fix; I left them alone to keep this PR reviewable.
- **Full 100% language-coverage proof** requires both a default and a quantum-enabled build; the release oracle runs that in CI. The build-free gate and the manifest freshness check both pass here.


tsotchke:
## Safety matrix — pasted output

`ESHKOL_ARENA_POISON=1 ./build/eshkol-run -r tests/memory/region_handle_safety_test.esk`

Lines without a `  ok  ` prefix are the runtime echoing a raised message to stderr before the `guard` catches it (`eshkol_runtime_fatal` does that on every catchable condition — the established behaviour the bytevector bounds contract also has). Each is immediately followed by the `ok` for the case that caught it.

```
  ok   open-returns-live-handle
  ok   close-returns-kept-value
  ok   closed-handle-is-not-live
  ok   kept-deep-list-car
  ok   kept-deep-vector
  ok   kept-deep-string
  ok   no-keeps-returns-empty-list
  ok   many-keeps-returns-list-in-order
  ok   named-region-closes
  ok   lone-numeric-arg-is-size-hint
  ok   name-plus-size-closes
region-close: invalid or already-closed region handle
  ok   double-close-raises
  ok   use-after-close-probe-is-false
region-close: invalid or already-closed region handle
  ok   use-after-close-with-keeps-raises
  ok   fabricated-token-not-live
  ok   zero-is-never-a-handle
  ok   negative-is-never-a-handle
region-close: invalid or already-closed region handle
  ok   fabricated-close-raises
  ok   cascade-keep-survives
  ok   cascade-keep-is-deep
  ok   cascade-closed-the-inner-handle
  ok   cascade-closed-the-outer-handle
region-close: invalid or already-closed region handle
  ok   cascaded-inner-close-raises
  ok   lifo-inner-close
  ok   lifo-outer-still-live
  ok   lifo-outer-close
  ok   raise-crossing-region-delivers-value
  ok   raise-crossing-region-value-is-deep
  ok   raise-closed-the-crossed-handle
region-close: invalid or already-closed region handle
  ok   close-after-unwind-raises
  ok   callcc-crossing-region-delivers-value
  ok   callcc-crossing-region-value-is-deep
  ok   callcc-closed-the-crossed-handle
  ok   raise-out-of-with-region-delivers-value
  ok   handles-work-after-with-region-unwind
  ok   reused-slot-rejects-old-token
  ok   reused-slot-accepts-new-token
  ok   reused-slot-closes
region-open: too many open region handles — a region-close is missing
  ok   unclosed-handles-hit-a-clean-limit
  ok   unkept-allocations-do-not-escape
region_handle_safety_test: PASS
```

PR #384, opened 2026-07-25
fix(tensor,ad): literals and gradients are general in rank, extent, member count and arity

tsotchke:
The maintainer's requirement, verbatim: *"i want literals to be n-dimensional, n-members, (i,j)
dimensional with k rank, and variadic"*. This branch was pushed but never proposed; opening it now,
merged with current `master`.

**Merge this via #383, not here.** This branch is one of four in the v1.3.4 tensor cluster
(#359, #364, #365 and this one), and the four collide in the same hunks of
`tensor_codegen.cpp` / `tensor_codegen.h` / `tensor_transformer_codegen.cpp` /
`tensor_linalg_codegen.cpp`. #383 resolves each of those decisions once and carries the full
verification battery for the combination. This PR exists so the branch is proposed, reviewable and
merged-with-master on its own terms; its conflict resolution is **byte-identical** to #383's
(`lib/backend/vm_native.c` and `lib/backend/vm_tensor.c` were compared file-for-file, and no other
cluster branch touches either file).

## What it removes

Four fixed-size limits, each of which failed silently rather than loudly:

- **rank-8 nesting cap** in `vm_tensor_from_nested` — the leftmost spine is now measured first and
  the shape buffer allocated to fit, so rank is the nesting's own depth.
- **`Value args[VM_TENSOR_CTOR_MAX_ARGS]`** (4096) in the variadic `(tensor …)` constructor — it
  clamped the argument walk and **silently dropped every argument past 4096**. The list is now sized
  from the call. The macro had no other user and is gone.
- **`int64_t indices[8]`** in `tensor-ref` / `tensor-set!` and **`int64_t shape[8]`** in `reshape` —
  replaced with `vm_extract_shape_dyn`, which sizes from the index/shape itself. It subsumes the old
  explicit scalar-index fallback (`dims[0] = as_number`) and additionally fixes the *vector*
  spelling, which `vm_extract_shape` had been misreading as a rank-1 shape of 0.
- **fixed-rank scratch buffers** in the VM transpose — now `vm_tensor_dim_scratch(rs, n_dims)`.

`VM_TENSOR_MAX_DIMS` is redefined as what it actually is: the bound on C-stack *scratch* buffers for
the few remaining fixed-rank consumers (model-file I/O, Riemannian-Adam state), which reject a
higher-rank operand explicitly instead of clamping. `VM_TENSOR_INLINE_DIMS` (8) is purely an
allocation fast path.

Master's #360 VM tensor semantics are preserved unchanged: materialized `transpose`, `tensor-data`
as a vector, the catchable bounds contract, and the fixed variadic constructor.

## Verification

Performed on the integrated branch (#383), which contains this branch: `ctest` **113/113**,
`run_vm_parity.sh` **110 passed / 0 failed** across both axes, `run_p8_escape.sh` **PASS** on all
eight axes, plus `run_features_tests` 38/38, `run_ml_tests` 44/44, `run_autodiff_tests` 58/58,
`run_ad_oracle` 60/60.

Gradient boundary, byte-identical native vs VM (`grad_i` of `sum_i (i+1)x_i^2` at `x_i = 1` is
`2(i+1)`, so every index is its own witness):

| params | len | grad[0] | grad[63] | grad[64] | grad[n-1] |
|---|---|---|---|---|---|
| 65 | 65 | 2 | 128 | 130 | 130 |
| 100 | 100 | 2 | 128 | 130 | 200 |
| 1000 (1000 formals, called by name) | 1000 | 2 | 128 | 130 | 2000 |
| 1000 (arity-1 whole-vector loss) | 1000 | 2 | 128 | 130 | 2000 |

`128 → 130` across the 63/64 boundary is the retired 64-element clamp. Rank-33 literals (extent 1)
report rank `33` identically on native and VM.

Note that the 1000-parameter rows are **not** gated by the committed corpus:
`43_gradient_beyond_64_params.esk` covers 65, 100 and two 80-coordinate forms, and
`44_tensor_literal_rank_extent_arity.esk` goes to rank 12 and 5,000 members. The 1000-parameter and
rank-33 cases above were run separately and are reported here rather than assumed.

## Two limits that remain, found while verifying

Both are pre-existing and outside the files this merge resolved; recording them rather than
widening scope.

1. **A 50,000-member flat literal does not load on the VM.** It compiles to 105,128 instructions
   against `ESHKOL_VM_MAX_CODE` (100,000), so `eskb_reader.c:224` rejects it as `invalid ESKB
   payload`. The constant pool is *not* the limit — its ceiling is 4,194,304 and the file carries
   50,893, which is exactly what this branch's pool-growth work fixed. Measured threshold: 47,000
   members → 99,109 instructions, loads and answers correctly; 50,000 → 105,128, does not. The
   emitter writes the oversized ESKB and exits 0, so this is also a fail-open: an artifact its own
   loader refuses. Native handles 50,000 fine.
2. **Native still caps nested-LIST rank at 8.** `(tensor (list (list …)))` deeper than 8 raises
   `tensor: nested list/vector nests deeper than 8 dimensions`
   (`lib/core/runtime_list_helpers.cpp:370`) while the VM accepts rank 33 through the same spelling
   — a native-vs-VM divergence at rank 9+ for the list form. The nested-*vector* spelling is
   uncapped on both substrates.


tsotchke:
Superseded by #383 (merged as 485c4cae): the integration branch contains this PR's full diff, verified by ancestry (ahead_by 0) and content checks. Closing in favor of the integrated merge.

PR #400, opened 2026-07-31
qllm-oracle: exact-AD golden gradient vectors for qLLM geometric primitives

tsotchke:
## What

An exact-gradient oracle for qLLM's geometric primitives, built on the AD tower's exact numerics (f64 + exact rationals, custom-VJP nodes):

- `tests/qllm_oracle/` — exporters that evaluate the geometric primitives (Poincaré projection/retraction, sphere projection/retraction, conformal factors, exp/log maps) and emit golden gradient vectors as JSON, with exact-rational reference values where the tower supports them.
- `scripts/run_qllm_oracle_tests.sh` — one-command runner for the oracle suite.
- Golden vectors under `tests/qllm_oracle/golden/`, vendored downstream by qLLM's `test_geometric_golden_oracle.py` ctest gate (200+ assertions against these values).

## Why

Finite-difference gradient checking is unusable exactly where the geometry gets delicate: at the Poincaré boundary (u → 1) FD invents gradients on the order of 5.8e+05 where the exact derivative is identically zero (artanh clamp), and exp_x is non-invertible in f64 at coordinate-scaled steps near the boundary. The exact-AD tower is the only instrument that can pin these regions, and the golden vectors make that instrument consumable by a repo that doesn't link Eshkol.

## Notes

- Second commit records why the three bridge backwards (embedding, attention, Fréchet mean) are blocked upstream — the work queue for the qLLM bridge's backward coverage.
- One measured finding downstream: torch's `sphere_retract` has a finite identity Jacobian at z=0 (masking), contradicting the oracle's initial prediction; pinned as measured behavior on the consumer side.
- No changes outside `tests/qllm_oracle/`, `scripts/`, and docs.


tsotchke:
Closing as superseded by #392, which carries this branch's entire oracle payload (exporters, runner, tests/qllm_oracle/golden/, README) rebuilt on current master — plus the embedding and Frechet-mean backward passes this branch's second commit records as blocked.

Two reasons this PR cannot merge as-is:

1. The golden vectors here were generated with a build affected by the reverse-mode attribution defect fixed on master (the write-barrier fix). The vectors in #392 were regenerated on the fixed build; values that were correct before the defect came back bit-identical, the rest changed. Vendoring these older vectors would bake wrong gradients into the consumer reference.

2. The branch base predates the current master line, so the diff drags unrelated regressions into .github/workflows, .icc oracle definitions, CMakeLists, CHANGELOG, and the docs tree — merging would revert current master state in all of those.

Nothing is lost: every oracle file lives on in #392 (with provenance for the regeneration documented in tests/qllm_oracle/README.md), and the blocked-backwards analysis is superseded by the implementations. Reopen if anything here turns out not to be covered there.

PR #402, opened 2026-07-31
fix(modules): resolve an import against the libraries this unit defines, on all three back ends

tsotchke:
## The defect

`define-library` validated its library name and threw it away, lowering the whole form to a bare sequence. `import` therefore had nothing to consult but `resolve_module_path()`, a filesystem search that can only ever find a library living in some *other* file, so

```scheme
(define-library (smoke v1_3) (export greet) (begin (define (greet who) (string-append "hi " who))))
(import (smoke v1_3))
(greet "world")
```

reported `Module 'smoke.v1_3' not found` about a library written one line above the import. R7RS-small 5.6.1 defines a library by its `define-library` form and lets the forms that follow import it, so the unit's own libraries have to come first in the resolution order.

This is the readiness probe `define_library_import_works`, one of the two remaining functional blockers on the v1.3.4 gate.

## The fix — native lanes

The frontend now carries a library identity through lowering: every `define-library` ends in a marker `provide` node holding the dotted library name and its export surface. `lib/frontend/library_registry.{h,cpp}` records those markers as they are processed, and both driver lanes consult it before anything else:

1. libraries established earlier in this compilation unit,
2. precompiled stdlib modules,
3. the filesystem search path.

Because a library becomes resolvable by being *processed*, not by living in a particular file, the ordering rule falls out rather than being special-cased: an import above its `define-library` still fails, now with a diagnostic that names the line the library is defined on and cites the rule it broke. Same-unit imports need no load — the body is already spliced into the unit — so their entries are dropped from the require node and only their `prefix` aliases are emitted, at the import's own position. Cross-file imports are untouched: an unregistered name falls through to exactly the search it used before.

## The fix — VM lane

The VM has its own reader and its own module machinery, so the front end being fixed proved nothing there: it knew none of `define-library`, `import` or `export`, and compiled the program above into bytecode that warned about five undefined variables and then died at run time calling a non-function — while the same file ran on the JIT and the AOT lane. A back end silently disagreeing with the other two about whether a program is even well formed is the real divergence, so the VM compiler gains the same resolution order: libraries this unit defined earlier, then the module search path.

`define-library` splices its `begin` bodies into the unit — that *is* its effect on a back end with one global top level — and records its name and the union of its `export` clauses afterwards, so the library is importable below its form and not above it. `export` needs no binding code for the same reason `provide` does not, and `only`/`except` need none either since there is no visibility boundary to enforce; `rename` and `prefix` do introduce names, so they lower to plain `define` aliases exactly as the front end lowers them. `require` consults the same registry, so the two module spellings share one namespace on both back ends.

## Three defects surfaced on the way

* **`(provide …)` emitted nothing at all.** The top-level and body compilers emit an `OP_POP` after any form that bound nothing, so an empty `provide` made that POP discard a *live* value and shift every later binding down a slot — the corruption `compile_form_require()` documents guarding against, reachable through the other module form.

* **The module loader compiled a required file's top-level forms without that POP discipline,** so any module containing a non-defining top-level form desynchronized `n_locals` from the real stack depth for everything after it.

* **An `import` above its `define-library` ran happily on the VM,** because the library body was spliced in below and the import had quietly resolved to nothing. Both native lanes refuse such a program, so the VM refuses it too. That needed a fail-closed signal the VM did not have — its diagnostics are warnings — so a compile-time defect flag now stops both drivers: `compile_and_run()` refuses to execute and the ESKB emitter refuses to write bytecode, which must never leave the process looking like a good artifact.

## Parity evidence

| lane | same-unit import | forward reference |
|---|---|---|
| native JIT (`-r`) | passes, 7/7 checks | refused, exit 1, names the `define-library` line |
| native AOT | passes, 7/7 checks | refused, exit 1, same diagnostic |
| VM (source) | passes, 7/7 checks | refused, exit 1, cites R7RS-small 5.6.1 |
| VM (`--emit-eskb`) | passes | emission refused, no bytecode written |

`tests/vm_parity/corpus/53_define_library_same_unit.esk` puts the contract through the native-vs-VM differential on both VM axes; native and VM output agree line for line.

`tests/vm_parity/PARITY.tsv`'s `op:IMPORT` / `op:PROVIDE` / `op:REQUIRE` rows move from `native-only-justified` to `vm-supported`, each with the behaviour spelled out. No new waiver is added.

## Tests

* `tests/modules/r7rs_define_library_same_unit_test.esk` — 7 execution-backed checks (same-unit import, the union of two `export` clauses, a `prefix` alias, the unprefixed binding, a library importing a sibling of the same unit, and the `require` spelling), globbed by `scripts/run_modules_tests.sh`.
* `tests/vm_parity/corpus/53_define_library_same_unit.esk` — the same contract as a native-vs-VM differential.
* Three CTest entries: `define_library_same_unit_runtime_smoke`, `define_library_same_unit_aot_smoke`, `define_library_same_unit_vm_smoke`.

## Gates on this head

* CTest 134/134
* `scripts/run_vm_parity.sh` 122/122
* `scripts/run_modules_tests.sh` 10/10 (100%)
* ICC smoke probe `define_library_import_works` green

## Docs

`docs/reference/language/modules.md` replaces the note that said same-file imports do not work with the rule that now holds: the resolution order, the forward-reference diagnostic, the import-set modifiers over a same-unit library, and the one thing that is still true — the body is spliced into the unit's top level, so non-exported names are not yet hidden. Strict library isolation stays tracked as module-privacy work.


tsotchke:
## Combined-tree readiness measurement

Measured on a tree carrying **both** v1.3.4 functional blockers — this branch merged with `fix/define-library-same-unit` (#402) — because `failure_free` consumes any FAIL event, so neither can be judged alone. Merge commit `82cc1ff6` (local measurement branch, not pushed); ICC repo `eshkol-v134-both`, indexed at that SHA; traces regenerated from scratch, no stale lines carried over.

```
icc readiness --repo eshkol-v134-both --target eshkol-compiler-readiness \
    --trace-dir scripts/icc_traces
```

**Score `69`, status `blocked`, 37 criteria.**

### Every criterion this PR adds is green

| criterion | evidence |
|---|---|
| `ctest_suite_green` | 136/136 |
| `fixed_point_exact_accumulation_gate` | 5/5 |
| `exact_input_ad_identity_gate` | 4/4 |
| `runtime_closure_arity_spread_gate` | 2/2 |
| `define_library_same_unit_gate` | 3/3 (present once #402 lands; ABSENT and failing without it, by design) |
| `vm_surface_regression_suite` | 55/55 |
| `vm_parity_gate` | 122/122 |
| `event_loop_works` | PASS |

### What is still blocking, and who owns it

Everything that remains is the **axis-6 five-way baseline** and the **`test_isolation.sh` prune bug**, both pre-existing on master and owned by a sibling agent:

* `five_way_surface_agreement` — WARN, and its roll-up `p8_escape_matrix_green` — FAIL. `scripts/p8/five_way_surface.py` reports `640 disagreements (baseline=640), NEW=4, resolved=4`. The four NEW are `native_missing::_region-open`, `native_missing::_region-close-list`, `vm_missing::region-open`, `vm_missing::region-close` — the #370 region rename. A shrink-only ratchet cannot absorb a swap, so the baseline needs re-anchoring. Every other P8 axis is green: axis-1 28/0, axis-2 7/0, axis-3 PASS, axis-4 6/0, axis-5 6/0, axis-7 PASS, axis-8 PASS.
* `language_surface_coverage` — 0 evidence, and the same gap fails ICC architecture invariant `INV-language-surface-exercise` (the other 7 invariants pass). Two causes, both infrastructure: the probe needs a `build-quantum` tree (built here), and `scripts/lib/test_isolation.sh`'s prune sweep dies under `set -e`/`pipefail` when the temp root holds zero `eshkol-test.*` directories (empty glob to `du`, diagnostic swallowed by `2>/dev/null`), which aborts the suite before it emits anything.

### Other evidence on the merged tree

* ICC smoke 54/56 — the two failures are exactly the two items above. `define_library_import_works`, the probe that was the blocker, is **green**.
* ICC architecture-verify 7/8.
* ICC guard-liveness 57 dead / 275 tokens — dead count identical to the post-train master baseline, so no new dead guard.

PR #403, opened 2026-07-31
feat(oracle): make CTest results oracle evidence and wire the v1.3.4 pillars

tsotchke:
## The systemic gap

**No completion-oracle criterion anywhere consumed a CTest result.** The only `ctest` mentions in `.icc/completion-oracles.yaml` were `action:` strings — advice on what to run — and the nine `test_evidence` criteria are index-level ("the tests exist and are runnable"), not execution-backed. The whole 131-test suite therefore had no oracle teeth: a red ctest run could not turn the release gate red, and a pillar could ship with a perfectly good CTest gate while the target that judges the cut never looked at it.

That is exactly what happened this cycle to the fixed-point / i128 engine (5 CTest gates, run by no workflow, no smoke probe and no criterion) and to the exact-input AD tier (4 gates, same). Both shipped into the v1.3.4 cut ungated.

## The root fix — `scripts/run_ctest_gate.sh`

Runs CTest, parses the per-test verdicts, and writes `scripts/icc_traces/ctest_gate.jsonl` in the trace shapes the oracle and the architecture invariants already read:

* `kind: "ctest"` — one event per test, one per named **group**, plus the `ctest_suite_green` roll-up. This is what the `runtime_event` criteria match.
* `kind: "test_result"` — the canonical execution proof ICC's architecture invariants consume.

It prints `PASSED ctest::<name>` node ids so a run is legible without opening the trace, and falls back to parsing the console summary when the host's `ctest` predates `--output-junit`, so an older toolchain never silently reports "no tests".

A **group** is a regex over test names with its own roll-up event, so one criterion gates a whole pillar and stays correct as tests are added to it. **A group whose regex matches no configured test is reported `ABSENT` and fails the gate** — a pillar cannot quietly stop being covered because its tests were renamed or configured out. Adding a pillar means one `CTEST_GATE_GROUPS` line plus one criterion, not another bespoke smoke probe that shells out to ctest and re-runs the same binaries.

Run it the same way the other emitters are run before measuring readiness:

```sh
BUILD_DIR=build ./scripts/run_ctest_gate.sh
```

## Criteria wired under `eshkol-compiler-readiness`

| event | kind | what it gates | status before |
|---|---|---|---|
| `ctest_suite_green` | ctest | the whole suite + every group | nothing consumed ctest at all |
| `fixed_point_exact_accumulation_gate` | ctest | #390, 5 gates | NOT WIRED |
| `exact_input_ad_identity_gate` | ctest | #393, 4 gates | NOT WIRED |
| `runtime_closure_arity_spread_gate` | ctest | #398 retro-catch | NOT WIRED (no ctest entry either) |
| `define_library_same_unit_gate` | ctest | R7RS 5.6.1 same-unit libraries | NOT WIRED |
| `vm_surface_regression_suite` | vm_surface | the self-checking VM half (549-check #394 probe rides here) | NOT WIRED (no emitter either) |
| `vm_parity_gate` | vm_parity | corpus differential on both VM axes | wired under the side `vm-parity` target only |
| `event_loop_works` | eshkol_smoke | ESH-0011 (#399) | criterion existed only under `v1.4-release` |

The event loop and VM parity entries reuse the existing probes and evidence — no new emitter, just the criterion under the target that actually judges the v1.3.4 cut.

## Two emitters were missing, and are built rather than faked

**`tests/autodiff/runtime_closure_arity_spread_test.esk`** was executed only by `scripts/run_autodiff_tests.sh`'s glob, which emits no ICC trace and is referenced by no criterion. It now counts its checks and prints a summary line, and has CTest entries on both native lanes. The counter is load-bearing: the pre-fix build does not print a `FAIL`, it *dies at check 19*, so a gate that only looked for `FAIL:` would read a truncated run as green.

**`scripts/run_vm_surface_tests.sh`** now emits `kind: "vm_surface"` events — one per probe plus a suite roll-up — and runs every probe instead of stopping at the first failure; one defect reported while the rest stay hidden is not evidence. This is the half of the VM surface that asserts against R7RS *inside a single run*, so unlike the parity differential a defect **shared** by native and the VM cannot pass by agreement.

## `define_library_same_unit_gate` is ABSENT on this branch

Deliberately. Its CTest entries land with the same-unit `define-library` fix (#402), the other functional blocker in this readiness cut. The gate reporting ABSENT rather than silently passing is the behaviour this PR exists to add. The two branches are measured together; readiness numbers on the merged tree are in the PR discussion.

## Evidence on this branch

* CTest 131/131; the gate reports `fixed_point_exact_accumulation_gate` 5/5 and `exact_input_ad_identity_gate` 4/4 green, and correctly flags the two then-absent groups.
* `scripts/run_vm_surface_tests.sh` 55/55, trace written with 56 events.
* `docs/TESTING.md` documents the gate, its trace shapes, and how to gate a new pillar.

## Not in scope

The qLLM oracle block from the wiring audit is not applied here: neither `scripts/run_qllm_oracle_tests.sh` nor `tests/qllm_oracle/` exists on master. It lands with #392.

No CI lane runs the ctest gate yet — that is the standing "no lane runs ctest anywhere" gap, and it belongs with the post-tag readiness-gate arming rather than with this wiring.


tsotchke:
## Combined-tree readiness measurement

Measured on a tree carrying **both** v1.3.4 functional blockers — this branch merged with `fix/define-library-same-unit` (#402) — because `failure_free` consumes any FAIL event, so neither can be judged alone. Merge commit `82cc1ff6` (local measurement branch, not pushed); ICC repo `eshkol-v134-both`, indexed at that SHA; traces regenerated from scratch, no stale lines carried over.

```
icc readiness --repo eshkol-v134-both --target eshkol-compiler-readiness \
    --trace-dir scripts/icc_traces
```

**Score `69`, status `blocked`, 37 criteria.**

### Every criterion this PR adds is green

| criterion | evidence |
|---|---|
| `ctest_suite_green` | 136/136 |
| `fixed_point_exact_accumulation_gate` | 5/5 |
| `exact_input_ad_identity_gate` | 4/4 |
| `runtime_closure_arity_spread_gate` | 2/2 |
| `define_library_same_unit_gate` | 3/3 (present once #402 lands; ABSENT and failing without it, by design) |
| `vm_surface_regression_suite` | 55/55 |
| `vm_parity_gate` | 122/122 |
| `event_loop_works` | PASS |

### What is still blocking, and who owns it

Everything that remains is the **axis-6 five-way baseline** and the **`test_isolation.sh` prune bug**, both pre-existing on master and owned by a sibling agent:

* `five_way_surface_agreement` — WARN, and its roll-up `p8_escape_matrix_green` — FAIL. `scripts/p8/five_way_surface.py` reports `640 disagreements (baseline=640), NEW=4, resolved=4`. The four NEW are `native_missing::_region-open`, `native_missing::_region-close-list`, `vm_missing::region-open`, `vm_missing::region-close` — the #370 region rename. A shrink-only ratchet cannot absorb a swap, so the baseline needs re-anchoring. Every other P8 axis is green: axis-1 28/0, axis-2 7/0, axis-3 PASS, axis-4 6/0, axis-5 6/0, axis-7 PASS, axis-8 PASS.
* `language_surface_coverage` — 0 evidence, and the same gap fails ICC architecture invariant `INV-language-surface-exercise` (the other 7 invariants pass). Two causes, both infrastructure: the probe needs a `build-quantum` tree (built here), and `scripts/lib/test_isolation.sh`'s prune sweep dies under `set -e`/`pipefail` when the temp root holds zero `eshkol-test.*` directories (empty glob to `du`, diagnostic swallowed by `2>/dev/null`), which aborts the suite before it emits anything.

### Other evidence on the merged tree

* ICC smoke 54/56 — the two failures are exactly the two items above. `define_library_import_works`, the probe that was the blocker, is **green**.
* ICC architecture-verify 7/8.
* ICC guard-liveness 57 dead / 275 tokens — dead count identical to the post-train master baseline, so no new dead guard.

PR #421, opened 2026-08-12
chore(oracle): gate readiness on the silent-wrong flaw ledger

tsotchke:
## Why

The ledger itself already landed on master via the closure PRs (#423, #424, #427).
What is still missing is the part that makes it **bind**: nothing reads the ledger, so
nothing stops a release while entries are open.

Every other criterion in the release oracle certifies that something WORKS. None can
certify that nothing silently LIES. A defect that returns a wrong value with no
diagnostic and exit 0 is structurally invisible to a probe that only asserts its own
happy path — which is how a cut can reach readiness 100 while carrying an AD operator
that differentiates the shadowed global, a `map` that ignores a shadowing binding, and
seven documented resource limits that are never enforced.

**This PR enforces. It fixes nothing and it changes no ledger entry.**

## What (3 files, +345/-0)

- **`scripts/gate_no_silent_wrong.py`** — grades `.icc/silent-wrong-ledger.yaml` and
  emits the `no_open_silent_wrong` runtime event. **Fails closed**: a missing,
  unparseable or schema-invalid ledger is FAIL, because absence of the ledger is not
  evidence of absence of the defects. An `IN-FLIGHT` entry tagged `silent_wrong_class`
  still gates — a merged PR is not a closed entry until somebody re-measures.
- **`.icc/completion-oracles.yaml`** — the criterion wired into both
  `eshkol-compiler-readiness` and `v1.3-evolve` at severity high.
- **`scripts/run_icc_smoke.sh`** — the probe wired in so the event lands in the ordinary
  trace bundle.

## Terminal statuses

The grader understands the two statuses the closure PRs introduced, `fixed` and
`guarded`, rather than forcing those entries back into a vocabulary that does not fit.
They are not decoration: **`guarded` means the wrong VALUE is gone but the operation
still cannot compute the right one**, which is a different promise from `fixed`.

Nothing closes by bare assertion — a terminal entry must carry a re-measurement SHA
**or** concrete evidence. The four entries currently closed without a SHA (`SW-06`,
`SW-09`, `SW-09b`, `SW-22`) are printed in the gate's own output under
`CLOSED WITHOUT A RE-MEASUREMENT SHA`, so the weaker form stays visible instead of
quietly becoming the norm.

## Current grade against master's ledger at `019ab413`

**FAIL — 23 blocking, 5 closed, 0 invalid.** Red by design.

Blocking: `SW-01 SW-02 SW-03 SW-04 SW-05 SW-07 SW-08 SW-10 SW-11 SW-12 SW-13 SW-14
SW-18 SW-19 SW-20 SW-21 SW-23 SW-24 IF-01 IF-02 IF-03 IF-04 IF-05`

Two things a reviewer should know:

1. **`IF-01`..`IF-05` block even though #420, #418, #419 and #413 are merged.** The
   fixes landed; nobody re-measured and closed the entries. That is the gate working as
   designed, and those five are the cheapest wins available — measure at `019ab413` and
   close. I did not close them by assertion.
2. **With #425 applied the count drops to 21** (verified by cherry-picking #425's commit
   onto master: applies cleanly, 58 insertions / 5 deletions, exactly the `SW-07`/`SW-08`
   closure).

## Rebase note

Rebased onto `019ab413`. Master had moved 9 commits and had absorbed the ledger, so the
ledger conflict resolved as **union with master winning** — master's file is a strict
superset (64 entries vs my 62, adding `SW-09b` and `SW-27`) with no entry lost, and it
is byte-identical in this branch. The seven ledger-only commits from the original branch
became empty and were dropped; their findings are already on master inside the ledger.

Master also **corrected one of my inferences** and is right: the `37_sort.esk`
cached-vs-nocache exit split (1 vs 139) was *not* the IF-01 object-cache signature — it
was a use-after-free in `ReplJITContext::addModule` destroying the rejected module's
`LLVMContext` before the `Module`. That correction is recorded on master at `a1c28375`.

### DIFF
--- a/.icc/completion-oracles.yaml
+++ b/.icc/completion-oracles.yaml
@@ -56,6 +56,32 @@ oracles:
         label: Codegen / runtime / type-checker production paths are real
         action: wire_real_path_or_classify_stub
 
+      # ── Silent-wrong flaw gate (2026-08-12) ──────────────────────────
+      # Every other criterion in this oracle certifies that something
+      # WORKS. None of them can certify that nothing SILENTLY LIES: a
+      # defect that returns a wrong value with no diagnostic and exit 0
+      # is invisible to a probe that only asserts its own happy path.
+      # That is exactly how a release cut reached readiness 100 while
+      # carrying an AD operator that differentiates the wrong function,
+      # a `map` that ignores a shadowing binding, and seven documented
+      # resource limits that are never enforced.
+      #
+      # `.icc/silent-wrong-ledger.yaml` is the enumeration of that class;
+      # `scripts/gate_no_silent_wrong.py` grades it and emits this event.
+      # The gate fails closed — a missing or unparseable ledger is FAIL,
+      # because absence of the ledger is not evidence of absence of the
+      # defects. An entry leaves the gate only via a terminal status
+      # (closed / fixed / guarded) backed by a re-measurement SHA or a
+      # committed test, or via a waiver with an owner, a reason and an
+      # unexpired date. Nothing closes by bare assertion.
+      - runtime_event:
+          event_kinds: [eshkol_smoke]
+          event_names: ["no_open_silent_wrong"]
+          event_values: ["PASS"]
+        severity: high
+        label: 'No open, unwaived SILENT-WRONG flaw ships: every entry in .icc/silent-wrong-ledger.yaml whose bucket is SILENT-WRONG (or which is IN-FLIGHT but silent_wrong_class) is closed/fixed/guarded with evidence, or carries an owner/reason/unexpired waiver — wrong values, wrong derivatives and wrong memory outcomes with no diagnostic and exit 0 are tag-blocking'
+        action: python3 scripts/gate_no_silent_wrong.py
+
       - runtime_event:
           event_kinds: [eshkol_smoke]
           event_names: ["llvm_module_verifier_clean"]
@@ -630,6 +656,18 @@ oracles:
   - name: v1.3-evolve
     aliases: [v1.3, release-1.3, evolve]
     requires:
+      # The release-line mirror of the compiler-readiness silent-wrong gate.
+      # A v1.3.x tag is a promise to consumers about answers, not about
+      # features, so the flaw ledger gates the release target too. See the
+      # long comment on the same criterion in eshkol-compiler-readiness.
+      - runtime_event:
+          event_kinds: [eshkol_smoke]
+          event_names: ["no_open_silent_wrong"]
+          event_values: ["PASS"]
+        severity: high
+        label: 'No open, unwaived SILENT-WRONG flaw ships in the v1.3 line (.icc/silent-wrong-ledger.yaml graded by scripts/gate_no_silent_wrong.py; fails closed)'
+        action: python3 scripts/gate_no_silent_wrong.py
+
       - runtime_event:
           event_kinds: [eshkol_smoke]
           event_names: ["string_interpolation_works"]
--- a/.icc/silent-wrong-ledger.yaml
+++ b/.icc/silent-wrong-ledger.yaml
@@ -485,6 +485,15 @@ entries:
   - id: SW-22
     bucket: SILENT-WRONG
     status: guarded
+    closed_at: "019ab413"
+    remeasured: >-
+      Independently re-measured at 019ab413. tests/xla/xla_codegen_test.cpp
+      required its own -DESHKOL_XLA_ENABLED=ON build (build-xla), which is why
+      this entry lagged the other three; built and run here, the suite is 11/11
+      passed / 0 failed, exit 0, with both of the rows that matter green:
+      "XLA Broadcast Compatible Shapes... PASS" and "XLA Broadcast Incompatible
+      Shapes... PASS". The runtime entry point now rejects incompatible shapes
+      instead of broadcasting them into a wrong answer.
     title: "eshkol_xla_broadcast now validates broadcast compatibility at its runtime entry point"
     guarded_by: "branch fix/silent-to-loud-sw06-09-22 (PR converting SW-06/SW-09/SW-22)"
     guard: >-
--- a/.icc/silent-wrong-ledger.yaml
+++ b/.icc/silent-wrong-ledger.yaml
@@ -151,26 +151,79 @@ entries:
 
   - id: SW-07
     bucket: SILENT-WRONG
-    status: open
+    status: closed
+    closed_at: "fc3c1257"
     title: "fg-marginal returns an empty result natively"
     repro: "(define fg (make-factor-graph 2 2)) (display (fg-marginal fg 0))"
     observed: "native (), VM #(0.5 0.5)"
     correct: "#(0.5 0.5) — the uniform marginal"
-    evidence: ".scratch/flaw-ledger/repro/sw_fg_marginal.esk; PR #413 Follow-ups item 2"
+    measured: >-
+      At fc3c1257 (PR #424, branch fix/fg-marginal-rationalize) both engines
+      print #(0.5 0.5) for all three dims spellings #(2 2), '(2 2) and 2, and
+      for both variables of a two-variable graph. Four defects, not one: the
+      native result tensor's dimensions[] array was never written
+      (lib/core/logic_builtins.cpp:52, so the documented #(2 2) spelling printed
+      #() around a correct distribution); native fg_read_numeric() read no list
+      form and eshkol_make_factor_graph_tagged() had no bare-scalar arm
+      (lib/core/inference.cpp), so '(2 2) and 2 answered the tagged NULL; and
+      the VM matched HEAP_TENSOR only, so a #(2 2) VECTOR literal fell to the
+      scalar arm and built one zero-state variable, while a real scalar was read
+      as a one-element dims list that clamped num_vars to 1
+      (lib/backend/vm_native.c native call 520).
+    evidence: >-
+      tests/vm_parity/corpus/57_factor_graph_marginal.esk (both engines, with an
+      explicit uniform-prior and extent oracle);
+      tests/differential/corpus/44_factor_graph_marginal.esk (native axes);
+      .scratch/flaw-ledger/repro/sw_fg_marginal.esk; PR #413 Follow-ups item 2
     caught_by: [pr-record-413, direct-measurement]
     missed_by: [icc-correctness-chain, icc-smoke, icc-readiness, KNOWN_ISSUES]
-    notes: "PR #413 explicitly leaves this alone. #413 is still OPEN, so even its factor-graph fixes are not in the cut."
+    fixed_by: "PR #424 (fix/fg-marginal-rationalize)"
+    notes: >-
+      PR #413 has since MERGED and its unified factor-graph sequence reader
+      (vm_fg_seq_doubles) supersedes #424's original VM vector arm; #424 was
+      rebased onto it and now layers only the bare-scalar broadcast on top, so
+      `(make-factor-graph 2 2)` is a two-variable graph on both engines rather
+      than being clamped to one. #424 does NOT close
+      tests/logic/workspace_test.esk, which is a ws-step! defect and is still
+      divergent at fc3c1257. #413's merge already removed
+      tests/logic/inference_test.esk and continuous_learning.esk from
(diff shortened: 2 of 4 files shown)


tsotchke:
## Update after two further measurement passes

**1. The correctness chain figure in the description was too harsh, in a way that
matters.** `icc contradictions` returns **30** findings at `9f2da2ab`, not zero. But
every one is a `constant_disagreement` or `env_default_disagreement` on a build or
tuning symbol (`HEAP_SIZE`, `MAX_FRAMES`, `PATH_MAX`, `BUILD_DIR`, `ESHKOL_TIMEOUT_MS`,
…). The detector compares symbol values, never behaviour, documentation or test-harness
metadata. It flags that `ESHKOL_TIMEOUT_MS` has two different defaults and misses that
the variable does nothing. Still 0 of the ledgered items, but for a specific and
fixable reason. Commit `6cafb349`.

**2. Every filed reproducer under `tests/vm_parity/found/` was re-run on both engines.**
38 programs: **25 still diverge, 13 now AGREE and print the correct value.**

Three of the thirteen had been carried into this ledger as open SILENT-WRONG entries on
the strength of their filed headers alone, and are now closed by measurement:

- `tensor_set_oob_silent_native` — both engines print `caught` twice; native raises. The
  header still claims the write silently vanishes.
- `tensor_ref_component_oob_native` — both print `caught`.
- `tensor_nested_collection_native` — both build `#((1 2) (3 4))` with shape `(2 2)`. The
  header still claims native flattens the outer level and zero-fills.

`found/` is documented as evidence rather than gate input, so no harness re-runs it. A
filed reproducer that goes stale in silence is the mirror image of a silent defect: it
makes the project look worse than it is, and it trains readers to discount the 25 that
are still real. Recorded as `DD-12`; re-running `found/` belongs in the parity gate.
Commit `88c06831`.

**Gate now reports FAIL with 25 blocking entries** (20 open SILENT-WRONG + 5 IN-FLIGHT
silent-wrong class), 3 closed, 0 invalid — replacing the 28 in the description.

**ICC readiness at this SHA is not reported here.** The smoke harness stalls on
`surface_parity_execution_backed` past the session budget, and `language_surface_coverage_floor`
is the documented prerequisite false-red that needs a `build-quantum` runner. A readiness
score computed from an incomplete trace bundle would be exactly the false-green this PR
exists to prevent, so it is left unstated rather than guessed.

tsotchke:
## Full ICC chain completed — two more gate-integrity findings

The harness finished. Results at `9f2da2ab`:

| Gate | Result |
| --- | --- |
| `scripts/run_icc_smoke.sh` | **57/58** — the one failure is `language_surface_coverage_floor` (exit 2), the documented prerequisite that needs a `build-quantum` runner |
| `scripts/run_vm_parity.sh` | 144 PASSED, 0 FAILED |
| `scripts/run_vm_surface_tests.sh` | 56/56 |
| `icc architecture-verify` | 8 PASS / 1 FAIL — the FAIL is `INV-language-surface-exercise`, the same prerequisite |
| `icc chain --preset correctness` | 30 `contradictions` (all constant/env-default) + 25 `duplicated-code` entities — **0 of 61 ledgered items** |

Two of the probes that PASSED are worth reading closely, and are now ledgered.

### PR-10 — the parity gates grade non-growth, not correctness

```
engine_semantic_parity  113/1114 constructs with differential evidence (10.14%),
                        10 divergent program(s), 0 new          -> PASS
surface_parity_probe    2506 probed / 318 divergences / 0 new   -> PASS
```

**1001 constructs (89.86%) have no differential evidence at all**, and both probes are
green because nothing is *new*. A defect present before the baseline is permanently
invisible. SW-01, SW-02, SW-06 and SW-08 all live in that unmeasured remainder — this is
the structural reason they escaped every gate. The gate needs a rising floor on
constructs-with-evidence, not only a ceiling on new divergences. Commit `a26dca51`.

### PR-11 — the invariant named for silent wrong answers is an existence check

`INV-engine-semantic-parity` is severity **critical** and its incident tag is literally
`vm-parity-silent-wrong-answer`. It grades PASS on:

```
capability 'engine_semantic_parity' exercised within 30d (1 event(s))
```

An `exercise` invariant proves a probe RAN. It cannot prove the probe found anything,
and it would pass identically at 1% coverage. Severity `critical` on an existence check
reads as far stronger assurance than it delivers. Pair every exercise invariant with a
threshold on its own payload; same shape as PR-10, fix them together. Commit `c4c137eb`.

### Also tightened

`SW-10`'s evidence is now exhaustive rather than a spot check: grepping `lib/`, `inc/`
and `exe/` for the four resource-limit enforcement entry points returns 15 hits — 11
declarations in `inc/eshkol/core/resource_limits.h` and 4 definitions in
`lib/core/resource_limits.cpp`. **Zero call sites anywhere else.** The functions are
written, compiled, and never invoked. Commit `fabbc640`.

Ledger is now 61 entries; the gate still reports **FAIL, 25 blocking**.

tsotchke:
## Full gate chain complete — readiness is 77, and LE-01 is the whole of the blocker

Everything rebuilt (all test targets + a quantum-enabled runner) and re-run at `9f2da2ab`:

| Gate | Result |
| --- | --- |
| `run_ctest_gate.sh` | **142/142 PASS** |
| `run_vm_parity.sh` | 144 PASSED / 0 FAILED |
| `run_vm_surface_tests.sh` | 56/56 |
| `run_language_coverage.sh` | **1091 constructs, 100.00%**, deficit 0, high-risk uncovered 0 — PASS |
| `icc architecture-verify` | 8 PASS / 1 FAIL |
| `icc readiness` | **BLOCKED, score 77** |

### What blocks it

`failure_free`. Its evidence is **LE-01 almost in its entirety** —
`tests/differential/corpus/37_sort.esk` fails a *different way on every compilation axis*:

| Axis | Exit |
| --- | --- |
| jit | 1 |
| jit-nocache | **139 (SIGSEGV)** |
| aot-o0 | 125 |
| aot-o2 | 125 |

Six failing pairwise axis comparisons plus the roll-up `1/41 corpus programs diverge
across axes`. The cached-JIT-vs-nocache split (1 vs 139) is the same object-cache
signature as IF-01, so those two are probably **one root cause seen from two directions**.

This is the honest detector working: readiness is red *because the defect is real*.
**Do not baseline `37_sort.esk` to make the number go green.** Commit `57c7965f`.

### One stale event, deliberately not edited

The only other FAIL in the bundle is a `language_surface_coverage_floor` event from
before `build-quantum` existed. The underlying gate now measures **PASS at 1091/100.00%**.
I left the stale event in place rather than hand-correcting the trace file — editing an
evidence file is precisely what this ledger exists to prevent, and the tooling correctly
refused the edit when I tried. It clears on the next full smoke run.

### The juxtaposition this whole exercise comes down to

> **Language-surface coverage is 100.00% — 1091 of 1091 constructs executed.**
> **Engine differential coverage is 10.14% — 113 of 1114 constructs compared.**

Executing a construct proves it *runs*. Comparing it across engines proves it is *right*.
The project measures the first exhaustively, the second barely, and reports both green.
That gap is where SW-01, SW-02, SW-06, SW-08 and SW-24 have been living.

Ledger is 62 entries; gate reports **FAIL, 26 blocking**.

tsotchke:
## Two closure commits added — gate goes 23 → **16 blocking**

### 1. `cebb4051` — #425 absorbed

Cherry-picked `5610a143` clean (author/committer already `tsotchke <tsotchkele@gmail.com>`).
`SW-07` and `SW-08` are now `closed` at `fc3c1257`. **#425 can be closed as absorbed.**

### 2. `dbde0b2b` — IF-01..IF-05 closed by measurement at `019ab413`

Each reproducer re-run on a fresh RelWithDebInfo LLVM-21 build of master. Not read off the
merge log — **all five genuinely pass**:

| Entry | Fix | What I measured |
| --- | --- | --- |
| IF-01 | #420 / `6dcdd168` | `ad_capture_global_shadow` **19/19** in all three configs that used to disagree: JIT with the object cache **ON** (where the miscompile lived), JIT with `ESHKOL_JIT_CACHE=0`, and **AOT — exits 0 instead of segfaulting**. Differential corpus 42 byte-identical across jit-cached / jit-nocache / aot-O2. |
| IF-02 | #418 / `9b3fd14d` | corpus 54 byte-identical native vs VM; SRFI-1 `(take lst n)` on both, including over-long-count rows. |
| IF-03 | #419 / `a83b309e` | `complex_transcendentals_test` end to end, incl. the two rows naming this defect ("uses its value, not its pointer"). `(sqrt -4)`=2i, `(exp i*pi)`=-1+1.22e-16i, `(log -1)`=pi*i. |
| IF-04 | #419 / `a83b309e` | `(inexact->exact 0.1)` = `3602879701896397/36028797018963968`, the exact IEEE double. The pre-fix answer was **wrong by a factor of 8 in both terms** and didn't round-trip. It round-trips now. |
| IF-05 | #413 / `f1ded1ed` | corpus 56 — the gate whose whole point is that free energy must MOVE with the evidence — byte-identical on both engines, 26 assertions all `#t`. |

### CLOSED-WITHOUT-SHA section: 4 → 1

Added re-measurement SHAs to `SW-06`, `SW-09`, `SW-09b` after re-running each: diff-on-quoted
raises `caught` on **both** engines, i128 `+` raises on the VM, and all **eleven** remaining
i128 operations raise. Section kept as a warning, not hard-blocking, per your call.

**`SW-22` is the only one left** — its evidence is a C++ test needing a separate
`-DESHKOL_XLA_ENABLED` build, which is running now. I'll add its SHA if it lands cleanly;
if not, it stays as a one-line warning rather than a guess.

### Three things I deliberately did NOT do

1. **IF-03's closure does not absorb the residual display divergence** — native prints `+2i`,
   VM prints `0+2i`. That's `PR-08` and it stays open on its own terms. Folding a cosmetic
   divergence into a value-defect closure is how ratchets rot.
2. **No waivers added** — those land via #422 as you said.
3. **`measured_at_sha` was NOT overwritten.** I added `remeasured_at_sha: 019ab413` with an
   explicit scope note saying only the eight named entries were re-run and every other
   entry's evidence still dates from `9f2da2ab`. A blanket "current at `019ab413`" is exactly
   the kind of claim this file exists to stop.

### Gate at `019ab413`: FAIL — 16 blocking, 12 closed, 0 invalid

```
SW-01 SW-02 SW-03 SW-04 SW-05 SW-10 SW-11 SW-12 SW-13 SW-14
SW-18 SW-19 SW-20 SW-21 SW-23 SW-24
```

Note the shape of what's left: **`SW-01`, `SW-02` and `SW-24` are one defect family** (name
resolution ignoring shadowing, on both engines in three subsystems), and `SW-03`/`SW-04`/`SW-05`
are the AD-nesting cluster. Six of the sixteen are really two root causes.

tsotchke:
## SW-22 landed — the CLOSED-WITHOUT-SHA section is now **empty**

`42a4da36`. SW-22 was the last entry closed on committed-test evidence with no SHA a
reader could look up. Its test needed its own `-DESHKOL_XLA_ENABLED=ON` build, which is
why it lagged the other three.

Built and run at `019ab413` — `tests/xla/xla_codegen_test.cpp`, **11 passed / 0 failed,
exit 0**, with both rows that matter green:

```
Test: XLA Broadcast Compatible Shapes... PASS
Test: XLA Broadcast Incompatible Shapes... PASS
```

The runtime entry point rejects incompatible shapes instead of broadcasting them into a
wrong answer.

**Every terminal entry in the ledger now carries a `closed_at`** — 12 of 12. The section
emptied *honestly, by measurement*, not by dropping the check. The check stays in the
grader on purpose: it is how the next entry closed on a bare test reference becomes
visible instead of blending in.

### Final state

| | |
| --- | --- |
| Gate | **FAIL — 16 blocking, 12 closed, 0 invalid** |
| Closed without a SHA | **NONE** |
| Branch | 4 commits, `%ae`/`%ce` = `tsotchke <tsotchkele@gmail.com>` |
| #421 | **MERGEABLE** |
| #425 | absorbed by `cebb4051` — safe to close |

Blocking: `SW-01 SW-02 SW-03 SW-04 SW-05 SW-10 SW-11 SW-12 SW-13 SW-14 SW-18 SW-19 SW-20
SW-21 SW-23 SW-24` — of which SW-01/02/24 are one shadowing family and SW-03/04/05 are the
AD-nesting cluster, so sixteen entries are closer to twelve distinct problems.

PR #423, opened 2026-08-12
fix(honesty): convert three silent-wrong paths to loud errors (SW-06, SW-09/SW-09b, SW-22)

tsotchke:
## Summary

Maintainer-ordered pre-tag scope: full feature work for these three
defects is deferred to v1.3.5, but shipping a silently wrong answer is
forbidden while shipping a clear error is acceptable. This PR converts
three entries from the skipped-flaws ledger (`.icc/silent-wrong-ledger.yaml`)
to an honest failure mode without implementing the underlying deferred
features.

- **SW-06** — `(diff 'expr 'x)` on a quoted s-expression silently returned
  a fabricated derivative of `0` on both engines. Real symbolic
  differentiation of quoted data is not implemented (v1.3.5 scope); both
  engines now raise a guard-catchable exception instead.
  - Native: `codegenDiff()` in `lib/backend/llvm_codegen.cpp` raises via
    `eshkol_raise()` / `eshkol_make_exception_with_header()` when the
    diff expression is a top-level quote, mirroring the existing
    non-exhaustive-`match` diagnostic pattern.
  - VM: native call 393 (shared by `derivative` and `diff`) in
    `lib/backend/vm_native.c` raises via `vm_raise_error_msg()` when the
    first argument isn't a callable closure.
  - The unquoted, documented form (`(diff (* x x) x)`) is untouched and
    still returns the correct symbolic derivative.

- **SW-09** — generic `+` over `i128` values silently answered `0` on the
  VM (native already raised a type error — LE-03). No i128 opcodes are
  implemented (unchanged, v1.3.5 scope); the VM's two independent
  `OP_ADD` implementations (switch-based and threaded/computed-goto, both
  in `lib/backend/vm_run.c`) now raise instead of silently coercing an
  i128 operand to `0.0` via `as_number_vm()`. Generic `*` over i128 has
  the same shape of bug and remains open — only the literal SW-09 repro
  (`+`) is converted here; noted in `docs/KNOWN_ISSUES.md`.

- **SW-22** — `eshkol_xla_broadcast()` performed no broadcast-compatibility
  validation, so an incompatible source dimension narrower than the
  target read past the end of the source buffer — a memory-safety
  defect, not merely a wrong answer. Adds a NumPy-style compatibility
  guard (mirroring `XLATypes::broadcastShape()`) that rejects
  incompatible shapes before any stride computation. Marked `guarded`
  rather than `closed` in the ledger: the XLA broadcast bridge has no
  live caller yet (`ESHKOL_XLA_ENABLED` defaults OFF, `emitBroadcast()`
  is unwired into codegen), so this closes the memory-safety defect
  without yet having an end-to-end user-facing diagnostic.

## Test plan

- [x] `ctest --test-dir build`: **142/142 passed**
- [x] `scripts/run_vm_parity.sh`: **144 passed, 0 failed**
- [x] `scripts/run_error_handling_tests.sh`: **7/7 passed**
- [x] Differential corpus (`scripts/run_differential.sh tests/differential/corpus/`):
      **246/252 axis pairs agree across 42 programs**; sole divergence is
      the pre-existing `37_sort.esk` (LE-01), unrelated and unchanged —
      confirmed identical on unmodified master (240/246 across 41 programs).
- [x] New differential entry `tests/differential/corpus/42_diff_quoted_raises.esk`
      agrees on all 4 native axes (jit / jit-nocache / aot-o0 / aot-o2).
- [x] VM standalone self-test (`eshkol-vm-standalone-test`) additions —
      `diff-quoted-expr-raises`, `derivative-non-callable-raises`,
      `i128-add-raises` — **62/62 source tests passing**.
- [x] XLA codegen test (`-DESHKOL_XLA_ENABLED=ON`) additions —
      `test_xla_broadcast_compatible`, `test_xla_broadcast_incompatible` —
      **11/11 passing**.
- [x] All three fixes **revert-validated**: temporarily undoing each guard
      reproduces the original silent-wrong/OOB behavior against both this
      branch and an unmodified baseline build, confirming the new tests
      are load-bearing.
- [x] ICC `architecture-verify`: **8/9 PASS** (`INV-engine-semantic-parity`
      now passes after regenerating evidence via
      `run_engine_parity_coverage.py`, 195 programs / 1114 constructs, 10
      pre-existing divergences matching the `PARITY-RATCHET` baseline, 0
      regressed). The one remaining FAIL, `INV-language-surface-exercise`,
      needs a `build-quantum` runner this environment doesn't have — a
      documented pre-existing gap, unrelated to this change.
- [x] ICC `guard-diff` / `gate-semantics-audit` / `pre-commit-check`: all
      **PASS**, 0 findings, scoped to `lib`/`inc`/`exe`.

Not run to completion: the full `scripts/run_icc_smoke.sh` (58-probe
release harness) was started but did not finish within the session —
it reached probe 44/58 and then stalled on `surface_parity_execution_backed`,
matching previously documented behavior for that probe on this machine
under concurrent load. This is a release-level gate, not scoped to this
3-file PR; the targeted path-scoped ICC chain above covers the change.

### DIFF
--- a/.icc/silent-wrong-ledger.yaml
+++ b/.icc/silent-wrong-ledger.yaml
@@ -189,16 +189,74 @@ entries:
       docs/KNOWN_ISSUES.md); this PR only stops the VM from silently
       coercing an i128 operand to 0.0 via as_number_vm(). Native is
       unchanged (LE-03 already raised "Type error in +: expected number,
-      vector, or tensor, got heap-object"). NOTE: generic `*` over i128 is
-      the SAME shape of bug and remains open/unfixed on the VM — only `+`
-      (the literal SW-09 repro) was converted here; see
-      docs/KNOWN_ISSUES.md's i128 entry.
-    fixed_by: "branch fix/silent-to-loud-sw06-09-22 (PR converting SW-06/SW-09/SW-22)"
+      vector, or tensor, got heap-object"). SW-09b (below), landed in the
+      same PR before merge, converts the rest of the family (`- * / modulo`,
+      unary `-`, `abs`, `= < > <= >=`) the same way — the family is no
+      longer split between "+ fixed" and "everything else open".
+    fixed_by: "branch fix/silent-to-loud-sw06-09-22 (PR converting SW-06/SW-09/SW-22, SW-09b same PR)"
     evidence: >-
       lib/backend/vm_tests.c source_test_expect "i128-add-raises" (VM
-      standalone self-test, 62/62 passing). Revert-validated: reverting the
+      standalone self-test, 73/73 passing). Revert-validated: reverting the
       guard reproduces the original "no" (silent 0, uncaught) output on the
       pre-fix VM binary.
+
+  - id: SW-09b
+    bucket: LOUD-ERROR
+    subtag: cheap-fix
+    status: fixed
+    reclassified_from: SILENT-WRONG
+    title: >-
+      The rest of the generic arithmetic/comparison family over i128 (-, *,
+      /, modulo, unary -, abs, =, <, >, <=, >=) now raises on the VM, same
+      as SW-09's +
+    repro: >-
+      (define a (i128 5)) (define b (i128 7)) (display (guard (e (#t
+      'caught)) (* a b) 'no)) — and the same shape for -, /, modulo, (- a),
+      (abs a), =, <, >, <=, >=.
+    observed_before: >-
+      Every one of these opcodes silently answered a fabricated value with
+      exit 0 on the VM (e.g. `*` printed 'no', meaning the guard body ran to
+      completion and returned its non-exceptional value, exactly like SW-09
+      before its fix). Confirmed by direct measurement: not caught by the
+      original SW-09 fix, which converted only `+`.
+    observed_after: >-
+      All twelve opcodes now raise a catchable "i128 arithmetic/comparison
+      is not supported on the VM" error: `+ - * / modulo` (binary
+      arithmetic), unary `-` and `abs` (unary arithmetic), `= < > <= >=`
+      (comparison). Fixed in BOTH of vm_run.c's independent dispatch
+      implementations — the threaded/computed-goto loop (lbl_ADD, lbl_SUB,
+      lbl_MUL, lbl_DIV, lbl_MOD, lbl_NEG, lbl_ABS, lbl_EQ, lbl_LT, lbl_GT,
+      lbl_LE, lbl_GE) and the switch-based fallback compiled only on
+      non-GCC/non-Clang toolchains (OP_ADD, OP_SUB, OP_MUL, OP_DIV, OP_MOD,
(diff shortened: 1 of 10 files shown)


tsotchke:
## Follow-up: SW-09 closed family-wide (SW-09b)

The initial commit converted only generic \`+\` over i128 on the VM (the
literal SW-09 repro), leaving \`- * / modulo\`, unary \`-\`, \`abs\`, and
\`= < > <= >=\` open with a "remains open" \`KNOWN_ISSUES.md\` note. That
was itself a silent-wrong shipping risk — every one of those opcodes has
the identical bug shape (falls through to \`as_number_vm()\`, which reads
a heap-boxed \`VAL_I128\` as \`0.0\`).

A second commit (\`7cab3c60\`) extends the same guard pattern to the
**entire** arithmetic/comparison opcode family, in both of
\`vm_run.c\`'s independent dispatch loops:

- **Threaded/computed-goto** (the loop this platform actually runs):
  \`lbl_SUB\`, \`lbl_MUL\`, \`lbl_DIV\`, \`lbl_MOD\`, \`lbl_NEG\`, \`lbl_ABS\`,
  \`lbl_EQ\`, \`lbl_LT\`, \`lbl_GT\`, \`lbl_LE\`, \`lbl_GE\`
- **Switch-based fallback** (compiled only on non-GCC/non-Clang
  toolchains — dead code here, so verified by temporarily flipping the
  \`#if defined(__GNUC__) || defined(__clang__)\` guard to \`#if 0\` to
  force it to compile, confirming all 12 opcodes raise correctly, then
  reverting): \`OP_SUB\`, \`OP_MUL\`, \`OP_DIV\`, \`OP_MOD\`, \`OP_NEG\`,
  \`OP_ABS\`, \`OP_EQ\`, \`OP_LT\`, \`OP_GT\`, \`OP_LE\`, \`OP_GE\`

Full opcode list guarded (12 total, both loops): \`+ - * / modulo\`
(binary arithmetic), unary \`-\` and \`abs\` (unary arithmetic), \`= < > <=
>=\` (comparison). Native was already loud for the whole family before
this change (LE-03's type-check helper covers both arithmetic and
comparison generically), so no native-side change was needed.

**Test results:**
- 11 new VM standalone self-test entries (\`i128-sub-raises\` through
  \`i128-ge-raises\`) — **73/73 total passing** (up from 62/62).
- Native i128 suite unaffected: **46/46**.
- \`ctest\`: **142/142**.
- VM parity: **144 passed, 0 failed** — no regression.
- Revert-validated at MUL specifically: temporarily disabling just the
  \`lbl_MUL\` guard reproduced the original silent \`"no"\` (wrong-answer,
  uncaught) output; restoring it reproduced \`"caught"\`.

\`docs/KNOWN_ISSUES.md\`'s i128 entry updated from "\`*\` ... not yet
converted" to describing the family-wide guard.
\`.icc/silent-wrong-ledger.yaml\` gets a new \`SW-09b\` entry
(\`LOUD-ERROR\`, \`fixed\`) alongside \`SW-09\`.

PR #424, opened 2026-08-12
fix(parity): fg-marginal native marginal + rationalize VM dispatch (SW-07, SW-08)

tsotchke:
Two silent engine divergences from the v1.3.4 skipped-flaws ledger: **SW-07**
(`fg-marginal`) and **SW-08** (`rationalize`). Both returned a wrong value with
exit 0 and no diagnostic. In both cases each engine was wrong on a *different*
spelling of the same call, so neither could serve as the other's oracle — which
is why nothing in the tree noticed.

## SW-07 — `fg-marginal`

A factor graph with no factor added and no inference run carries the uniform
prior `log(1/dim)` that `eshkol_make_factor_graph()` installs, so the marginal
of a d-state variable is exactly `1/d` in every component. Four independent
defects stood between that arithmetic and the answer:

| where | defect |
|---|---|
| `lib/core/logic_builtins.cpp:52` | The result tensor is allocated with `arena_allocate_tensor_full()`, which sets `num_dimensions` and `total_elements` but leaves `dimensions[]` **unwritten**. Every other caller in the tree fills it in; this one did not. The documented `(fg-marginal (make-factor-graph 2 #(2 2)) 0)` therefore printed `#()` with a correctly computed distribution inside it, shape `(0)`. |
| `lib/core/inference.cpp` `fg_read_numeric()` | Read a tensor or a vector, but not a proper list. `'(2 2)` — the spelling `tests/vm/vm_kb_tensor_test.esk` uses and the VM has always accepted — made `make-factor-graph` answer the tagged NULL, and every later `fg-marginal` answered `()`. |
| `lib/core/inference.cpp` `eshkol_make_factor_graph_tagged()` | No arm for a bare scalar dims argument. A tagged INT64 is not a heap pointer, so the ledger's own reproducer `(make-factor-graph 2 2)` also produced `()`. A scalar now broadcasts: every variable gets that many states. |
| `lib/backend/vm_native.c` native call 520 | Matched `HEAP_TENSOR` only, but a `#(2 2)` literal reaches the VM's mini-compiler as a **VECTOR**. It fell through to the bare-scalar arm, `as_number()` of a heap vector is `0`, and the VM built one zero-state variable — so for the *documented* spelling the VM was the wrong side. The same arm read a scalar as a one-element dims list and then clamped `num_vars` down to that length, truncating `(make-factor-graph 2 2)` to a single variable. |

## SW-08 — `rationalize`

R7RS-small 6.2.6 returns the simplest rational within the tolerance, ordered by
`|p1| <= |p2|` and `q1 <= q2`, with the worked example
`(rationalize (exact .3) 1/10) => 1/3`. The interesting arguments are exact
rationals, and:

- `lib/backend/vm_native.c` native call 349 coerced both arguments with the
  plain `as_number()`, which reads a `VAL_RATIONAL`'s heap index through the
  int64 arm of the union and answers `0.0`. Both arguments became 0, so the VM
  answered the exact integer `0` for every rational input. Native calls 342-347
  already special-case `VAL_RATIONAL`; 349 now uses the heap-aware
  `as_number_vm()`, which also covers bignum and dual arguments. That function's
  own doc comment describes this failure mode exactly.
- `lib/core/rational.cpp` took any exact-integer `x` as its own answer whatever
  the tolerance, and its integer-in-range test tried `floor(lo)+1` before
  `ceil(lo)`. Both made `(rationalize 3 1)` answer `3` where `2` is simpler
  (`|2| <= |3|`, denominators both 1) — and where the VM already answered `2`.
  The exact-integer shortcut is narrowed to a **zero** tolerance, where it is
  what keeps an integer beyond 2^53 exact through the double round-trip, and the
  integer selection now takes `ceil(lo)`, matching `vm_rationalize()`.

This second half is a native-side divergence found while building the parity
corpus, not in the ledger's SW-08 text. It is fixed here because a parity corpus
that routes around a live divergence is exactly the silence the ledger exists to
end. `tests/rational/rationalize_test.esk`'s `(rationalize 3 1)` row is
corrected from 3 to 2 with the spec citation.

## Coverage

- `tests/vm_parity/corpus/57_factor_graph_marginal.esk` and
  `58_rationalize_r7rs.esk` — **both engines**, over every accepted spelling,
  carrying explicit R7RS and uniform-prior oracles (plus an extent oracle for
  the tensor shape), so agreement on a wrong answer still fails.
- `tests/differential/corpus/43_rationalize.esk` and
  `44_factor_graph_marginal.esk` — pin the native jit / jit-nocache / aot-O0 /
  aot-O2 axes to each other.
- `tests/rational/rationalize_test.esk` — gains the spec example and the
  zero-tolerance identity.

**Revert-validate.** With the four source files reverted to `origin/master` and
the tests kept, on a full rebuild: `57` prints `#()`/`()` natively against
`#(0.5 0.5)`/`()` on the VM (parity FAIL) and its `extent-2` oracle FAILs;
`58`'s `simplest-integer` FAILs natively and `spec-example` +
`zero-tolerance-identity` FAIL on the VM. Restoring the fix turns all of them
green.

## Gates (RelWithDebInfo, LLVM 21, arm64-apple-darwin24.1.0)

| gate | result |
|---|---|
| `ctest --test-dir build` | **144/144** passed |
| `scripts/run_vm_parity.sh` | **158 passed, 0 failed** (audit + corpus + oos + fatal; +4 checks from the two new corpus programs) |
| `scripts/run_engine_parity_coverage.py` | **OK** — 102 programs both engines ran clean, 7 divergent (all baselined, **0 new**), 0 regressed, 136/1114 constructs with differential evidence (12.21%, floor 10.95%) |
| `scripts/run_differential.sh tests/differential/corpus/` | 264/270 axis pairs agree across 45 programs; the one divergent program is `37_sort.esk`, pre-existing (LE-01, first-class builtins) and untouched here. Both new programs pass 6/6 axis pairs each. |
| `scripts/run_rational_tests.sh` | 3/3 |
| `scripts/run_logic_tests.sh` | 7/7 |
| `scripts/run_v1_2_edge_cases_tests.sh` | 100/100 |
| `icc chain --preset correctness` (scoped to `lib/core`, `lib/backend`, `tests/vm_parity`, `tests/differential`, `tests/rational`) | both steps MEASURED (417 / 192 scanned), **0 files flagged by more than one detector** |

## Ratchet

`ENGINE_PARITY_BASELINE.json`: `divergent_programs` is **unchanged at 8**.
Neither defect was represented there. #413's merge already took it from 10 to 8
by removing `inference_test.esk` and `continuous_learning.esk`; the remaining
`workspace_test.esk` is a `ws-step!` defect this change does not touch. Both new parity programs are added to
`both_ran_programs`, so a future VM regression that stops running them is a
REGRESSED program rather than a silent reclassification to native-only.

## Also observed, not fixed here

`(vector-ref <marginal-tensor> 0)` answers `0.5` natively and `()` on the VM —
a separate silent divergence in `vector-ref` over a tensor, outside this PR's
scope and not currently ledgered.

---

## Rebased onto master `8ff91b1e` (2026-08-13)

Six squash merges landed under this PR (#413 #418 #419 #420 #423 #428). Rebased, two conflicts:

1. **`lib/backend/vm_native.c` native call 520.** #413 replaced the whole dims-reading block with one unified sequence reader, `vm_fg_seq_doubles()`, that already handles tensor / vector / list / scalar-leaf. **That supersedes this PR's original VM vector arm and is the better structure**, so HEAD's version was taken wholesale. What #413 does *not* do is broadcast: a bare scalar reads as a one-element sequence and the `if (nd < nv) nv = nd` clamp below still cuts `(make-factor-graph 2 2)` to a single variable, where the native half of this PR builds two. Re-resolved as a scalar-**leaf**-only broadcast layered on top of #413's reader — a *short* sequence is an operand mismatch, not a broadcast, and still clamps. The `rationalize` hunk in the same file merged clean.
2. **`tests/vm_parity/ENGINE_PARITY_BASELINE.json`.** Union: kept #413's `56_factor_graph_inference.esk` and this PR's two entries.

No `docs/KNOWN_ISSUES.md` conflict — this PR does not touch it.

**Corpus renumbered** off master's new files (`54_take_drop_srfi1_order`, `54_complex_transcendentals`, `55_inexact_exact_exactness`, `56_factor_graph_inference`, `56_lexical_shadow_nearest_binding`, `42_ad_capture_global_shadow`, `42_diff_quoted_raises`):

| was | now |
|---|---|
| `tests/vm_parity/corpus/56_factor_graph_marginal.esk` | `57_factor_graph_marginal.esk` |
| `tests/vm_parity/corpus/57_rationalize_r7rs.esk` | `58_rationalize_r7rs.esk` |
| `tests/differential/corpus/42_rationalize.esk` | `43_rationalize.esk` |
| `tests/differential/corpus/43_factor_graph_marginal.esk` | `44_factor_graph_marginal.esk` |

Cross-references inside the paired files were updated with them. Full rebuild before re-running anything; all gate numbers in the table above are post-rebase measurements, and both defects re-verified fixed on both engines across all three dims spellings.

## Stale ratchet entry found while re-measuring (not this PR's to retire)

`tests/parser/special_forms_test.esk` is still listed in `divergent_programs` but now **agrees on both engines** at `fc3c1257` — stable over three runs under the gate's own normalisation, and it references neither `rationalize` nor any factor-graph builtin, so it is not this change. One of today's merges fixed it (most likely #428's nearest-enclosing-binding fix) without tightening the ratchet. Left in place here so this PR does not take on another lane's risk or credit; flagged for whoever owns that lane. The ratchet is shrink-only, so the entry is currently silence, not tolerance.

### DIFF
--- a/lib/backend/vm_native.c
+++ b/lib/backend/vm_native.c
@@ -7779,7 +7779,16 @@ static void vm_dispatch_native(VM* vm, int fid) {
         case 348: { Value b_v = vm_pop(vm), a_v = vm_pop(vm);
             vm_push(vm, INT_VAL(vm_rational_gcd((int64_t)as_number(a_v), (int64_t)as_number(b_v)))); break; }
         case 349: { Value tol = vm_pop(vm), x = vm_pop(vm);
-            VmRational* r = vm_rationalize(rat_arena, as_number(x), as_number(tol));
+            /* as_number_vm, NOT as_number: `rationalize` is the one operation
+             * whose arguments are USUALLY exact rationals — the R7RS-small
+             * example is `(rationalize (exact .3) 1/10) => 1/3` — and the
+             * plain as_number() reads a VAL_RATIONAL's heap index through the
+             * int64 arm of the union and answers 0.0.  Both arguments then
+             * became 0, and `(rationalize 1/3 1/10000)` returned the exact
+             * integer 0 with exit 0 where the native engine returns 1/3
+             * (SW-08).  Rationals, bignums and duals all need the heap-aware
+             * coercion, exactly as native calls 342-347 above already do. */
+            VmRational* r = vm_rationalize(rat_arena, as_number_vm(vm, x), as_number_vm(vm, tol));
             if (!r) { vm_push(vm, NIL_VAL); break; }
             if (r->denom == 1) { vm_push(vm, INT_VAL(r->num)); break; }
             int32_t ptr = heap_alloc(&vm->heap); if (ptr < 0) { vm->error = 1; break; }
@@ -9064,6 +9073,17 @@ static void vm_dispatch_native(VM* vm, int fid) {
             var_dims[nd++] = (d > 0) ? d : 2;   /* a non-positive dim is not a
                                                  * variable domain; binary is
                                                  * the documented default */
+        }
+        /* A BARE SCALAR dims argument means "every variable has this many
+         * states", so `(make-factor-graph 2 2)` is a two-variable graph and
+         * `(fg-marginal fg 1)` is a marginal rather than `()`.  Read as a
+         * one-element sequence it hits the clamp below and cuts num_vars to 1,
+         * where the native engine broadcasts — SW-07's own reproducer.  Only a
+         * genuine scalar LEAF broadcasts: a short sequence is an operand
+         * mismatch, not a broadcast, and still clamps. */
+        if (nd == 1 && nv > 1 && vm_tensor_collection_len(vm, dims_val) < 0) {
+            for (int i = 1; i < nv && i < 64; i++) var_dims[i] = var_dims[0];
+            nd = (nv < 64) ? nv : 64;
         }
         if (nd == 0) {                          /* dims omitted/unreadable */
             int want = (nv > 0 && nv < 64) ? nv : 0;
--- a/lib/core/inference.cpp
+++ b/lib/core/inference.cpp
@@ -88,10 +88,16 @@ static bool extract_doubles_from_vector(const eshkol_tagged_value_t* tv,
     return true;
 }
 
-/* Read numeric elements from a tensor (#(…)) OR a Scheme vector ((vector …)) as
- * doubles into out[0..max-1]; returns the count read (0 if neither / empty).
+/* Read numeric elements from a tensor (#(…)), a Scheme vector ((vector …)) OR a
+ * proper list ('(…)) as doubles into out[0..max-1]; returns the count read
+ * (0 if none of those / empty).
  * Lets the FG builtins accept runtime-constructed vectors, not only tensor
- * literals (bug-QQ-2). INT64 elements promote to double. */
+ * literals (bug-QQ-2). INT64 elements promote to double.
+ *
+ * The list spelling is not cosmetic: the VM's native dispatch has always read
+ * `(2 2)` for both the factor-graph dims and the fg-add-factor! variable
+ * indices, and `tests/vm/vm_kb_tensor_test.esk` uses it, so a native path that
+ * rejected it answered `()` where the VM answered the marginal (SW-07). */
 static uint32_t fg_read_numeric(const eshkol_tagged_value_t* tv, double* out, uint32_t max) {
     if (!tv || tv->type != ESHKOL_VALUE_HEAP_PTR || !tv->data.ptr_val) return 0;
     void* ptr = (void*)(uintptr_t)tv->data.ptr_val;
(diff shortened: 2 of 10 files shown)


tsotchke:
## ICC verdicts at `5b848fa4` (repo `eshkol-quickfix1`)

Following the gate recipe in `.swarm/runtime/CODEX_HANDOFF.md` — fresh `build` (RelWithDebInfo, Ninja, LLVM 21) plus `build-quantum` (`-DESHKOL_QUANTUM_ENABLED=ON -DESHKOL_BUILD_TESTS=ON`), which the language-coverage floor probe requires or it false-reds with exit 2.

| step | verdict |
|---|---|
| `scripts/run_icc_smoke.sh` | **58/58 passed — ICC smoke gate PASS** |
| `scripts/run_ctest_gate.sh` | **PASS (142/142)**, trace emitted |
| `scripts/run_vm_parity.sh` | **148 passed, 0 failed** |
| `icc architecture-verify --emit-trace` | **OK — 9 PASS / 0 FAIL / 0 UNCHECKABLE** across 9 invariants, 0 drift refs. `INV-engine-semantic-parity` (severity critical, incident `vm-parity-silent-wrong-answer`) PASS. |
| `icc chain --preset correctness` (path-scoped to `lib/core`, `lib/backend`, `tests/vm_parity`, `tests/differential`, `tests/rational`) | both steps MEASURED (417 / 192 scanned), **0 files flagged by more than one detector** |
| `icc readiness --target eshkol-compiler-readiness` (both trace routes) | **score 77**, 38 criteria PASS |

Readiness 77 is the pre-existing figure, not a regression: the two blockers are

1. `failure_free` — 5 events, **all of them `tests/differential/corpus/37_sort.esk`** (`exit codes differ: jit rc=1 vs jit-nocache rc=139`, aot rc=125). That is LE-01, first-class builtins SIGSEGV, which the v1.3.4 skipped-flaws ledger already records as "the `failure_free` blocker holding readiness at 77 on #420's branch" — same number, same single program, untouched here.
2. `no_flag_without_contract_…` contract-graph gap = 1, also pre-existing.

Both new differential-corpus programs pass all 10 native axis pairs; the differential suite is 252/258 with `37_sort.esk` as the only divergent program.

PR #425, opened 2026-08-12
chore(oracle): close SW-07 and SW-08 by measurement (fixed by #424)

tsotchke:
Stacked on #421, which owns `.icc/silent-wrong-ledger.yaml`. Marks SW-07 (`fg-marginal`) and SW-08 (`rationalize`) **closed**, `closed_at: 5b848fa4`, `fixed_by: PR #424` — with the measured evidence rather than an assertion, per the ledger's invariant 3.

`scripts/gate_no_silent_wrong.py` goes from **26 blocking entries to 24** on this ledger.

Both entries turned out to be more than the ledger recorded, and in both cases each engine was wrong on a different spelling, which is why nothing in the tree noticed:

- **SW-07** is four defects. The native result tensor's `dimensions[]` array was never written, so the *documented* `#(2 2)` form printed `#()` around a correct distribution (shape `(0)`). Native dims parsing accepted neither a list nor a bare scalar, so `'(2 2)` and the ledger's own `(make-factor-graph 2 2)` reproducer answered `()`. And the VM matched `HEAP_TENSOR` only, so a `#(2 2)` VECTOR literal fell through to its bare-scalar arm and built one zero-state variable, while a real scalar was read as a one-element dims list that clamped `num_vars` to 1.
- **SW-08** is two defects, one per engine. The VM read both arguments through the non-heap-aware `as_number()`, so every exact-rational input answered the exact integer `0`. Natively an exact-integer `x` was taken as its own answer whatever the tolerance, so `(rationalize 3 1)` answered `3` where R7RS-small 6.2.6 makes `2` the simplest rational in `[2,4]` — and where the VM already answered `2`.

The SW-07 notes record what #424 does **not** close: the three baselined logic-suite divergences (`inference_test.esk`, `continuous_learning.esk`, `workspace_test.esk`) are `fg-infer!`, `free-energy` and `ws-step!` defects, still divergent at 5b848fa4, so `ENGINE_PARITY_BASELINE.json`'s `divergent_programs` stays at 10.

Split out rather than committed onto `chore/no-silent-wrong-oracle` directly so #421's lane stays under its author's control.

### DIFF
--- a/.icc/completion-oracles.yaml
+++ b/.icc/completion-oracles.yaml
@@ -56,6 +56,30 @@ oracles:
         label: Codegen / runtime / type-checker production paths are real
         action: wire_real_path_or_classify_stub
 
+      # ── Silent-wrong flaw gate (2026-08-12) ──────────────────────────
+      # Every other criterion in this oracle certifies that something
+      # WORKS. None of them can certify that nothing SILENTLY LIES: a
+      # defect that returns a wrong value with no diagnostic and exit 0
+      # is invisible to a probe that only asserts its own happy path.
+      # That is exactly how a release cut reached readiness 100 while
+      # carrying an AD operator that differentiates the wrong function,
+      # a `map` that ignores a shadowing binding, and seven documented
+      # resource limits that are never enforced.
+      #
+      # `.icc/silent-wrong-ledger.yaml` is the enumeration of that class;
+      # `scripts/gate_no_silent_wrong.py` grades it and emits this event.
+      # The gate fails closed — a missing or unparseable ledger is FAIL,
+      # because absence of the ledger is not evidence of absence of the
+      # defects. Entries close only by measurement at a named SHA, and a
+      # waiver needs an owner, a reason and an unexpired date.
+      - runtime_event:
+          event_kinds: [eshkol_smoke]
+          event_names: ["no_open_silent_wrong"]
+          event_values: ["PASS"]
+        severity: high
+        label: 'No open, unwaived SILENT-WRONG flaw ships: every entry in .icc/silent-wrong-ledger.yaml whose bucket is SILENT-WRONG (or which is IN-FLIGHT but silent_wrong_class) is closed at a named SHA or carries an owner/reason/unexpired waiver — wrong values, wrong derivatives and wrong memory outcomes with no diagnostic and exit 0 are tag-blocking'
+        action: python3 scripts/gate_no_silent_wrong.py
+
       - runtime_event:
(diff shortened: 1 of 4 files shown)


tsotchke:
## Heads-up: this PR's base moved under it, and its diff is no longer what it looks like

I force-pushed `chore/no-silent-wrong-oracle` (the base of this PR) after rebasing it
onto `origin/master` at `019ab413`. Master had absorbed the ledger via #423/#424/#427, so
my branch's seven ledger-only commits became empty and were dropped; the branch now
carries **only** the enforcement machinery (gate script, oracle criterion, smoke probe).

That broke this PR's diff. It now reports **4 files / +1127** instead of one file and two
entries, because its head still carries my eight pre-rebase commits that the new base no
longer has. Merging it as-is would revert master's `SW-09b`, `SW-27`, and the
`fixed`/`guarded` closures.

## This PR is NOT superseded — its content is still needed

I checked: master still has `SW-07` and `SW-08` at `status: open`. The closure evidence
in this PR exists nowhere else. Please keep it.

## The fix is small, and I verified it applies cleanly

Cherry-picking this branch's unique commit `5610a143` directly onto `origin/master`
succeeds with no conflict and produces exactly the intended change:

```
.icc/silent-wrong-ledger.yaml | 63 ++++++++++++++++++++++++++-----
1 file changed, 58 insertions(+), 5 deletions(-)
  SW-07: status open -> closed, closed_at "fc3c1257"
  SW-08: status open -> closed, closed_at "fc3c1257"
```

So the remedy is: **rebase this head onto `master` (dropping my eight old commits) and
retarget the base from `chore/no-silent-wrong-oracle` to `master`.** It no longer needs
to be stacked on my PR at all — the ledger it edits lives on master now.

I did not do this myself because it means force-pushing another lane's branch. Say the
word and I will.

## Measured effect once it lands

With this commit applied on top of master, the silent-wrong gate goes from **23 blocking
to 21**. The `SW-07`/`SW-08` closures are both properly SHA-backed at `fc3c1257`, which
is the discipline the gate wants and which four other terminal entries currently lack.

Also noting your separate finding for whoever owns it: `tests/parser/special_forms_test.esk`
is still in `divergent_programs` but now AGREES on both engines — a stale ratchet entry,
most likely left by #428.

tsotchke:
Absorbed into #421 at cebb4051 — the SW-07/SW-08 closure commit (5610a143) was cherry-picked onto the rebased chore/no-silent-wrong-oracle branch after this PR's base moved under it. No content lost; closing rather than merging to avoid reverting master's newer ledger entries.

PR #427, opened 2026-08-12
fix(codegen): builtins as first-class values — close the LE-01 SIGSEGV family

tsotchke:
## What this fixes

**LE-01** — builtins used as **first-class values** (passed as arguments, stored,
returned) crashed or failed to compile, while the same builtins in call position
worked. Driving reproducer: `tests/differential/corpus/37_sort.esk`, i.e.
`(sort lst string<?)`, the program holding v1.3.4 readiness at 77 through the
`failure_free` criterion.

## Root cause — two defects, which is why every axis failed differently

**1. No first-class representation for inline-lowered builtins.**
`codegenCall` (`lib/backend/llvm_codegen.cpp`) string-matches the head NAME and
lowers the operation inline — `string<?` to `StringIOCodegen::stringCompare`,
`vector-ref` to the vector path, `expt`/`min`/`max` to `ArithmeticCodegen`. That
is the right lowering for `(string<? a b)`, but it leaves the NAME denoting
nothing. `codegenVariable` carried a hand-maintained cascade of per-builtin
wrapper factories (`createBuiltinComparisonFunction`,
`createBuiltinPredicateFunction`, `createBuiltinCharFunction`, …), each
re-implementing the operation a SECOND time in IR. Anything not on that list fell
through to:

- `codegen_error_at(ast, "Undefined variable: %s", …)` at
  `lib/backend/llvm_codegen.cpp:11697` — for a program containing no undefined
  variable; or
- the raw-`Function*` fallback at `lib/backend/llvm_codegen.cpp:11235`, which
  handed the closure dispatcher a pointer to a symbol with a **foreign ABI**.
  Calling that is undefined behaviour: `(h remainder 7 3)` SIGSEGV'd,
  `(h append '(1) '(2))` silently answered `1`.

**2. The JIT-execution gate crashed while refusing.**
`ReplJITContext::addModule` (`lib/repl/repl_jit.cpp:1802`) throws to refuse a
module whose compilation reported an error. `module` and `module_context` are
both parameters of that frame, and the order in which an exception unwind
destroys function parameters is **unspecified**. `llvm::Module` holds a bare
reference to its `LLVMContext` and calls `Context.removeModule(this)` from its
destructor, so when the context went first the refusal path dereferenced freed
memory: SIGSEGV at address `0x0` in `LLVMContext::removeModule` — the compiler
crashing while **correctly rejecting** a program it had already diagnosed.

That second defect is why a plain "Undefined variable" presented as exit 139 on
exactly the axes that reach the in-process JIT. **It is not the #420 object-cache
signature the ledger speculated it was**; #420's defect is unrelated.

## The mechanism chosen

Generalise the wrapper-closure **idiom** without the duplicated bodies.
`createInlineBuiltinWrapper` synthesises `builtin_fc_<name>` with the closure ABI
`(tagged_value…) -> tagged_value` and generates its **body by re-entering
`codegenCall`** on a synthetic `(name p0 … pN-1)` AST whose arguments are the
wrapper's own parameters. The authoritative call-site lowering *is* the wrapper
body, so a first-class reference cannot disagree with a direct call, and
extending the first-class surface is one row of data rather than a new IR
implementation that can drift.

`addModule` now destroys `module` then `module_context` explicitly on the refusal
path.

### Scope discipline

The new path is consulted **only** where the old code would have produced a
garbage callable or an "Undefined variable" error. It never changes which binding
a name resolves to: locals, captures, user definitions and the REPL namespaces
are all decided strictly before it. `(define (min a b) (+ a b))` still shadows
the builtin — pinned by a check in the test. **This does not touch the
name-resolution ordering** worked on by the map-shadowing lane
(SW-01/SW-02/SW-24); the two changes are in different sites.

## Measurement — `tests/differential/corpus/37_sort.esk`

| axis | before | after |
|---|---|---|
| `jit` | exit 1 | exit 0, correct output |
| `jit-nocache` | exit 139 (SIGSEGV) | exit 0, correct output |
| `aot-O0` | compile fail (rc 125 in the runner) | exit 0, correct output |
| `aot-O2` | compile fail (rc 125 in the runner) | exit 0, correct output |

All four now print byte-identical output. **37_sort.esk was not baselined,
excluded, or weakened** — it is green because the compiler is fixed.

## Tests

Every name in the LE-01 family plus a sweep of ~60 more is passed **through** a
higher-order function — the only way to force the value representation — and also
stored in a variable, returned from a procedure, and read out of a list.

- `first_class_builtins_runtime_smoke` (`-r`, run cache on)
- `first_class_builtins_nocache_smoke` (`-r`, `ESHKOL_JIT_CACHE=0`)
- `first_class_builtins_aot_o0_smoke`
- `first_class_builtins_aot_o2_smoke`
- `first_class_builtins_vm_smoke`
- `undefined_variable_refusal_is_clean_nocache`
- `undefined_variable_refusal_is_clean_cached`

**Revert-validated**: with the fix disabled, 5 of the 7 go red and 37_sort
returns to 1 / 139 / compile-fail / compile-fail. The two that stay green are the
VM test and the cached-refusal test, whose axes were never broken.

## The VM engine

**Not affected — verified, not assumed.** The VM reaches builtins through a
native dispatch table, so a builtin is a value there by construction. Pinned by
`tests/integration/first_class_builtins_vm_test.esk` so a future
call-position-only lowering in the VM is caught the same day.

## Two adjacent defects found and deliberately NOT absorbed here

Both live in a different lowering path, and folding them in would have made this
change unreviewable:

1. **A rest-arg stdlib procedure referenced as a value loses its variadic flag.**
   `(h append '(1) '(2))` answers `1`; `(h string-copy "abc")` SIGSEGVs under
   AOT. `append` and `string-copy` are Scheme procedures in
   `lib/core/list/transform.esk` and `lib/core/strings.esk`, not inline builtins,
   so neither reaches the new path. Recorded in the code comments and in the test
   file rather than papered over with a table row that would never be reached.
2. **The VM's builtin surface is narrower than native.** `string>?`,
   `string<=?`, `string>=?`, `string-length`, `string-ref`, `symbol=?` and
   `square` are undefined variables on the VM **in call position too**, and
   `boolean=?` reaches an unimplemented native id. Standing parity-ratchet gaps
   (PR-02), not this defect class.

## Merge-order note

`.icc/silent-wrong-ledger.yaml` lives on the unmerged branch
`chore/no-silent-wrong-oracle`; it is carried here so the LE-01 closure is
visible to `scripts/gate_no_silent_wrong.py`. If that branch lands first this is
a trivial one-entry conflict; if this lands first, that branch rebases.


tsotchke:
## Gate results — measured at `168bd8a4` (branch head, rebased onto `9b3fd14d`)

Fresh `RelWithDebInfo` LLVM-21 build plus a `-DESHKOL_QUANTUM_ENABLED=ON` build
(the language-coverage floor probe needs the latter or it exits 2 as a false red).

| Gate | Result |
|---|---|
| `ctest --test-dir build` | **149/149 PASS** (142 before this branch + the 7 added here) |
| `scripts/run_ctest_gate.sh` | **PASS**, 149/149, all gated groups green |
| `scripts/run_icc_smoke.sh` | **58/58 PASS** |
| `scripts/run_vm_parity.sh` | **146 passed / 0 failed** |
| `scripts/run_differential.sh tests/differential/corpus/` | **246/246 axis pairs agree across 41 programs** |
| `scripts/run_engine_parity_coverage.py` | **OK** — 10 divergent (all ratcheted), 0 new, 0 regressed, construct floor held |
| `scripts/run_language_coverage.sh` | **PASS** — 1091 constructs, 100.00% execution-backed, deficit 0, high-risk uncovered 0 |
| `icc architecture-verify` | **OK — 9 PASS / 0 FAIL / 0 UNCHECKABLE** across 9 invariants |
| `icc readiness --target eshkol-compiler-readiness` | **score 89**, 37 PASS / 1 FAIL (was **77** at `9f2da2ab`) |

### `37_sort.esk`, the driving reproducer, per axis

| pair | result |
|---|---|
| `jit-vs-jit-nocache` | PASS |
| `jit-vs-aot-o0` | PASS |
| `jit-vs-aot-o2` | PASS |
| `jit-nocache-vs-aot-o0` | PASS |
| `jit-nocache-vs-aot-o2` | PASS |
| `aot-o0-vs-aot-o2` | PASS |

All four axes exit 0 with byte-identical output. Not baselined, not excluded.

### `failure_free`

**Zero FAIL events in the whole regenerated trace corpus** — 1111 events across 17
sources (`differential_smoke` 247, `vm_parity` 148, `ctest` 155, `eshkol_smoke` 58,
`escape_matrix` 28, `wasm_parity` 63, `edge_coverage` 172, …). That criterion was
consuming the six 37_sort divergence events and the corpus roll-up.

### The one remaining readiness FAIL is not this defect

`artifact_without_test_or_trace` on `Artifact:results-%s.json`, defined at
`scripts/doc_audit/run_examples.py:107` — severity **low**, a contract-graph
completeness gap, not a runtime failure. That file is untouched by this branch
(last changed by #411) and the gap is pre-existing.

### One measurement correction worth recording

The first smoke run on this branch reported `engine_semantic_parity` FAIL on
`tests/vm_parity/corpus/54_take_drop_srfi1_order.esk`. That was a **stale-binary
artifact, not a defect**: the branch had just been rebased onto #418, which adds
that corpus program and fixes the VM prelude's `take`/`drop` order, but the build
tree still held the pre-rebase binaries — so a new test ran against an old VM.
Rebuilt and re-measured: OK, divergent count 11 → 10, floor held. Recorded because
"rebase, then run the harness before rebuilding" is an easy way to manufacture a
phantom divergence.

tsotchke:
## Second defect closed on this branch — SW-27, rest-arg procedures as first-class values

Follow-up on the residue the LE-01 fix deliberately did not absorb, ruled in scope as the
same mechanism family. Measured at `2dee5d39`.

### What was wrong

A rest-arg procedure is called through a **different ABI** from a fixed-arity one: the
closure dispatcher conses the caller's surplus arguments into a list and passes
`fixed_params` values plus that one list. It selects that path from the closure metadata —
bit 63 of `packed_info` (which the allocator turns into `CLOSURE_FLAG_VARIADIC`) and, for
the capture-less case where `env` is null, the `input_arity` byte, which it reads back as
`fixed_params`.

Every function-as-value site packed a plain arity and **no variadic bit**:
`emitFunctionAsCallableValue`, the `function_table` branch of `codegenVariable`, the REPL
hot-reload branch, and the R7RS §5.3.1 redefinition store. A rest-arg procedure referenced
as a value was therefore dispatched as if fixed-arity, and its rest parameter received a
bare argument where a list belonged:

| reference | before | correct (= call position) |
|---|---|---|
| `(h append '(1) '(2))` | `1` — the first argument, **exit 0, silent** | `(1 2)` |
| `(h iota 3 10)` | `(0 1 2)` — rest argument ignored, **silent** | `(10 11 12)` |
| `(h make-list 3 7)` | raises `cdr: argument is not a pair` | `(7 7 7)` |
| `(h partial + 10)` | raises `car: argument is not a pair` | a closure |
| `(h string-copy "abc")` | **SIGSEGV under AOT** | `abc` |

### NOT a rest-arg ABI change

The dispatcher already implemented the rest-arg calling convention and already read these
fields — the value side simply never set them. The reference paths now resolve the shape
from the same two registries the **call** path consults (`variadic_function_info`,
`g_repl_variadic_functions`) through one shared `lookupVariadicProcedure` helper, so value
and call cannot disagree about whether a procedure takes a rest argument.

One asymmetry the fix has to respect: for a fixed-arity closure the arity field is the
parameter count, but for a variadic one the dispatcher reads it as the **fixed** count. A
1-fixed rest-arg procedure stores `1`, not `2`.

### The oracle is differential by construction

Each check compares the value obtained **through** a higher-order function against the
identical call **in call position, in the same program**. Nothing hard-codes what `iota` or
`json-get` ought to return, so the test cannot pass by agreeing with a wrong expectation.
Coverage: stdlib rest-arg procedures at 0/1/2 fixed parameters (`append`, `make-list`,
`iota`, `string-copy`, `partial`, `json-get`), user rest-arg procedures at each fixed count
with empty and non-empty rest, the rest parameter's list-ness and length, the builtin
variadics (`list`, `+`, `string-append`) as non-regression, and stored / returned /
read-from-a-list references.

Gated by `first_class_variadics_{runtime,nocache,aot_o0,aot_o2,vm}_smoke`.
**Revert-validated**: with the fix disabled all four native gates go red; the VM gate stays
green, which is correct — the VM's closures carry their rest-arg shape natively and never
had this defect (verified, not assumed, and now pinned).

### How this escaped every gate we own

Worth recording, because the answer is not "add a corpus program". Every gate was green
while `(h append '(1) '(2))` returned `1`: the differential corpus compares **axes against
each other** and all four axes were wrong *identically*; vm-parity excuses native-only
programs; and no test passed a rest-arg procedure as a value at all. A defect that is wrong
the same way everywhere is invisible to differential testing by construction. The escape
fix is a **value-position axis** on the higher-order sweep.

## Gates re-run at `2dee5d39`

| Gate | Result |
|---|---|
| `ctest` | **154/154 PASS** (142 before this branch, +12) |
| `scripts/run_ctest_gate.sh` | **PASS**, 154/154 |
| `scripts/run_icc_smoke.sh` | **58/58 PASS** |
| `scripts/run_vm_parity.sh` | **146 passed / 0 failed** |
| `scripts/run_differential.sh` | **246/246 axis pairs across 41 programs** |
| `scripts/run_engine_parity_coverage.py` | **OK** — 10 ratcheted, 0 new, 0 regressed |
| `icc architecture-verify` | **OK — 9 PASS / 0 FAIL / 0 UNCHECKABLE** |
| `icc readiness` | **89**, 37 PASS / 1 FAIL (unchanged from the LE-01 measurement; was **77** at `9f2da2ab`) |
| FAIL events in the trace corpus | **0** across 1122 events |

The single remaining readiness FAIL is still `artifact_without_test_or_trace` on
`Artifact:results-%s.json` (`scripts/doc_audit/run_examples.py:107`) — low severity,
pre-existing (#411), untouched by this branch.

### Adjacent, found by the same sweep, left open

`(zset-union …)` in `core.dbsp` SIGSEGVs **in call position too**, so it is a different
defect and this oracle can say nothing about it. Recorded in the ledger rather than
absorbed.

tsotchke:
## Rebased onto `8ff91b1e` — conflicts resolved, all gates re-measured

Master moved under this PR: six squash merges landed (#413, #418, #419, #420, #423, #428).
Rebased and re-verified at `2f85890e`.

### Conflicts hit: exactly one

| File | Class | Resolution |
|---|---|---|
| `.icc/silent-wrong-ledger.yaml` | **add/add** | **UNION.** Master's copy (which #423 landed, carrying its SW-06 / SW-09 / SW-09b / SW-22 closures and the new SW-09b entry) is the base; only the LE-01 entry was replaced with this branch's closed version. Entry set is byte-identical to master's — 63 entries, SW-09b present — plus LE-01 closed and SW-27 added. YAML re-validated. |

The other two predicted conflict classes **did not occur**: `llvm_codegen.cpp` applied
cleanly over #423's `codegenDiff` guard and #420's capture fixes, and `docs/KNOWN_ISSUES.md`
is untouched by this branch so it never conflicted. The remaining four commits rebased
without conflict.

One extra commit was needed: the rebase rewrote this branch's SHAs, so the `closed_at`
values LE-01 and SW-27 cited no longer resolved. The ledger's own invariant 3 is that an
entry closes by measurement at a NAMED SHA, and a name that resolves to nothing is not a
name — re-pointed at `9387bbdf` (LE-01) and `256c31a8` (SW-27).

### #420 landing strengthens the retraction rather than weakening it

This branch's earlier commit retracted the ledger's guess that LE-01's cached-vs-nocache
axis split was "the same object-cache signature as IF-01, so the two are probably one root
cause". #420 is now **in master**, and the LE-01 defect was still fully present on this
rebased tree until commit `9387bbdf`. The claim is no longer an argument; it is a
measurement.

### Gates re-measured after a **full rebuild** at `2f85890e`

Rebuilt `build` and `build-quantum` from scratch before running anything — a harness run
against a pre-rebase binary is how this branch previously manufactured a phantom
`engine_semantic_parity` divergence.

| Gate | Before rebase | After rebase |
|---|---|---|
| `ctest` | 154/154 | **156/156 PASS** |
| `scripts/run_ctest_gate.sh` | PASS 154/154 | **PASS 156/156** |
| `scripts/run_icc_smoke.sh` | 58/58 | **59/59 PASS** |
| `scripts/run_vm_parity.sh` | 146 / 0 | **154 passed / 0 failed** |
| `scripts/run_differential.sh` | 246/246 across 41 | **258/258 axis pairs across 43 programs** |
| `scripts/run_engine_parity_coverage.py` | OK, 10 divergent | **OK, 7 divergent**, 0 new, 0 regressed, floor 12.03% |
| `icc architecture-verify` | 9 PASS / 0 FAIL | **9 PASS / 0 FAIL / 0 UNCHECKABLE** |
| `icc readiness` | 89 (37 PASS / 1 FAIL) | **89 (38 PASS / 1 FAIL)** |
| FAIL events in trace corpus | 0 / 1122 | **0 / 1152** |

The counts rose because master's merges added corpus programs and gates; the divergent
count fell from 10 to 7 because #413/#418/#428 closed three of them. `37_sort.esk` remains
**6 PASSED / 0 FAILED** across all axis pairs.

The single remaining readiness FAIL is unchanged and unrelated:
`artifact_without_test_or_trace` on `Artifact:results-%s.json`
(`scripts/doc_audit/run_examples.py:107`), low severity, pre-existing (#411).

PR #435, opened 2026-08-13
fix(vm): captured-parameter set! writeback (SW-25); VM heap-growth guard (SW-14 NEEDS-DECISION)

tsotchke:
Two VM-engine defects from `.icc/silent-wrong-ledger.yaml`. **One is fixed at the
root; the other carries the measurement that changes what it is, ships the guard
that closes its *silent* half, and is now dispositioned by maintainer ruling
(2026-08-13) as the v1.3.5 flagship item.**

## SW-25 — FIXED

```scheme
(define (f n) ((lambda () (set! n (+ n 1)))) n)
(display (f 5))          ; VM 5, native 6 — silently, exit 0, no diagnostic
```

### Root cause

`compile_form_let()` (`lib/backend/vm_compiler.c`) has always run
`needs_boxing()` (`lib/backend/vm_parser.c:941`) over its bindings, so a
`let`-bound variable that is both `set!`-mutated **and** captured by a nested
lambda lives in a shared 1-element vector and every closure reads and writes the
same cell.

**Parameters never got that treatment.** `compile_form_define()` (the
fixed-parameter loop), `compile_form_lambda()` and `compile_form_lambda_2()`
called `add_local()` for each parameter and went straight to the body, so
`OP_CLOSURE` copied the parameter's **value** into the closure's upvalue array
and the inner `set!` mutated a private copy that was discarded on return.

The open-slot relay (native calls 151/252) is not an alternative: it aliases an
absolute VM stack slot and is deliberately restricted to the top-level chunk,
because a function's frame is destroyed on return and the alias would outlive
it.

### Fix

`vm_box_mutable_captured_params()` heap-boxes every parameter `needs_boxing()`
reports as mutated-and-captured, emitting `GET_LOCAL` / `VEC_CREATE 1` /
`SET_LOCAL` at function entry — **after** `OP_PACK_REST`, which reads the raw
argument window — and marking the `Local` `boxed`, so every later read and write
compiles through the existing boxed paths and a nested closure captures the
**box**.

The same gap existed in three more binding forms, all closed with the same
mechanism:

| form | what was wrong |
|---|---|
| `let*` | doc comment literally said *"No mutable-capture boxing (unlike plain `let`)"*; scan now widened to the later initializers plus the body, because a `let*` binding is in scope for both |
| `letrec` | box created with the placeholder, before any initializer runs, so the box pointer is already final when an initializer's closure captures it; initializers store through `VEC_SET` |
| `letrec*` | same |

No new opcode, no ABI change, `eskb` round-trips unchanged.

### Class-kill

`tests/vm_parity/corpus/62_mutable_capture_writeback.esk` — byte-identical VM vs
native on every line: depths 1-3, stored **and** immediately-invoked closures,
two closures sharing one capture, mutation from inside a loop, a closure that
outlives the frame that bound the parameter, `lambda` as well as `define`
parameters, two instantiations *not* sharing a box, `let*`/`letrec`/`letrec*`,
plus controls (unmutated capture, mutated-but-uncaptured parameter, and the
SW-24/#428 shadowing control).
`tests/differential/corpus/50_mutable_capture_writeback.esk` covers the native
axes.

**Revert-validated**: restoring `lib/backend/vm_compiler.c` alone reproduces
VM 5.

### Two adjacent defects found while class-killing, deliberately NOT baked into a gate

Both are **native-side** and out of this entry's scope; recorded in the ledger so
the next pass starts from measurement rather than rediscovery:

1. a mutated+captured **rest** parameter does not write back on the native axes —
   `(define (r . xs) ((lambda () (set! xs (cons 99 xs)))) (car xs))` is `99` on
   the VM after this fix and `1` natively;
2. `(do ((i 0 (+ i 1)) (acc 0)) ((= i 3) acc) ((lambda () (set! acc (+ acc i)))))`
   is killed by SIGKILL (exit 137) natively.

A third, pre-existing and **loud** (so not silent-wrong): an internal
`(define (bump) (set! n ...))` capturing an enclosing parameter is a fatal VM
runtime error on unmodified master.

### The measurement changes what the entry is

The same 20000 × 200-cons fixture, `/usr/bin/time -l` peak RSS at `2ae3787f`:

| | peak RSS |
|---|---|
| **with** the `with-region` wrapper | 1,503,297,536 bytes |
| **without** it | 1,504,067,584 bytes |

The wrapped run is fractionally *smaller* — the difference is allocator noise.
`with-region` does not reclaim less than native, it reclaims **nothing**, and so
does everything else: **the bytecode VM has no heap reclamation of any kind.**
`heap_alloc()` bump-allocates from an arena that is never released and hands out
a monotone object index that is never recycled; `vm_region_pop()` exists
(`lib/backend/vm_arena.h`) but no bytecode calls it, because
`compile_form_with_region()` lowers the form to `begin`.

### Why it is not fixed here

A correct fix is a **VM-heap escape evacuator** — a campaign the size of the
native ESH-0214c/ESH-0214d work, not a patch. It needs, at minimum:

1. **root enumeration** over the operand stack plus ~15 `Value`-carrying side
   tables in `struct VM` (`outputs`, `handler_stack.promise_mark`,
   `current_exception`, `wind_stack`, `parameter_bindings`, `promise_eval_head`,
   `lru_caches`, `event_emitters`, `channels`, `timers`, `exit_handlers`, …) — a
   missed root frees live memory;
2. **per-subtype interior traversal** for all 27 `HeapType`s and their arena
   payloads (`VmVector.items`, `VmString.data`, `VmBignum.limbs`,
   `VmSubstitution.terms`, `VmFact.args`/`datum_ptr`, `VmFactorGraph`'s five
   pointer arrays, `VmContinuation`'s six saved arrays, `VmParameter.save_stack`,
   …) — exactly the coverage problem ESH-0214d had to close natively, which
   still carries a documented watchlist there;
3. a **mutation write barrier** at every channel that can store a `Value` into a
   longer-lived container (~130 candidate sites across `vm_native.c`/`vm_run.c`),
   or a full mark from roots in its place;
4. **index-space recycling or compaction**, because the object-table index is a
   monotone counter — freeing only the arena still leaks 8 bytes per object and
   still terminates at `ESHKOL_VM_HEAP_MAX_SIZE`.

**A partial implementation is strictly worse than the defect.** Natively a missed
escape dangles into freed arena memory; on the VM it aliases a *live* object
through a recycled index and yields a silently **wrong value**. Landing that
under a SILENT-WRONG closure would invert the ledger's own contract.

### What does land: the silent half

Neither guard changes any answer.

- The first region form executed on the VM prints a **one-time note** naming the
  boundary (`vm_region_reclaim_notice()`), from both
  `compile_form_with_region()` and native 2210 `region-open`.
  `ESHKOL_VM_REGION_QUIET=1` silences it.
- The VM arena is sampled every 4096 allocations; crossing
  `ESHKOL_VM_HEAP_BUDGET_MB` (default 1024) prints a diagnostic naming the size,
  the budget and the cause (`heap_check_budget()`).
  `ESHKOL_VM_HEAP_BUDGET_FATAL=1` makes it exit nonzero so a lane can gate on it;
  `ESHKOL_VM_HEAP_BUDGET_MB=0` disables it.

The env knobs are read through host hooks in the **hosted** `vm_io.c`, because
the VM **core** component is freestanding by contract —
`tests/toolchain/vm_source_boundary_test.cpp` fails the build if `vm_core.c` so
much as names `getenv(`.

Gated by `tests/memory/vm_region_growth_watchdog_test.sh` (8 checks) and the
icc-smoke probe `vm_region_growth_watchdog`, so the guard cannot be removed
silently. Documented in `docs/KNOWN_ISSUES.md`,
`docs/reference/runtime/memory-model.md` and
`docs/reference/runtime/environment-variables.md`.

### Maintainer ruling 2026-08-13 — applied

**The loud guard ships for v1.3.4.** The full VM evacuator is the **v1.3.5
flagship item**: an OALR Stage-1 port of the native design — Cheney-style
copying evacuation with a forwarding map and a mutation write barrier — with all
27 heap subtypes **deep-walked rather than leaf-copied**. That last clause is the
ESH-0214d lesson written into the scope up front: natively, the subtypes left as
shallow leaf copies were exactly the ones that had to be reopened later.

Applied in the second commit:

**1. Docs — a ruled walk-back of an overclaim.** The docs promised region
reclamation *engine-neutrally*, which was true of the native engine and false of
the VM. Nothing the native engine does has changed and the VM never reclaimed;
what changes is that the pages now say which engine they mean.
`docs/reference/runtime/memory-model.md` gains a **Which engine reclaims**
section directly under the title — an engine/binary/reclamation table, the
measured RSS pair, and the rule that *unless a section says otherwise, every
reclamation statement on the page describes the native engine* — plus scoping on
the four remaining neutral claims in its body. Also corrected: `README.md`
(release blurb, OALR feature section, both `with-region` samples, the v1.3.1
history line), `docs/API_REFERENCE.md` (the canonical `with-region` definition),
`docs/FAQ.md`, `docs/COMPLETE_LANGUAGE_SPECIFICATION.md` (6.1.1 and the 6.2.1
semantics list), `docs/KNOWN_ISSUES.md`, `docs/reference/runtime/INDEX.md`,
`docs/LONG_RUNNING_LOOPS.md`, `docs/tutorials/WHY_ESHKOL.md`,
`docs/ESHKOL_LANGUAGE_GUIDE.md`, `docs/ESHKOL_V1_ARCHITECTURE.md`,
`docs/internal/ESHKOL_V1_LANGUAGE_REFERENCE.md`.

**Deliberately not edited, reported instead** — `ANNOUNCEMENT.md`, `press/*`,
`CHANGELOG.md` and `RELEASE_NOTES.md` are published or historical records of
shipped releases (correcting them is history revision, not a doc fix), and
`site/static/content/*.html` is generated by `scripts/build-site.sh` and gated by
`scripts/verify_site_release.py`. All four are also modified in the maintainer's
working tree. Exact file/line inventory is in a comment below.

**2. Ledger.** SW-14 moves `SILENT-WRONG` -> `LOUD-LIMITATION`, **stays OPEN**,
disposition *"v1.3.5 flagship: VM OALR Stage-1 evacuator port"*, with the RSS
measurements as evidence. A `LOUD-LIMITATION` bucket is defined in the ledger's
own vocabulary: *a capability the engine does not implement, which announces
itself at the point of use and again when it starts to bite* — distinct from
`LOUD-ERROR`, which fails; a LOUD-LIMITATION succeeds while doing less than the
surface suggests, so the announcement **is** the honesty. Admissible only by
maintainer ruling and only once the announcement is gated by a test that fails if
it is removed.

The move is a change of **category, not a claim that anything was fixed**, and it
is **not a waiver** — waivers expire, this has a build item. The entry carries
`reclassified` (from/to/on/by/why) and `ruling` (date, decision, consequences) so
it is auditable, plus a `close_condition` that forbids closing by assertion: the
fixture's peak RSS must go flat, with region-escape semantics preserved and all
27 subtypes covered by a deep-walk test.

**Effect on the release gate:** SW-14 stops gating the tag.
`scripts/gate_no_silent_wrong.py` blocking count **13 -> 12**; the 12 that remain
are untouched by this branch.

## Rebase onto master (#431, #433, #436, #437, #441)

**One content conflict**, `.icc/silent-wrong-ledger.yaml`: master's SW-26
(closed by #433) landed exactly where this branch inserts SW-25. Resolved as a
union in id order — both closed, both survive. SW-14's LOUD-LIMITATION
reclassification applied cleanly on top. `lib/backend/vm_compiler.c` and
`lib/backend/vm_native.c` auto-merged. No `.swarm/` hunks exist on this branch,
so #441 required nothing.

**Corpus renumbered — the part that would have gone wrong silently.** These two
files were `59_` (vm_parity) and `47_` (differential); master's #433 has since
taken both numbers for `vector_ref_tensor_indexed_access`. They are *added*
files with *different* names at the same index, so **git reports no conflict**:
the tree would have carried two `59_` and two `47_` files and the numbering
would have quietly stopped being a total order. Renumbered to **`62_`** and
**`50_`** — next free on master once #438's pending `61_`/`49_` are accounted
for — with every cross-reference rewritten through a simultaneous sentinel
mapping so no intermediate name could collide. Provenance is recorded in the
SW-25 entry's `renumbered` field, not only in the commit message.

## Gates

Clean rebuild (`cmake --fresh` + `--clean-first`), RelWithDebInfo, LLVM 21,
arm64-apple-darwin24.1.0.

| gate | result |
|---|---|
| `ctest --test-dir build` | **173/173** |
| `scripts/run_vm_parity.sh` | **164/164** |
| `scripts/run_differential.sh` | **306/306** axis pairs over 51 programs |
| `tests/memory/vm_region_growth_watchdog_test.sh` | **8/8** |
| `scripts/gate_no_silent_wrong.py` | **11 -> 10** blockers vs master; single delta is SW-14 |

Re-run on the rebased tree after a clean `--fresh` + `--clean-first` rebuild.
The single blocking delta is SW-14's ruled reclassification to
`LOUD-LIMITATION`; SW-25 was never on master and is added closed.

The one FAIL is `no_open_silent_wrong`, red on this branch **exactly as it is on
`master`**: 13 open unwaived SILENT-WRONG entries, unchanged by this PR (SW-25
moves from absent to `closed`; SW-14 was already open). Verified with
`python3 scripts/gate_no_silent_wrong.py` before and after — same 13 names.

Three probes are worth quoting because they are the ones that would catch this
change going wrong:

- `surface_parity_execution_backed` — **2507 probed / 318 divergences / 0 new**.
  The ratchet holds; this PR adds no surface divergence.
- `language_surface_coverage_floor` — **1091/1091 constructs execution-backed,
  100.00%, policy floor PASS, deficit ratchet PASS, high-risk uncovered 0**.
  This probe first recorded FAIL inside a smoke run that had three other
  worktrees executing the same sweep concurrently on the host; re-run in
  isolation it exits 0 with the numbers above, and every guard it wraps passes
  standalone (`gen_language_surface.py --check` OK,
  `test_language_coverage_gate.py` 17/17,
  `test_runtime_language_coverage.py` 14/14). Recorded as a contention
  false-red, attributed rather than assumed.
- `p8_escape_matrix_green` — all 8 axes PASS, including axis-3 (native-vs-VM
  arity parity ratchet, no NEW divergence) and axis-8 (flat-RSS invariant across
  workload shapes).

The smoke run itself was interrupted at probe 60 by the host; probes 60-62
(`p8_escape_matrix_green`, `event_loop_works`, `no_open_silent_wrong`) were
therefore executed directly and are reported above.

### DIFF
--- a/.icc/silent-wrong-ledger.yaml
+++ b/.icc/silent-wrong-ledger.yaml
@@ -629,12 +629,93 @@ entries:
   - id: SW-14
     bucket: SILENT-WRONG
     status: open
+    resolution: NEEDS-DECISION
     title: "with-region on the VM never reclaims — silent unbounded memory growth"
     observed: "A VM program using regions computes the right answer and never gets the memory back; the VM heap has no escape evacuator"
-    evidence: "docs/KNOWN_ISSUES.md:211"
+    evidence: >-
+      docs/KNOWN_ISSUES.md, section "Region handles and `with-region` on the
+      VM"; reproducer tests/memory/vm_region_growth_watchdog_test.esk; guard
+      gated by tests/memory/vm_region_growth_watchdog_test.sh and the
+      icc-smoke probe vm_region_growth_watchdog
     caught_by: [KNOWN_ISSUES]
     missed_by: [icc-correctness-chain, icc-smoke, icc-readiness]
     notes: "Classified silent because a long-running VM process degrades with no diagnostic and exit 0."
+    measured:
+      at_sha: "2ae3787f"
+      with: "build/eshkol-vm-standalone-test (RelWithDebInfo, LLVM 21, arm64-apple-darwin24.1.0)"
+      fixture: "20000 iterations x 200 conses per iteration, /usr/bin/time -l peak RSS"
+      with_region_wrapper: "1,503,297,536 bytes"
+      without_region_wrapper: "1,504,067,584 bytes"
+      finding: >-
+        The two are the SAME to within 0.06% — the wrapped run is fractionally
+        SMALLER, i.e. the difference is allocator noise, not reclamation.
+        `with-region` does not reclaim
+        less than native — it reclaims NOTHING, and so does everything else:
+        the bytecode VM has no heap reclamation of any kind. heap_alloc()
+        (lib/backend/vm_core.c) bump-allocates from an arena that is never
+        released and hands out a monotonically increasing object index that is
+        never recycled; vm_region_pop() exists (lib/backend/vm_arena.h) but no
+        bytecode ever calls it, because compile_form_with_region()
+        (lib/backend/vm_compiler.c) lowers the form to `begin`.
+    root_cause: >-
+      lib/backend/vm_compiler.c::compile_form_with_region() emits no region
+      bracket at all: the region specifier is recognised and skipped and the
+      body is lowered exactly like `begin`. That is deliberate and documented
+      in the function's own comment — a real push/pop there would free the
+      object graph the body just returned, because the VM heap has no escape
+      evacuator to promote a kept value out with. The missing evacuator, not
+      the missing bracket, is the defect.
+    why_not_fixed_here: >-
+      NEEDS-DECISION: a correct fix is a VM-heap escape evacuator, which is a
+      campaign of the same size as the native ESH-0214c/ESH-0214d work
+      (multiple dedicated PRs), not a patch. It requires, at minimum:
+      (1) ROOT ENUMERATION over the operand stack plus ~15 Value-carrying side
+          tables inside `struct VM` (outputs, handler_stack.promise_mark,
+          current_exception, wind_stack, parameter_bindings, promise_eval_head,
+          lru_caches, event_emitters, channels, timers, exit_handlers, ...) —
+          a missed root frees live memory;
+      (2) PER-SUBTYPE INTERIOR TRAVERSAL for all 27 HeapTypes plus their
+          arena-allocated payloads (VmVector.items, VmString.data,
+          VmBignum.limbs, VmSubstitution.terms, VmFact.args/datum_ptr,
+          VmFactorGraph's five pointer arrays, VmContinuation's six saved
+          arrays, VmParameter.save_stack, ...) — this is exactly the coverage
+          problem ESH-0214d had to close natively, and it still carries a
+          documented watchlist there;
+      (3) a MUTATION WRITE BARRIER at every VM channel that can store a Value
+          into a longer-lived container (~130 candidate sites across
+          vm_native.c/vm_run.c) or, in its place, a full mark from roots;
+      (4) INDEX-SPACE recycling or compaction, because the object table index
+          is a monotone counter — freeing only the arena still leaks 8 bytes
+          per object and still terminates at ESHKOL_VM_HEAP_MAX_SIZE.
+      The failure mode of a partial implementation is STRICTLY WORSE than the
+      defect: a missed root or an uncovered subtype turns an honest leak into
+      a dangling object index, which on the VM aliases a live object and
+      produces a silently wrong VALUE rather than a crash. Shipping that under
+      a SILENT-WRONG closure would invert the ledger's own contract.
+    interim_guard_landed: >-
+      The SILENT half is closed on this branch even though the reclamation is
+      not. Two guards, neither of which changes any answer:
+      (a) the first region form executed on the VM prints a one-time note to
+          stderr naming the boundary (vm_region_reclaim_notice(),
+          lib/backend/vm_core.c; called from compile_form_with_region() and
+          from native 2210 region-open). ESHKOL_VM_REGION_QUIET=1 silences it.
+      (b) the VM arena is sampled every 4096 allocations and crossing
+          ESHKOL_VM_HEAP_BUDGET_MB (default 1024) prints a diagnostic naming
+          the size, the budget and the cause (heap_check_budget()).
+          ESHKOL_VM_HEAP_BUDGET_FATAL=1 makes it exit nonzero so a lane can
+          gate on it; ESHKOL_VM_HEAP_BUDGET_MB=0 disables it.
+      Gated by tests/memory/vm_region_growth_watchdog_test.sh (8/8) and by the
+      icc-smoke probe vm_region_growth_watchdog, so the guard cannot be
+      removed silently. Documented in docs/KNOWN_ISSUES.md,
+      docs/reference/runtime/memory-model.md and
+      docs/reference/runtime/environment-variables.md.
+    decision_requested: >-
+      Choose one: (A) accept the interim guard for v1.3.4 and schedule the
+      VM-heap evacuator as a v1.3.5 campaign — the growth is now loud, named
+      and gate-pinned, and the docs no longer imply otherwise; or (B) hold the
+      tag for the evacuator campaign. Recommendation: (A). Option (B) lands a
+      moving collector in the VM core days before a release, where the cost of
+      an incomplete root or subtype list is a silently wrong value.
 
   - id: SW-15
     bucket: SILENT-WRONG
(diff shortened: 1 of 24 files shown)


tsotchke:
## Out-of-scope finding, surfaced by the mandatory ICC step: the readiness oracle does not parse on `master`

Not touched by this PR, and **not fixed here** — `.icc/completion-oracles.yaml` is
being edited in the maintainer's working tree, so a repair from this branch would
collide. Recording it because it is tag-blocking and currently invisible.

`icc readiness --target eshkol-compiler-readiness` reports:

```
Repo: eshkol-vmmem-435  Status: blocked  Score: 85
- .icc/completion-oracles.yaml: ParserError: while parsing a block mapping
  expected <block end>, but found '<block mapping start>'
  line 217, column 9 -> line 223, column 11
```

**Bisected to the current `master` head:**

| commit | parses? |
|---|---|
| `2ae3787f~1` (#421) | OK |
| `2ae3787f` — *fix(codegen): higher-order builtins ignore shadowing bindings (#429)* | **ParserError** |

#429 added four lines for its new `higher_order_shadowing_oracle` criterion but
omitted the list-item opener, so the keys land inside the *previous* criterion's
`runtime_event:` mapping and duplicate it:

```yaml
        label: 'AD capture scoping: ...'
          event_names: ["higher_order_shadowing_oracle"]     # <- orphaned
          event_values: ["PASS"]
```

The missing three lines are the preceding entry's `action:` plus the new entry's
own opener:

```yaml
        label: 'AD capture scoping: ...'
        action: ./scripts/run_icc_smoke.sh
      - runtime_event:
          event_kinds: [eshkol_smoke]
          event_names: ["higher_order_shadowing_oracle"]
          event_values: ["PASS"]
        severity: high
        label: 'Lexical scope in higher-order builtins: ...'
        action: ./scripts/run_icc_smoke.sh
```

**Why it matters:** the v1.3.4 tag is gated on `icc readiness`, and the oracle
definition that gate reads is unparseable at the exact commit the tag would be
cut from. The gate cannot evaluate the criterion #429 shipped, nor any criterion
after it in the file. Whoever holds that file should take the three lines.

For the record, the rest of the ICC step is green on this branch:
`icc architecture-verify` is **9 PASS / 0 FAIL / 0 UNCHECKABLE across 9
invariants**, including `INV-dispatch-table-completeness` and
`INV-engine-semantic-parity`.

tsotchke:
## Engine-neutral reclaim claims deliberately NOT edited

A full survey found the overclaim in 18 files. The user-facing documentation is
corrected in commit `46380772`. These are the ones I left alone, with the reason
— they need a decision from whoever owns them, not a drive-by edit from a VM
bugfix branch.

### 1. Published / historical release records

Correcting these is history revision, not a documentation fix: they describe what
was claimed at the time of a shipped release. They are also all modified in the
maintainer's working tree right now.

| File | Lines | Claim |
|---|---|---|
| `ANNOUNCEMENT.md` | 3, 9, 25, 78, 175, 201 | "Run an Eshkol daemon for 100,000 ticks and its memory stays flat"; "Automatic memory reclamation matches explicit regions"; the `region-open`/`region-close` announcement; the evacuator paragraphs |
| `press/ESHKOL_PRESS_INFORMATION_SHEET.md` | 107-109, 232-233, 278-282, 289-292 | "byte-for-byte identical to its explicit `with-region` twin"; per-iteration arena-scope reclamation; the escape-evacuator paragraph |
| `press/ESHKOL_DESCRIPTION_COPY.md` | 13, 170, 179-181 | same three claims in short-copy form |
| `RELEASE_NOTES.md` | 63-66, 112-113, 604-607, 658, 1335 | "flat at 34 MB — identical to its explicit `with-region` twin"; "Reverse-mode training loops stay flat under `with-region`" |
| `CHANGELOG.md` | 11-12, 57-61, 908-939 (esp. 935), 1802-1803, 3330 | the v1.3.4 blurb; the `region-open`/`region-close` entry ("peak RSS is flat at 131-132 MB"); the ESH-0214e nursery entry |

**Suggestion, not applied:** rather than rewriting shipped entries, add one line
to the *current* (v1.3.4) section of `CHANGELOG.md` and `RELEASE_NOTES.md`
stating the engine split, and leave the historical entries as the record of what
was said then. `CHANGELOG.md` L782-783 already does exactly this correctly for an
earlier entry — *"VM-side reclamation remains absent and is documented as such"* —
so the house style for it already exists.

### 2. Generated site content

`site/static/content/*.html` is generated by `scripts/build-site.sh` /
`scripts/build-site-content.sh` from the markdown, and the deploy is gated by
`scripts/verify_site_release.py`. Hand-editing the HTML would be overwritten on
the next regeneration and risks the release-contract string check that took Deploy
Site red for a week once before.

- `site/static/content/api_reference.html` L340-345
- `site/static/content/complete_language_specification.html` L1845-1852, L3938
- `site/static/content/announcement.html` L9, L52-65, L459, L720-722

**These will pick up the corrections automatically** once the site is regenerated
from the now-corrected markdown — except the `announcement.html` ones, which
mirror `ANNOUNCEMENT.md` and therefore depend on decision (1) above. For the
record, `verify_site_release.py` passes on this branch as-is:
`PASS: website matches 15 platform packages + SHA256SUMS.txt, the publication loader is present, and VM parity wording is scoped`.

### 3. Vision / design / narrative docs — low priority

Arena-level "memory reclaimed via arena reset" phrasing, in documents that only
ever discuss the native engine. Listed for completeness; correcting them is
optional and I would not spend a reviewer's time on it.

- `docs/vision/SCIENTIFIC_COMPUTING.md` L76, L89, L451
- `docs/vision/DIFFERENTIATION_ANALYSIS.md` L307-308, L470-471
- `docs/vision/AI_FOCUS.md` L625
- `docs/vision/TECHNICAL_WHITE_PAPER.md` L83
- `docs/breakdown/OVERVIEW.md` L145; `docs/breakdown/COMPILER_ARCHITECTURE.md` L841; `docs/breakdown/MACHINE_LEARNING.md` L596; `docs/breakdown/CONSCIOUSNESS_ENGINE.md` L332
- `docs/design/adr/0001-oalr-concurrent-resident.md` — already hedges at L45 and L438-450; no change needed

### Already correct before this PR

`docs/reference/runtime/environment-variables.md` L77-80 and L87, and
`CHANGELOG.md` L782-783 — these were the model wording the corrections above were
written to match.

PR #438, opened 2026-08-13
fix(vm): pop do-loop variables and box captured ones (LE-10, SW-34)

tsotchke:
Two VM defects from PR #437's class-kill, both fixed at the root, plus a
pre-existing third that the second one exposed. Everything here is in
`lib/backend/vm_compiler.c`; no native codegen is touched.

Based on `origin/master`, **not** stacked on #435 — the fix needs only
`needs_boxing()`, which is already on master. #435's
`vm_box_mutable_captured_params()` is for parameters, which arrive from the
caller; a `do` variable is bound like a `let` variable and boxes at its binding
site.

## LE-10 — the loop variables were never popped, and it is wider than filed

`compile_form_let()` ends with `OP_POPN`. `compile_form_do()` ended by
decrementing the compiler's own `c->n_locals` — a **compile-time** bookkeeping
change with no runtime effect — so every `do` form stranded one operand-stack
value per loop variable. Top-level bindings live in stack slots handed out by
counting (`add_local`), and the top-level driver emits exactly one `OP_POP` per
expression that grew no local, so the stranded values shift every later slot.
Same slot-shift corruption already documented on `compile_form_require()` and
`compile_form_with_region()` in this file, reached a third way.

**Re-measurement changed the entry.** It was filed LOUD (heap-object limit,
`calling non-function`, hangs) on the shapes #437's class-kill happened to hit.
The same root cause also produces **silent wrong values**, and needs *neither a
closure nor a capture*:

```scheme
(do ((i 0 (+ i 1))) ((= i 2)) 1)
(define q 42)
(display q)              ; VM 2, native 42 — exit 0, no diagnostic
```

Two ordinary sequential `do` loops diverge the same way (VM `3` then `13`,
native `3` then `40`). The loud shapes are the same corruption landing on a slot
whose stranded junk happens to be a loop counter or a callee. **Closing this
against only the filed reproducer would have left the `do`-then-`define` case
shipping.**

**Fix.** Emit `OP_POPN` for the loop variables at the exit, keeping the result
on top, exactly as `compile_form_let()` does; compile the result expressions out
of tail position when there are locals to pop, since a `TAIL_CALL` would skip
the `OP_POPN`.

**Second hazard removed by the same rewrite.** The function used to emit a
complete first version of the loop and then "discard" it with
`c->code_len = loop_top` — leftover debugging scaffolding, its reasoning about
`JUMP_IF_FALSE` polarity still sitting in the source as commentary. That
compiled the body and steps **twice**. Rewinding `code_len` un-emits CODE and
nothing else: the discarded pass had already appended to the constant pool,
registered upvalues, written closure PC constants pointing into the region about
to be overwritten, and pushed entries onto the global forward-reference repatch
arrays. The correct structure is now emitted once.

## SW-34 — a captured and mutated loop variable was captured by value

```scheme
(define (f n) (do ((i 0 (+ i 1)) (a 0)) ((= i n) a)
                  ((lambda () (set! a (+ a 1))))))
(f 5)                    ; VM 0, native 5
```

`compile_form_let()` has always run `needs_boxing()` over its bindings; a `do`
variable was bound with a bare `add_local()`, so `OP_CLOSURE` copied its VALUE
into the closure's upvalue array and the inner `set!` mutated a private copy.
The same loop **at top level was already correct**, which is the tell: only the
root chunk has the open-slot relay (native calls 151/252), and that relay is
deliberately restricted to the root chunk because a function's frame dies on
return.

**Fix.** Box every loop variable `needs_boxing()` reports as
mutated-and-captured, scanning the test, the results, the body **and** every
step expression, since a `do` variable is in scope for all four. A boxed
variable's step store goes through `VEC_SET` into the same cell. Step values are
held in anonymous locals first, so R7RS 4.2.4 parallel update (every step sees
the pre-step values) is preserved while still giving `VEC_SET` its box and index
*under* the value.

This is the fourth binding form of the SW-25 family — parameters, `let*`,
`letrec`, `letrec*` were closed by #435; `do` is closed here. Same root cause
throughout: a binding mutated across a closure boundary needs a shared **cell**,
and only `let` ever had one.

## Residue, fixed at the root: `compile_form_set` left two values

Boxing a `do` variable exposed a **pre-existing** defect. Both boxed paths of
`compile_form_set()` (local and upvalue) left **two** values on the operand
stack — `OP_VEC_SET` pops three and pushes NIL, and the shared tail pushes the
NIL that `set!` evaluates to — where every caller pops one.

Inside a frame about to `RETURN` the surplus was absorbed, which is why it had
never been seen. In a `do` body, whose inline statements are each followed by
exactly one `OP_POP`, it accumulated once per iteration until the operand stack
overflowed. Found by disassembly, not by guessing:

```
[5220] VECST  0     ; box[0] = a+1  -> pushes NIL
[5221] NIL    0     ; set! returns void -> pushes another
[5222] POP    0     ; only one taken
```

## Class-kill

`tests/vm_parity/corpus/61_do_loop_variable_binding.esk` — **23 shapes,
byte-identical VM-vs-native**: sequential and nested loops, a binding made after
a loop, depths 1-3, stored and immediately-invoked closures, two closures
sharing one variable, a loop inside a lambda and inside a define with a runtime
trip count, parallel update, tail position, multiple result expressions, and
controls (no closure, a shadowing body parameter, an empty body).
`tests/differential/corpus/49_do_loop_variable_binding.esk` covers the native
axes.

**Four native-side shapes were held out** while this branch was written, each
with its measured pair recorded in the corpus SCOPE comment. **They are now
promoted.** #437 landed on master and fixed the native side of all four; with
both changes present the engines agree on every one — at exactly the values the
VM already produced:

| shape | agreed value | native before #437 |
|---|---|---|
| step that must see a body closure's mutation | **33** | 30 |
| capture in the **step** | **8** | 0 |
| capture in the **result** | **10** | LLVM module verification failed |
| read-only capture | **6** | no output |

All four are asserted in both corpus files and the exclusion list is empty.
Writing the measured pairs down rather than only naming the shapes is what made
this a re-run instead of a re-derivation.

**Revert-validated.** With `lib/backend/vm_compiler.c` restored, the corpus file
**hangs** on the VM (killed at 25s), and the individual shapes answer `2`, `00`,
`2` and `0` where the fixed VM and native both answer `42`, `50`, `6` and `4`.

## Ledger

SW-34 and LE-10 both **closed** with measured evidence. Neither existed on
`master` (both were opened by #437, unmerged), so both are added here as closed
rather than edited in place. LE-10's entry records that re-measurement found it
wider and partly silent than filed, and why it was closed on the wider evidence.
`gate_no_silent_wrong.py` blocking count is **unchanged at 13**.

## Gates

Rebased onto master (which absorbed #431, #433, #436, #437), then a clean
`--fresh` + `--clean-first` rebuild. RelWithDebInfo, LLVM 21,
arm64-apple-darwin24.1.0.

**Rebase:** one conflict, `.icc/silent-wrong-ledger.yaml`, resolved as a union
with closed entries winning — master's LE-09 (closed by #437) kept, and master's
**open** SW-34 and LE-10 superseded in place by the closed entries here, with
master's filed reproducers and measured shapes folded in so no evidence is lost.
`lib/backend/vm_compiler.c` rebased without conflict.

| gate | result |
|---|---|
| `ctest --test-dir build` | **173/173** |
| `scripts/run_vm_parity.sh` | **164/164** |
| `scripts/run_differential.sh` | **306/306** axis pairs over 51 programs |
| `scripts/gate_no_silent_wrong.py` | **11 -> 10** blockers vs master; single delta is SW-34 |

`scripts/run_icc_smoke.sh` was **not** run as a whole, reported rather than
papered over: the host had three other worktrees compiling and a 1-minute load
average of 7-13, and PR #435 already recorded one contention false-red from
exactly that situation. Path-scoped probes were run verbatim from the smoke
script instead, chosen for what this diff touches:

| probe | result |
|---|---|
| `closure_set_tco_loop_oracle` (ESH-0094 — closure in a TCO loop `set!`s a capture) | PASS |
| `stdlib_sort_filter_scale_oracle` | PASS |
| `higher_order_shadowing_oracle` (JIT) | PASS |
| `ad_forward_over_reverse_oracle` | PASS |
| #437's own `rest_capture_and_do_closure_test.esk` | PASS |
| `define_loop_flat_rss_aot` | PASS |

### DIFF
--- a/.icc/silent-wrong-ledger.yaml
+++ b/.icc/silent-wrong-ledger.yaml
@@ -1001,48 +1001,72 @@ entries:
 
   - id: SW-34
     bucket: SILENT-WRONG
-    status: open
-    title: "VM: a `do` loop variable captured and mutated by a body closure loses the mutation inside a procedure and in a nested loop"
-    repro: >-
-      (define (f n) (do ((i 0 (+ i 1)) (a 0)) ((= i n) a) ((lambda () (set! a (+ a 1)))))) (display (f 5))
-    observed: "VM 0 — exit 0, no diagnostic"
-    correct: "5 — native answers 5 after the LE-09 fix"
-    affected_measured:
-      - "(define (f n) (do ((i 0 (+ i 1)) (a 0)) ((= i n) a) ((lambda () (set! a (+ a 1)))))) (f 5) -> VM 0, native 5"
-      - "(do ((i 0 (+ i 1)) (t 0)) ((= i 3) t) (do ((j 0 (+ j 1))) ((= j 2)) ((lambda () (set! t (+ t 1)))))) -> VM 2, native 6"
-    unaffected_measured:
-      - "the same loop at TOP LEVEL: (do ((i 0 (+ i 1)) (acc 0)) ((= i 3) acc) ((lambda () (set! acc (+ acc i))))) -> VM 3, native 3"
-      - "a read-only capture: (do ((i 0 (+ i 1)) (s 0)) ((= i 4) s) (set! s (+ s ((lambda () i))))) -> VM 6, native 6"
-    discovered_by: >-
-      class-killing LE-09 on the native side. Every `do` shape the native fix
-      had to handle was run on both engines; these two are the ones where the VM
-      answers wrongly and native answers correctly.
-    not_fixed_here: >-
-      NOT a native defect and NOT closed by branch fix/native-rest-capture-and-do-loop,
-      which changes lib/backend/llvm_codegen.cpp only. Measured against BOTH the
-      VM at this SHA and the VM on the SW-25 branch fix/vm-region-and-upvalue:
-      identical wrong answers, so PR #435's parameter/let boxing does not reach
-      `do` bindings. The likely shape of the fix is the same mechanism applied to
-      compile_form_do()'s bindings — the SW-25 entry's own root_cause names
-      compile_form_let() as the only binding form that had ever run
-      needs_boxing() — but that is a VM-compiler change and wants its own
-      measurement lane.
+    status: closed
+    closed_at: "2ae3787f"
+    closed_by: "branch fix/vm-do-loop-capture"
+    title: "A `do` loop variable captured and mutated by a body closure loses the mutation inside a procedure"
+    repro: "(define (f n) (do ((i 0 (+ i 1)) (a 0)) ((= i n) a) ((lambda () (set! a (+ a 1)))))) (display (f 5))"
+    observed: "VM 0"
+    correct: "5 — native answers 5"
+    cross_engine: >-
+      the nested-loop form
+      (do ((i 0 (+ i 1)) (t 0)) ((= i 3) t) (do ((j 0 (+ j 1))) ((= j 2)) ((lambda () (set! t (+ t 1))))))
+      was VM 2, native 6. The same loop at TOP LEVEL was already correct on the
+      VM (3), which is the tell: only the root chunk has the open-slot relay.
+    observed_after: "VM 5 and VM 6, byte-identical to native across 19 shapes"
+    root_cause: >-
+      lib/backend/vm_compiler.c::compile_form_do() bound its loop variables
+      with a bare add_local(). compile_form_let() has always run needs_boxing()
+      (lib/backend/vm_parser.c) over its bindings, so a `let` variable that is
+      both set!-mutated and captured by a nested lambda lives in a shared
+      1-element vector; a `do` variable never got that treatment, so OP_CLOSURE
+      copied its VALUE into the closure's upvalue array and the inner set!
+      mutated a private copy discarded at the end of the iteration. The
+      top-level case was correct only because the root chunk aliases an
+      absolute VM stack slot through the open-slot relay (native calls
+      151/252), which is deliberately restricted to the root chunk because a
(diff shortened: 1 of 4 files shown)


tsotchke:
## Heads-up for PR #435: corpus filename collisions after this master refresh

Not an issue for this PR — recording it here because I found it while checking
numbering for the rebase, and it will bite #435 the moment it is rebased.

Master's #433 took the numbers #435 is already using:

| #435 has | master now has |
|---|---|
| `tests/vm_parity/corpus/59_mutable_capture_writeback.esk` | `tests/vm_parity/corpus/59_vector_ref_tensor_indexed_access.esk` |
| `tests/differential/corpus/47_mutable_capture_writeback.esk` | `tests/differential/corpus/47_vector_ref_tensor_indexed_access.esk` |

These are *added* files with different names at the same index, so git will not
report a conflict — both will simply coexist and the corpus will carry two `59_`
and two `47_` files. That is not fatal (the runners glob the directory), but the
numbering stops being a total order, which is the only thing it is for.

Next free after master: `62_` for `tests/vm_parity/corpus/`, `50_` for
`tests/differential/corpus/` — this PR takes `61_` and `49_`, so #435 should
renumber to `62_` / `50_` when it rebases.

PR #442, opened 2026-08-13
feat(testing): the value-position axis — close SW-34/SW-35, kill the class that produced them

tsotchke:
**Stacked on #439** (`fix/numeric-tower-bignum-tail`). Base this on #439's branch, not master; merge #439 first. If #439 rebases, this rebases with it.

## What this is

The **value-position axis** — the gate the escape analyses for SW-27 and SW-35 both prescribed, in the same words — plus the two defects that prompted it and two more the gate found on its first run.

Eshkol lowers most builtins **inline at the call site**. Referencing the same builtin as a **value** takes a different route through the codegen — in fact **two** different routes — and defects keep being found living there while call position is correct: LE-01, SW-27, SW-31, SW-34, SW-35. Five, each found by hand.

That invisibility is structural, not bad luck:

- the differential corpus compares **execution axes against each other**, and a value-position defect is usually wrong *identically* on all four, so they agree and the gate is green;
- vm-parity excuses native-only programs;
- the arity sweep, surface-parity probe and language-coverage floor all exercise builtins in **call position only**.

## The sweep — `scripts/run_value_position_sweep.py`

For every builtin the manifest can type, one program evaluates the **same call** in call position and through three value-position routes, and compares them in-program:

| probe | route | codegen site |
|---|---|---|
| `passed` | `(h N a…)` — user higher-order procedure | `codegenVariable` |
| `stored` | `(let ((v N)) (h v a…))` | `codegenVariable` |
| `returned` | `((lambda () N))` via a HOF | `codegenVariable` |
| `mapped` | `(car (map N (list a)))` — builtin combinator | `resolveLambdaFunction` |

**The oracle is differential by construction.** Nothing hard-codes what a builtin ought to return, so it needs no reference implementation and cannot pass by agreeing with a wrong expectation. A builtin is judged only when its **call position works**; if that raises, the name is skipped and reported — reporting someone else's bug as your own finding is how a gate loses meaning.

`mapped` is **not** redundant with `passed`. SW-35 was live in *both* routes: after the first fix `(h floor 7/3)` was correct while `(map floor (list 7/3))` still printed a heap address. A gate probing one route would have certified that half-fix as complete.

The **VM is in `--all-axes` on purpose** — an independent implementation of the same question, so it catches what four agreeing native axes cannot. It immediately did.

## SW-35 — rounding as a value returned a rational's heap address

`floor`/`ceiling`/`truncate`/`round` had **two implementations**. Call position (`codegenMathFunction`) dispatches on **exactness**: exact integers and bignums are the identity, rationals go through `eshkol_rational_<op>_tagged`, only inexact reals reach libm. The value route used `createBuiltinUnaryMathFunction`, which unpacks the operand as a double — for a rational, that reads its **pointer**:

```
(map floor (list 7/3 -7/3 (/ (expt 2 100) 3)))
  => (4966570936 4966571152 4966571728)   addresses, varying per run
(floor 7/3) => 2                          call position, correct
```

Both routes now go through the generic wrapper that **generates its body by re-entering `codegenCall`**, so the exactness-aware lowering *is* the wrapper body. That is the LE-01 mechanism applied to a second instance of the LE-01 shape, not a second patch to the duplicate. The ledger's stated correct answer `(2 -3 422550200076076467165567735125)` is reproduced exactly.

## SW-34 — `complex?` could not be used as a value at all

Loud compile-time `Undefined variable: complex?` the moment the bare name was read as a value, while call position answered `#t` and the VM bound it fine. No row in the first-class builtin table. The fix lists the R7RS numeric-tower predicates as a **family** rather than name-by-name as need arises — name-by-name is how the gap appeared. The predicate matrix no longer needs its `complex?` exclusion.

## Two more, found by the sweep on its first run, fixed here

- **SW-36** — `make-string` as a value **silently dropped a supplied fill character**: `(h make-string 3 #\x)` → `"   "`, exit 0. The table LE-01 introduced listed it at arity 1, so a fixed-arity closure never passed the second argument. Live on master since #427. Corrected to the full form; R7RS 6.7 leaves the 1-argument contents *unspecified*, so wrapping the form whose result **is** specified is the right choice.
- **SW-37** — `pow` as a value **SIGSEGV'd**: LE-01 residue, a raw libm symbol reaching the foreign-ABI fallback.

## Two found and deliberately NOT fixed here

Both are in the **VM's own** value representation — a different subsystem from this lane's native lowering. Filed precisely and ratcheted rather than fixed badly:

- **SW-38** — VM only: `cadr`/`caddr` as first-class values return `()`. **Silent wrong**, exit 0, and `()` is truthy in Scheme. Native correct on all four axes.
- **SW-39** — VM only: `string-length`/`string-ref`/`vref` as first-class values **abort the VM**. Native correct on all four axes.

**SW-38 is an open SILENT-WRONG entry and therefore gates the release.** That is the honest detector working on the day it was built. It needs a coordinator ruling — dispatch a VM lane, or accept the red — not a quiet reclassification.

## Wired as a permanent gate

- **ctest** — `value_position_axis_sweep`, plus `first_class_predicates_{runtime,aot,vm}_smoke` and `first_class_rounding_{runtime,aot}_smoke`
- **smoke** — probe `value_position_axis` in `scripts/run_icc_smoke.sh`
- **oracle** — two criteria in `.icc/completion-oracles.yaml`, on the sweep's own `runtime_event` and on the smoke probe

## Coverage is stated out loud

**199 of 205** selected builtins are actually compared; **6 are not exercised** and the sweep prints them by name every run. "0 findings" over a table this size invites being read as "the surface is proven", and it is not — an unexercised name is an untested name.

## Measured

| Gate | Result |
|---|---|
| `ctest` | **170/170 PASS** |
| value-position sweep | **2352 comparisons over 5 axes, 0 NEW findings**, 17 known (all SW-38/SW-39, all VM) |
| `run_vm_parity.sh` | **172 passed / 0 failed** |
| differential corpus | **294/294 axis pairs across 49 programs** |

## Ledger ID collision

SW-34 was allocated concurrently by more than one lane. This PR's closure cites the entry **by branch and text** — "complex? cannot be used as a first-class value natively, though the VM binds it", filed on `fix/numeric-tower-bignum-tail` — and speaks only to that one. New entries use SW-36..SW-39, verified free across every remote branch. No other lane's entries were renumbered.

### DIFF
--- a/.icc/completion-oracles.yaml
+++ b/.icc/completion-oracles.yaml
@@ -251,6 +251,38 @@ oracles:
         action: ./scripts/run_icc_smoke.sh
 
       # ─────────────────────────────────────────────────────────────
+      # THE VALUE-POSITION AXIS (scripts/run_value_position_sweep.py).
+      #
+      # Builtins are lowered inline at the call site; referencing one as a
+      # VALUE goes through different codegen. Four defects have been found
+      # living there while call position was correct — LE-01 (no value
+      # representation at all), SW-27 (rest-arg procedures lost their
+      # variadic flag, silently), SW-31, SW-35 (rounding returned a
+      # rational's heap address). None was caught by any criterion above,
+      # and structurally none could be: the differential corpus compares
+      # execution AXES against each other, and these defects are wrong
+      # identically on every axis.
+      #
+      # The sweep's oracle is differential BY CONSTRUCTION — call position
+      # versus value position for the same call in the same program — so it
+      # asserts no expected values and cannot certify a wrong answer.
+      # ─────────────────────────────────────────────────────────────
+      - runtime_event:
+          event_kinds: [runtime_event]
+          event_names: ["value_position_sweep"]
+          event_values: ["PASS"]
+        severity: high
+        label: 'the VALUE-POSITION axis is clean: every builtin the manifest can type answers the same referenced as a value — passed to a higher-order procedure, stored, returned, reached through a builtin combinator — as it does in call position, on every execution axis'
+        action: python3 scripts/run_value_position_sweep.py
+      - runtime_event:
+          event_kinds: [eshkol_smoke]
+          event_names: ["value_position_axis"]
+          event_values: ["PASS"]
+        severity: high
+        label: 'the value-position axis probe is wired into the smoke harness, so the blind spot that produced SW-27/SW-31/SW-34/SW-35 one at a time is gated rather than rediscovered'
+        action: ./scripts/run_icc_smoke.sh
+
+      # ─────────────────────────────────────────────────────────────
       # P8 "escape-closure" pillar (scripts/run_p8_escape.sh). Each axis
       # is built so that a 2026-07 externally-reported bug CLASS would have
       # been caught by our own framework first; retro-catch evidence
--- a/.icc/silent-wrong-ledger.yaml
+++ b/.icc/silent-wrong-ledger.yaml
@@ -1993,6 +1993,250 @@ entries:
     caught_by: [pr-record-416]
     missed_by: [icc-correctness-chain, icc-smoke, icc-readiness]
 
+  - id: SW-41
+    bucket: PARITY-RATCHET
+    status: closed
+    closed_at: "fix/value-position-axis"
+    closed_by: "PR stacked on #439 (branch fix/value-position-axis)"
+    title: "complex? cannot be used as a first-class value natively, though the VM binds it"
+    repro: "(display (map complex? (list 1 2)))"
+    observed: >-
+      native -r fails to compile: "Undefined variable: complex?" followed by
+      "Failed to resolve procedure for map". The VM resolves it and answers (#t #t).
+    correct: "both engines should bind it; it is an R7RS 6.2.1 procedure"
+    measured: >-
+      In CALL position (complex? 5) works on both engines. Only the value position
+      diverges. Found while building the SW-31 predicate matrix, which had to drop
+      complex? from its first-class section for exactly this reason — the file says so in
+      a comment rather than quietly omitting it.
+    caught_by: [direct-measurement, sw-31-corpus-authoring]
+    missed_by: [icc-correctness-chain, icc-smoke, icc-readiness, ctest, vm-parity, differential-corpus]
+    notes: >-
+      Classified PARITY-RATCHET rather than SILENT-WRONG: native fails LOUDLY at compile
+      time, so nothing computes a wrong answer. It is a binding-table gap on the native
+      side, and it blocks one row of the predicate matrix from being gated.
+    root_cause: >-
+      complex? had no row in the codegen's first-class builtin table
+      (lookupInlineBuiltin, lib/backend/llvm_codegen.cpp). Reading a bare builtin name as
+      a value consults that table; a name absent from it falls through to
+      "Undefined variable". Call position never consults it, which is why the two
+      positions disagreed.
+    fix: >-
+      The R7RS numeric-tower predicate family is now listed as a FAMILY — number?,
+      complex?, real?, rational?, integer?, exact?, inexact?, exact-integer?, nan?,
+      infinite?, finite? — together with the type predicates, rather than name-by-name
+      as need arises. Name-by-name is how the gap appeared, so the fix removes the
+      mode of failure and not just the instance.
+    verified_at: >-
+      tests/integration/first_class_predicates_test.esk, gated on the JIT lane, the AOT
+      lane and the VM (first_class_predicates_{runtime,aot,vm}_smoke). The oracle is
+      differential: every predicate is compared against the identical application in call
+      position in the same program, through BOTH value-position routes (a user
+      higher-order procedure, and map). The predicate matrix no longer needs its
+      complex? exclusion.
+    id_collision_note: >-
+      The ID SW-34 was allocated concurrently by more than one lane. THIS closure refers
+      to the entry whose title is "complex? cannot be used as a first-class value
+      natively, though the VM binds it", filed on branch
+      fix/numeric-tower-bignum-tail with repro "(display (map complex? (list 1 2)))".
+      It does not speak to whatever else may carry the number SW-34 elsewhere;
+      reconciliation happens at merge.
+    renumbered_from: >-
+      Filed as SW-34 on branch fix/numeric-tower-bignum-tail (PR #439), which did not carry
+      the entry into master. Renumbered TWICE as master moved underneath this PR, each time
+      because the number had been taken by a defect that reached master first and which is
+      not renumbered here:
+        SW-34 -> SW-40   master's SW-34 is #438's "`do` loop variable captured and mutated
+                         by a body closure loses the mutation"
+        SW-40 -> SW-41   master's SW-40 is #443's "a derivative taken through min/max is
+                         silently 0 on the VM"
+      Cite this defect as SW-41. Note for anyone auditing the second hop: git reported NO
+      conflict on that rebase, because the two SW-40 entries landed in different regions of
+      the file. A clean auto-merge is not evidence of a correct ledger; the duplicate was
+      caught by parsing the YAML and counting IDs, which is the check worth keeping.
+
+  - id: SW-35
+    bucket: SILENT-WRONG
+    status: closed
+    closed_at: "fix/value-position-axis"
+    closed_by: "PR stacked on #439 (branch fix/value-position-axis)"
+    title: "Natively, floor/ceiling/truncate/round used as FIRST-CLASS VALUES return a rational's heap pointer as a number"
+    repro: "(display (map floor (list 7/3 -7/3 (/ (expt 2 100) 3))))"
+    observed: >-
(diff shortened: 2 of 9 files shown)


tsotchke:
## Gate results at `b49a8240`

Built once and reused, per the machine note (load ~12, disk at 99% with ~20 GiB free).

| Gate | Result |
|---|---|
| `ctest --test-dir build` | **170/170 PASS** |
| value-position sweep, all 5 axes | **2352 comparisons, 0 NEW findings**, 17 known (all SW-38/SW-39, all VM) |
| `scripts/run_vm_parity.sh` | **172 passed / 0 failed** |
| `scripts/run_differential.sh` | **294/294 axis pairs across 49 programs** |
| `scripts/run_icc_smoke.sh` | **60/62 pass** — see the two below |

### The new probe passes

```
✓ value_position_axis   every builtin answers the same when referenced as a VALUE
                        (passed to a higher-order procedure, stored, returned,
                        reached through map) as it does in call position
```

### The two red probes, both honestly

**1. `language_surface_coverage_floor` — exit 2, the documented false-red.** That probe requires a second `-DESHKOL_QUANTUM_ENABLED=ON` toolchain (`build-quantum/eshkol-run`) which was deliberately not built: the machine note said build once and reuse, and the disk is at 99%. Not a regression, and not something this branch touches.

**2. `no_open_silent_wrong` — exit 1.** This gate was **already red on this branch before my commit**. It lists 12 open SILENT-WRONG entries:

```
SW-05 SW-10 SW-11 SW-12 SW-14 SW-19 SW-20 SW-21 SW-23 SW-30 SW-33 SW-38
```

Eleven are pre-existing. **SW-38 is mine** — the VM `cadr`/`caddr` first-class defect the new sweep found on its first full run. I filed it rather than fixing it, because it lives in the VM's own value representation and this lane works on the native lowering.

So the accurate statement is: **this branch did not turn a green gate red; it added one true entry to an already-red list.** Correcting my own commit message on that point — it implied the gate flipped, and it did not.

The coordinator ruling still stands as needed: dispatch a VM lane for SW-38/SW-39, or accept them as known. Either is fine; quietly reclassifying SW-38 out of SILENT-WRONG is not, since `()` is both a wrong answer and truthy in Scheme.

### Coverage, stated rather than implied

**199 of 205** selected builtins are actually compared. **6 are not exercised** — `caar`, `hash-keys`, `hash-table-keys`, `hash-table-values`, `hash-values`, `string->i128` — and the sweep prints them by name on every run. An unexercised name is an untested name; the fix is to extend `ARG_OVERRIDES`, not to let "0 findings" read as "the surface is proven".

tsotchke:
## Rebased onto master (`438ea1fc`) — retargeted, conflicts resolved, all gates re-measured

Replayed with `git rebase --onto origin/master b0743134`, where `b0743134` was the old #439
tip. Only this branch's own commit was replayed; the branch is now **1 commit on master**.

### Conflicts hit: two

| File | Class | Resolution |
|---|---|---|
| `CMakeLists.txt` | content, same insertion point | **UNION.** #438 added the `rest_capture_and_do_closure_*` gates and this branch added the first-class / value-position gates at the same place. Disjoint additions; both kept. Verified all 8 test names present. |
| `.icc/silent-wrong-ledger.yaml` | content, two hunks | **Master is the base**, this branch's entries re-appended under free IDs. Details below. |

`lib/backend/llvm_codegen.cpp` **auto-merged**, and I verified rather than assumed it:
`diff` of #439's tagged-rounding call site between master and this tree is **byte-identical**,
so the `eshkol_rational_<op>_tagged` out-param signature (`void(ptr arena, ptr rational,
ptr out)`) is untouched and this branch's wrapper routing still targets exactly it. Both
routing sites (`codegenVariable` line 11282, `resolveLambdaFunction` line 38492) survived,
and the table rows for `floor`/`pow`/`complex?`/`make-string` are intact.

### The ID collision map was thicker than expected

Two things had to be established by reading master rather than trusting the numbers:

1. **Master's `SW-34` is #438's defect** — "a `do` loop variable captured and mutated by a
   body closure loses the mutation" — **not** the `complex?` entry this branch closed. #438
   reached master first, so it keeps the number and is **not renumbered**.
2. **#439's merge did not carry either of this branch's base entries.** Master has no
   `SW-35` at all, and no entry for `complex?`. The coordinator's note that master's
   closures "adopted your entries' credit" does not hold for these two — they simply were
   not in the squash.

Resolution, renumbering only what actually collides and carrying provenance in the file:

| entry | old ID | new ID | why |
|---|---|---|---|
| `complex?` not usable as a value | SW-34 | **SW-40** | master's SW-34 is #438's; `renumbered_from:` records the change and points readers at #438 for the number |
| rounding returns a rational's heap address | SW-35 | **SW-35** (kept) | SW-35 was free on master; `id_provenance:` says so explicitly |
| `make-string` dropped a fill char | SW-36 | SW-36 | free on master (max was 34) |
| `pow` SIGSEGVs as a value | SW-37 | SW-37 | free |
| VM `cadr`/`caddr` return `()` | SW-38 | SW-38 | free |
| VM `string-length`/`string-ref`/`vref` abort | SW-39 | SW-39 | free |

Verified mechanically: **82 entries, zero duplicate IDs**, YAML parses.

### `SW-38` / `SW-39` baseline entries retained

Per the coordinator's note, the 17 `tests/value_position/BASELINE.json` entries covering the
VM first-class families are **left in place**. They are the ratchet for codex PR #445 and
drop when #445 merges, not before — removing them now would make the gate silently stop
watching the thing #445 is fixing.

### Gates re-measured after a full rebuild

Rebuilt from scratch before running anything.

| Gate | Before rebase | After rebase |
|---|---|---|
| `ctest` | 170/170 | **180/180 PASS** |
| value-position sweep, 5 axes | 2352 comparisons, 0 NEW | **2372 comparisons, 0 NEW**, 17 known (all VM, SW-38/39) |
| `run_vm_parity.sh` | 172 / 0 | **180 passed / 0 failed** |
| differential corpus | 294/294 across 49 | **318/318 axis pairs across 53 programs** |
| SW-35 tower probe | correct | `(2 -3 422550200076076467165567735125 2 5)` — correct, bignum included |
| first-class predicates | native + VM green | **native + VM green** |

Counts rose because master's merges added corpus programs, gates and one builtin. Coverage:
**200 of 206** builtins compared; the same 6 remain unexercised and are printed by name on
every run.

`gh pr view 442` → `mergeable: MERGEABLE`, base `master`, 1 commit.

tsotchke:
## Final hop: rebased onto `18e09a3e` (#443) — MERGEABLE

### git reported NO conflict, and that was the problem

The rebase onto master applied cleanly. It also produced a **duplicate `SW-40`** — mine
(`complex?` first-class, PARITY-RATCHET) and #443's (min/max tangent drop, SILENT-WRONG) —
because the two entries live in different regions of the file, so there was nothing for git
to flag.

A clean auto-merge is not evidence of a correct ledger. The duplicate was caught by parsing
the YAML and counting IDs, which is the check worth keeping; it is now written into the
entry's provenance so the next person renumbering here knows not to trust a green rebase.

### Renumbered, with both hops recorded

`SW-40` -> **`SW-41`**, the next truly-free ID (master's max is 40). Master's SW-40 is
#443's and is **not** renumbered. The entry's `renumbered_from:` now records the whole
chain rather than only the latest move:

```
SW-34 -> SW-40   master's SW-34 is #438's do-loop capture defect
SW-40 -> SW-41   master's SW-40 is #443's min/max tangent defect
```

Cite this defect as **SW-41**. Verified mechanically: **83 entries, zero duplicate IDs**,
YAML parses. No stale `SW-40` references remain in the baseline, tests, sweep or codegen.

### Code: auto-merged, and verified rather than assumed

#443 touched `lib/backend/vm_native.c`, `lib/backend/vm_run.c` and two corpus files. This
branch touches `lib/backend/llvm_codegen.cpp`. Disjoint, as expected — and confirmed: every
one of this branch's six code/config files is **byte-identical** to the pre-rebase commit
(`git diff 91c9f896 HEAD -- <file>` empty for each).

### I did not skip the rebuild, and here is why

The instruction allowed skipping it since only the ledger conflicted. I rebuilt anyway,
incrementally: **#443 changed the VM**, and this branch's sweep has a **VM axis** whose
17 baseline entries are assertions about VM behaviour. Running that axis against a binary
built before #443 would have been measuring the old VM — the exact stale-binary trap this
lane hit once already and wrote down. The incremental build cost about a minute.

| Gate (re-run against the post-#443 VM) | Result |
|---|---|
| value-position sweep, **VM axis** | **692 comparisons, 0 NEW findings**, 17 known (SW-38/SW-39) |
| `scripts/run_vm_parity.sh` | **182 passed / 0 failed** (was 180 — #443 added two corpus programs) |
| ctest, first-class + value-position + rest-capture lanes | **20/20 PASS** |

The full native sweep, ctest 180/180 and differential 318/318 from the previous hop stand
unchanged: no native code moved, and every native file is byte-identical.

`SW-38`/`SW-39` baseline entries are still in place, dropping when codex #445 merges.

`gh pr view 442` -> `mergeable: MERGEABLE`, base `master`, 1 commit.

PR #449, opened 2026-08-14
build(release): support durable final-gate evidence roots

tsotchke:
## Summary

Adds an opt-in `ESHKOL_DURABLE_WORK_ROOT` mode for the canonical v1.3.4 final gates.

When enabled, release probes claim deterministic, gate-specific directories beneath a caller-provided absolute root; stale targets and symlink roots are rejected, evidence is retained, and no `/tmp`, `/private/tmp`, or `mktemp` path is used. Existing developer and CI behavior remains unchanged when the variable is unset.

This is temporarily stacked on #448 so CI can run in parallel. It will be retargeted to `master` after #448 lands.

## Coverage

- ICC smoke, VM parity, readiness, and Taylor equivalence
- AD/tensor/fuzz transitive gates
- memory RSS, JIT-cache, and static-link CTest gates
- existing release workflow surface test expanded; no CTest-count change

## Local verification

- release workflow surface test: PASS
- durable JIT-cache gate: PASS
- VM parity audit-only durable evidence: PASS
- bash syntax for every touched shell file: PASS
- ICC guard/pre-commit/reindex: PASS/FRESH
- author and committer identity: `tsotchkele@gmail.com`

### DIFF
--- a/scripts/lib/durable_work_root.sh
+++ b/scripts/lib/durable_work_root.sh
@@ -1,0 +1,64 @@
+#!/usr/bin/env bash
+# Optional deterministic evidence roots for release gates.
+#
+# A caller that sets ESHKOL_DURABLE_WORK_ROOT opts out of ephemeral scratch
+# space.  Each gate must claim a unique direct child through
+# eshkol_durable_prepare_dir; claiming an existing child is an error so a
+# release run cannot accidentally consume stale evidence.  The caller owns the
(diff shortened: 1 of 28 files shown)


tsotchke:
Temporarily closing and immediately reopening after retargeting to master so the master-targeted CI matrix is emitted for the unchanged exact head.

PR #460, opened 2026-08-25
TEST (throwaway, will be closed): mixed docs+code proof for PR #455

tsotchke:
Throwaway proof PR for #455, based on #455's branch (base=ci/docs-only-contexts, temporarily includes a TEMP-TEST-ONLY trigger-widening commit that will be reverted before #455 is finalized). This PR touches `docs/README.md` (doc) and `lib/backend/vm_core.c` (comment-only). Expect: docs_only=false, the full matrix RUNS exactly as today. This is the safety-critical direction. Will be closed and branch deleted after verification.


tsotchke:
Proof successful: docs_only correctly computed false for a mixed docs+code change, and the full matrix was scheduled/started (confirmed jobs reaching in_progress before cancellation to save CI resources — the correctness signal, not full completion, was what mattered here). Closing throwaway PR.

PR #479, opened 2026-08-25
feat(assurance): close A10 packaging manifest, A11 diagnostic corpus, A12 leak detection

tsotchke:
## Summary

Closes three ADR-0010 / v1.4-connection-gate assurance gaps, each measured
NOT IMPLEMENTED, with real gates that self-test their own ability to fail.

- **A10 — packaging manifest.** `.icc/package-manifest.yaml` declares every
  binary, static library, stdlib artifact, and stdlib source file a release
  archive must contain, with `.github/workflows/release.yml`'s three
  "Package Release Asset" steps (macOS/Linux + Windows) as ground truth.
  `scripts/check_package_manifest.py` verifies a staged package directory
  against it — including a full PATH-SET mirror check of `lib/**/*.esk`
  against `share/eshkol/lib/**`, not a bare count — wired into every
  packaging step before the existing smoke test and before archiving.
- **A11 — diagnostic golden corpus.** `tests/diagnostics/` pairs 8 input
  programs with an `expected.json` (message substrings, an `esh_ref`
  traceability tag, and file:line:col span), seeded from ESH-0364
  (wrong-file attribution), ESH-0365 (import position), arity-fatal,
  type-ascription (SW-11), linearity no-cloning, the guard-catchable
  runtime type errors from PRs #451/#452, and a positive control.
  `scripts/check_diagnostic_corpus.py` compiles/runs each case against a
  real build and fails on drift in text OR span.
- **A12 — leak detection.** `linux-x64-asan-ubsan` now runs with
  `detect_leaks=1` against a checked-in, per-entry-reasoned
  `.icc/lsan-suppressions.txt` (was `detect_leaks=0` unconditionally).
  `ESHKOL_ARENA_POISON` now covers `lib/backend/vm_arena.h` — the VM arena
  primitives every non-evacuated VM path allocates through — on top of the
  existing `vm_region_evac.c` coverage from #461.
  `scripts/check_leak_detection_selftest.sh` proves the detector is live by
  leaking on purpose under the exact CI options.

## What each gate rejects

| Gate | Rejects |
|---|---|
| `check_package_manifest.py` | a staged release package missing (or empty) any manifest-declared binary/library/stdlib file/doc, or one `lib/**/*.esk` source module not mirrored into `share/eshkol/lib/**` |
| `check_diagnostic_corpus.py` | a diagnostic whose message drifts, whose span (file/line/col) drifts, a diagnosed program that still produces a binary, a runtime-guardable case that stops being catchable, or a positive-control program that stops compiling/running |
| `check_leak_detection_selftest.sh` | a suppression file grown broad enough to swallow a real leak, or `detect_leaks` silently reverting to off |

## Self-test evidence (all `--self-test` fixtures pass; details below are from the actual runs)

- `check_package_manifest.py --self-test`: 5/5 fixtures correct (well-formed PASS; missing binary, missing mirror file, wrong-platform archive naming, malformed manifest all FAIL as expected).
- `check_diagnostic_corpus.py --self-test`: 11/11 fixtures correct (span-match, span-drift, "must not build and run", run/stdout contract, forbidden-text, and 4 `coverage_position_pin` fixtures for the ESH-0365 mechanism).
- `check_leak_detection_selftest.sh`: leaking probe caught (exit 1, LeakSanitizer report present), clean probe passes (exit 0, no report); fails closed when the suppression file is missing.
- `tests/core/vm_arena_poison_test.cpp` (`vm_arena_poison_test` / `vm_arena_poison_off_test` ctest entries): both pass, including the disarmed mode that proves the armed mode's assertions are not vacuous.

Real (non-self-test) corpus run against a full RelWithDebInfo build:
`diagnostic_corpus_clean: PASS (8/8)`.

**ESH-0365 regression proof (not just self-test fixtures):** I deliberately
reverted the fix in `lib/frontend/parser.cpp`
(`make_r7rs_import_ast(specs, ast.line, ast.column)` back to
`token.line, token.column`, i.e. the desugared import/require node reports
the closing paren's position again), rebuilt `eshkol-run`, and re-ran the
corpus:

```
[FAIL] esh_0365_import_closing_paren
  reason: dispatch/accept position mismatch for 'import': dispatch at
  (1, 2), accept at (2, 16) -- this is exactly the ESH-0365 shape
diagnostic_corpus_clean: FAIL (7/8)
```

All 7 other cases stayed green (the gate isolates the regressed construct,
not a blanket failure). Restored the fix, rebuilt, back to 8/8.

`check_package_manifest.py` was also proven against a real-shaped synthetic
package built from this worktree's actual manifest and actual 96-file
`lib/**/*.esk` tree: PASS when correctly staged, FAIL naming the exact
missing file when I deleted `lib/eshkol-agent-yoga.a`.

## Verify

- Baseline build (`-G Ninja -DCMAKE_BUILD_TYPE=RelWithDebInfo` + the
  standard flags): `ctest` 193/193 passed, 262.85s.
- `gate_no_silent_wrong.py`, `check_ledger_integrity.py`,
  `check_oracle_schema.py` (pre-existing gates): unaffected, still PASS.
- ICC `guard-diff --staged`: 0 violations on every commit in this branch.
- ICC `pre-commit-check --architecture-model`: my 3 new invariants
  (`INV-package-manifest-enforced`, `INV-diagnostic-corpus-no-drift`,
  `INV-leak-detection-armed`) are PASS/UNCHECKABLE, never FAIL. Its overall
  exit code also reflects two **pre-existing, unrelated** blocking
  invariants (`INV-engine-semantic-parity`, `INV-language-surface-exercise`)
  that require dedicated full differential-VM-parity and
  language-surface-coverage campaign runs — confirmed pre-existing on
  master before this branch, not introduced by this diff, and out of scope
  for A10/A11/A12.
- Identity: every commit is `tsotchke <tsotchkele@gmail.com>`, no
  Co-Authored-By trailers.

Not merging — leaving for review.


tsotchke:
Superseded in place by #486, which contains every commit on this branch plus the fix that makes the A12 leak lane able to fail. Holding this as a draft so the two do not land out of order; #486 delivers A10, A11 and A12 together.

tsotchke:
Superseded by #486 (merged) — A10/A11/A12 landed there.

PR #480, opened 2026-08-25
fix(runtime): make gensym reachable on every engine

tsotchke:
## Summary

`gensym` was implemented in `lib/core/introspection.cpp` but registered in no
dispatch table anywhere, so `(gensym)` failed on every engine. Fixed and
wired end to end, plus the coverage ratchet the fix requires.

### `gensym` reachable on all three engines

| Engine | Before | After |
|---|---|---|
| native JIT (`-r`) | compile error: `Unknown function: gensym` | works — distinct symbols, round-trips |
| native AOT | same compile error | works |
| bytecode VM | compiled as an undefined global (NIL); `(gensym)` -> fatal `calling non-function` | works |

All three failed **loudly** before (no silent-wrong case). Root cause went
one level deeper than dispatch wiring: `introspection.cpp` includes
`eshkol/llvm_backend.h` and is compiled only into `eshkol-static` (the
compiler/tool aggregate) — never into `libeshkol-runtime.a`, the slim archive
AOT/JIT user binaries actually link against. Codegen dispatch alone would
have failed to link (`undefined symbol: eshkol_gensym_ptr`). Fixed by
extracting gensym into a new `lib/core/runtime_gensym.cpp` (added to the
runtime-support source set in `CMakeLists.txt`) — the same split
`lib/core/symbol_intern.cpp` already uses for `eshkol_intern_symbol_lookup`,
for exactly the same reason.

Wired:
- `llvm_codegen.cpp`: `gensym` -> `codegenGensym`, calling `eshkol_gensym_ptr`
  and packing the result as a `HEAP_PTR` tagged value, mirroring
  `string->symbol`'s existing pattern.
- `eshkol_vm.c` `BUILTINS[]`: `{"gensym", 2227, 0}`; `vm_native.c` case 2227
  implements it with its own counter, distinct from `vm_macro.c`'s
  compile-time hygienic-renaming gensym (a different feature).
- `tests/coverage/language_surface.json` regenerated: 1040 -> 1041 builtins;
  `gensym` backends = `[vm, native_llvm]`.

Tests: `tests/control_flow/gensym_test.esk` (new) — callability, `symbol?`,
`eq?`-distinctness across successive calls, `symbol->string`/`string->symbol`
round-tripping. Verified 9/9 on native JIT, native AOT, and the bytecode VM.
Un-skipped the stale "gensym is VM-only" note in
`tests/v1_2_edge_cases/symbol_consistency_test.esk` — that claim was wrong
too (nothing called it successfully on the VM either).

### ESH-0080 stale note removed

`docs/reference/language/continuations.md` carried a "Known limitation —
deep CPS chains (ESH-0080)" section describing a SIGILL crash past roughly
recursion depth 16 as an ongoing limitation. That was fixed on 2026-06-29 by
commit `47338adb` ("run native O0 cleanup for deep CPS"): native `-O0`
codegen was skipping LLVM's function cleanup passes, leaving oversized stack
frames for deep closure chains. Re-verified this session:
`tests/sicp/ch4_amb_deep_cps_test.esk` passes on both native JIT and AOT. The
section is retitled "— FIXED" with the fixing commit cited. This is the only
change to that file in this PR; the rest of the page (including its
"Re-invocable (multi-shot) continuations" section) is untouched here and
is being tracked separately.

### Coverage ratchet

`gensym` grew the live language surface from 1107 to 1108 constructs.
`tests/coverage/coverage_policy.json`'s count floors and
`tests/coverage/execution_deficit.json`'s baseline are bumped to match, so
`scripts/test_language_coverage_gate.py`'s build-free guard
(`test_policy_count_floors_equal_the_live_surface_total`,
`test_ledger_baseline_equals_the_live_surface_total`) stays green instead of
reporting the manifest trailing the ledger. `gensym` is execution-backed
covered, not just spelled: `tests/control_flow/gensym_test.esk` is picked up
by `scripts/run_control_flow_tests.sh`'s directory glob, part of the corpus
`scripts/language_coverage.py` measures.

## Verification

Build: `-G Ninja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DLLVM_CONFIG_EXECUTABLE=/opt/homebrew/opt/llvm@21/bin/llvm-config -DBUILD_TESTING=ON -DESHKOL_BUILD_TESTS=ON -DESHKOL_BUILD_AGENT_FFI=ON -DESHKOL_FIXED_POINT_ENGINE=ON -DESHKOL_GPU_ENABLED=ON -DESHKOL_QUANTUM_ENABLED=OFF`, rebased onto current master.

| Gate | Result |
|---|---|
| `ctest -j8` | 198/198 passed |
| `tests/control_flow/gensym_test.esk` | 9/9 on native JIT, native AOT, bytecode VM |
| `tests/sicp/ch4_amb_deep_cps_test.esk` | passes on native JIT and AOT |
| `scripts/test_language_coverage_gate.py` | 17/17 passed |
| `scripts/check_ledger_integrity.py` | PASS, 98 entries |

**Identity audit**: every commit — author and committer
`tsotchke <tsotchkele@gmail.com>`. No `teumbach@gmail.com`, no
`Co-Authored-By`, no AI/assistant attribution, no emoji.

## Test plan

- [x] Verified `(gensym)` failed loudly on all three engines before the fix
- [x] Verified `(gensym)` works correctly on all three engines after the fix
- [x] Verified `tests/sicp/ch4_amb_deep_cps_test.esk` still passes (ESH-0080)
- [x] `ctest` 198/198, coverage gate 17/17, ledger integrity clean
- [ ] Maintainer review

---

**Merge-order note (2026-08-26):** PR #491 (multishot-continuations) independently
carries the same gensym native-codegen/VM-dispatch wiring this branch does (it was
developed in parallel, before either merged current master). Both branches currently
ratchet `tests/coverage/coverage_policy.json` / `execution_deficit.json` to 1108 —
consistent, since it is the same single construct (`gensym`) counted once against
master's baseline of 1107, not two independent additions. Whichever of #480/#491
merges second WILL hit a real content conflict on `lib/backend/eshkol_compiler.c`,
`lib/backend/eshkol_vm.c`, `lib/core/runtime_gensym.cpp` and `CMakeLists.txt` (both
add the identical dispatch entries) — that conflict resolves by keeping either side's
copy (they're the same change) and dropping the duplicate, with no further coverage-
baseline change needed since the surface total is already correct at 1108.


tsotchke:
Superseded: the gensym reachability fix landed on master inside #491 (f02ad8d6). After rebase this branch is byte-identical to master — nothing left to merge.

PR #500, opened 2026-08-26
fix(dispatch): a `default:` may not stand in for an invariant — closed-enum dispatch is compiler-enforced

tsotchke:
## The class, not the four sites

PR #498 found four tensor-valued AD node types falling into an early `default:` in `eshkol_tensor_backward_dispatch` that did **nothing at all**: gradients came back exactly `0.0`, no diagnostic, exit 0. That default's comment asserted the case was impossible, reasoning from a **numeric band of enum values** — *"every TENSOR op (19-32 and the qLLM bridge ops 67-80) has an explicit case above, so a tensor-op gradient can never silently fall through this default"* — a proxy that four newer node types had already stepped outside of. The comment was the only thing enforcing the claim, and a comment enforces nothing.

The shape is general: **a dispatch switch over a closed enum, carrying a `default:` that encodes an assumption instead of failing, in a language where the enum can grow without the switch ever knowing.** ESH-0214d is the same defect one subsystem over — heap subtypes carrying interior pointers fell to a shallow-leaf `default:` in the region evacuator and left those pointers aimed into a reclaimed arena.

### The same defect is still live on master, in four more node types

Not inferred from reading the source — **measured against master's own library (`afbaaf5b`)**:

```
master (afbaaf5b): reverse pass through qLLM geometric tensor nodes
  AD_NODE_HYPERBOLIC_DISTANCE  returned normally; input grad = STILL NULL (nothing propagated)
  AD_NODE_POINCARE_EXP_MAP     returned normally; input grad = STILL NULL (nothing propagated)
  AD_NODE_POINCARE_LOG_MAP     returned normally; input grad = STILL NULL (nothing propagated)
  AD_NODE_GEODESIC_ATTENTION   returned normally; input grad = STILL NULL (nothing propagated)
exit 0
```

`lib/bridge/qllm_bridge.cpp` records all four as **tensor** nodes (`make_node(tape, AD_NODE_..., out, shape, ...)` with a live `tensor_value`), and the reverse sweep `autodiff_codegen.cpp` emits routes *any* node with a non-null `tensor_gradient` to this dispatcher. `33-40` is exactly the band the comment did not cover. Every gradient through a Poincaré exp/log map, a hyperbolic distance or a geodesic attention returned zero, silently.

They survived review because the adjoint *does* exist — codegen emits scalar backward blocks for node types 33-40. It just is not on the sweep the bridge's tensor nodes actually take.

---

## Inventory

180 switches with three or more enum-member case labels were classified. **They are not all this defect.** The discriminator is what the default *does*:

| default arm | count | verdict |
|---|---:|---|
| **produces** — answers with a value | 120 | candidate: a member nobody considered gets a plausible answer |
| **refuses** — `eshkol_error` / `return false` / `*ok = 0` | 20 | correct as written: a partial classification stating its own boundary |
| **no default** | 40 | already exhaustive |

Of the 120 "produces", by enum:

| enum | sites | disposition here |
|---|---:|---|
| `eshkol_op_t` (113 members) | 24 | **deferred** — codegen dispatch; large mechanical diff |
| `eshkol_type_t` (22) | 21 | **deferred** |
| `ValType` / `OpCode` / `HeapType` (VM) | 31 | **loud backstop, correct** — dispatches on bytes read from bytecode |
| `hott_type_kind_t`, `pattern_type_t`, `TokenType`, … | ~42 | **deferred** |
| `eshkol_value_type_t` (24) | 10 | **deferred** — tag byte carries flag bits; partly open |
| `heap_subtype_t` | 2 | 1 **converted**, 1 **deferred** (`sexp_to_ast.cpp`) |
| `ad_node_type_t`, `callable_subtype_t`, `EvacKind` | — | **converted in full** |

**Converted (compiler-enforced, no `default:`):**

| site | enum |
|---|---|
| `lib/backend/tensor_backward.cpp` · `eshkol_tensor_backward_dispatch` | `ad_node_type_t` (83) |
| `lib/backend/tensor_backward.cpp` · `eshkol_ad_node_type_name`, `…_is_tensor` | `ad_node_type_t` |
| `lib/bridge/tensor_backward.cpp` · `get_tensor_backward_fn` | generated table, switch removed |
| `lib/core/runtime_regions.cpp` · `evac_kind_for` | `heap_subtype_t` (25) + `callable_subtype_t` (5) |
| `lib/core/runtime_regions.cpp` · interior-walk worklist | `EvacKind` (15) |
| `lib/core/runtime_regions.cpp` · ESH-0214d debug watchlist | `heap_subtype_t` |
| `lib/core/runtime_errors_hosted.cpp` · `eshkol_format_value_type_tag` | `heap_subtype_t`, `callable_subtype_t` |
| `lib/core/runtime_display_hosted.cpp` · `display` | `heap_subtype_t`, `callable_subtype_t` |
| `lib/core/introspection.cpp` · `eshkol_type_of` | `heap_subtype_t`, `callable_subtype_t` |
| `inc/eshkol/eshkol.h` · `eshkol_{heap,callable}_subtype_is_declared` | both |

**Left as loud backstops, deliberately:** the VM's `OpCode` / `ValType` / `HeapType` switches. A value that arrives from *outside* the program is an open set whatever the enum declares; demanding exhaustiveness there enforces a fiction. `lib/backend/vm_region_evac.c` already carries the registry pattern (a coverage table with a startup assertion, `INV-vm-region-evac-subtype-total`).

**Deferred as build items, mechanism proven:** `lib/core/sexp_to_ast.cpp`'s heap-subtype switch, whose `default: break` falls through to a null AST rather than refusing — same class, different subsystem; and the ~20 `eshkol_op_t` switches, whose conversion is a large mechanical diff that does not belong in the same change as the fix. Both recorded in SW-66 `notes`.

---

## The fix

**1 · Registry.** `inc/eshkol/ad_node_registry.def` is now the single declaration point for an AD node type: `ESHKOL_AD_NODE(NAME, VALUE, PAYLOAD, TENSOR_BACKWARD, BRIDGE_FN)`. The `ad_node_type_t` enum, the dispatcher's homogeneous arms, and the bridge backward table are all **generated** from these rows. A node type cannot exist without stating its backward disposition. A row claiming `BRIDGE` names a function; if that symbol does not exist, `lib/bridge/tensor_backward.cpp` does not compile — that is what makes *registered* a fact rather than an intention.

Values are unchanged (they are an ABI: emitted IR compares `node->type` against integer literals) and a static assertion proves the set stays dense, which the node-type-indexed table relies on.

**2 · Compiler enforcement.** No `default:`, plus `-Werror=switch -Werror=switch-enum` — per translation unit via `eshkol_require_exhaustive_dispatch()`, or per switch via `ESHKOL_EXHAUSTIVE_SWITCH_BEGIN/_END` (`inc/eshkol/exhaustive_dispatch.h`) where a file legitimately mixes closed dispatch with open. Both flags, because they are not redundant: `-Wswitch` is what fires for an unhandled member in a default-less switch; `-Wswitch-enum` is what *still* fires once someone re-adds a `default:` — the exact move that reopens this class, and precisely where `-Wswitch` goes quiet.

**3 · Open sets stay loud.** `eshkol_heap_subtype_is_declared` / `eshkol_callable_subtype_is_declared` split the open half off — a subtype byte read out of an object header is untrusted input — and each caller answers it with a value-naming fallback. That lets the closed half be exhaustive without pretending a corrupt byte is impossible. The single `default:` those switches used to carry answered *"corrupt byte"* and *"enum member added last week"* with the same shrug.

**4 · Refusal, not invention.** The eight geometric node types are declared `UNREGISTERED` — an explicit, written-down refusal — and the dispatcher aborts naming the node type. Writing exact Riemannian adjoints is AD work with its own correctness bar and belongs to the lane doing it; inventing one to close a dispatch hole is how a wrong gradient ships. **The four qLLM ops go from silently zero to loudly absent.**

**5 · ICC invariant.** `INV-closed-enum-dispatch-exhaustive` + `scripts/gate_exhaustive_dispatch.py` re-derive each enum's members **from its own definition** (not a hand-typed list, which would drift in exactly the situation being guarded) and fail on any registered site that carries a default, omits a member, or is not armed. It is the half the compiler cannot provide on MSVC — and the half that catches the enforcement being *removed*. Plus `INV-ad-node-declared-in-registry` as key-space equality between registry `BRIDGE` rows and the backward functions actually defined, catching the other direction: a backward rule that exists, works, and is wired to nothing (which is how seven bridge ops sat with correct backwards registered while the dispatcher dropped their gradients).

---

## Red proofs (measured, not asserted)

**Compile-time — new member, no arm:**
```
error: enumeration value 'AD_NODE_REDPROOF_NEW_OP' not handled in switch [-Werror,-Wswitch]
make[3]: *** [CMakeFiles/…/tensor_backward.cpp.o] Error 1
```

**Compile-time — new member with a `default:` restored (the #498 shape exactly):**
```
error: enumeration value 'AD_NODE_REDPROOF_NEW_OP' not explicitly handled in switch [-Werror,-Wswitch-enum]
```
The default cannot defeat the flag. This is why both diagnostics are escalated.

**A `default:` re-added alone** compiles (it is then dead code) and is caught by the gate instead:
```
FAIL  lib/backend/tensor_backward.cpp:eshkol_tensor_backward_dispatch  [ad_node_type_t, 84 members]
        - carries a `default:` — a default over a closed enum answers 'member I have not
          thought about' with a plausible value
```

**Gate self-test:** 5 fixtures, including one whose `default:` is annotated with an it-cannot-happen comment — the shape that fooled review — proving the gate is not fooled by the annotation.

All three were run and then reverted; the tree is clean.

---

## Gates

| gate | result |
|---|---|
| `ctest` (full) | **204/204 passed**, including the 3 new tests |
| `exhaustive_dispatch` (new) | 11/11 — registry density, name/payload totality, table↔disposition agreement, and 4 fork death tests proving the qLLM ops refuse |
| `exhaustive_dispatch_gate` + `_selftest` (new) | PASS |
| arena-poison suite (`ad_tape_region_growth`, `parallel_map_scope_reclaim`, `runtime_regions`, `parameter_region_barrier`) | PASS under `ESHKOL_ARENA_POISON=1` |
| `vm_parity_audit.py` | OK — every codegen symbol VM-supported or waived |
| `gate_ad_shared_node_model.py` | `ad_carrier_model_clean: PASS` |
| `check_ledger_integrity.py` | PASS |
| `gate_no_silent_wrong.py` | PASS |
| `check_oracle_schema.py` | PASS |

Ledger: **SW-66**, bucket SILENT-WRONG, status `guarded` — the wrong *value* is gone, but four of the eight geometric node types still cannot compute the right one, so the entry stays open until the Riemannian adjoints land.

---

## Coordination

- `.worktrees/evac-subtypes` (DNC/SDNC/TAYLOR deep-walk) currently touches only `.icc/architecture-model.yaml`. This PR's evac change is **structural only** — `default:` removal and explicit case labels; DNC/SDNC/TAYLOR keep their present `EVAC_LEAF` classification and their ESH-0214d watchlist entry. If that lane lands first, the conflict is three case labels moving from the leaf list to deep-walk kinds, and the exhaustiveness structure makes that change *easier* to verify, not harder.
- `.worktrees/abi-v2-stage1` is untouched.
- The qLLM geometric-backwards lane: this PR does not implement any adjoint. When those land, the four rows move from `UNREGISTERED` to `BRIDGE` with a named function, and the generated arms follow automatically.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -445,6 +445,7 @@ jobs:
           python3 scripts/check_required_context_consistency.py --self-test
           python3 scripts/check_package_manifest.py --self-test
           python3 scripts/check_diagnostic_corpus.py --self-test
+          python3 scripts/gate_exhaustive_dispatch.py --self-test
 
       # B5/N2 (2026-08-26 audit): every gate below this point deliberately
       # DROPS --no-trace. This job builds nothing (pure Python over
@@ -487,6 +488,21 @@ jobs:
       - name: Every AD op routes through a declared, exact, source-verified carrier
         shell: bash
         run: python3 scripts/gate_ad_shared_node_model.py
+
+      # Source-derived, and it has to be: the PRIMARY enforcement for this
+      # class is the compiler (-Werror=switch -Werror=switch-enum, or
+      # ESHKOL_EXHAUSTIVE_SWITCH_BEGIN per switch), which cannot report
+      # anything here — a build that fails produces an absence, not evidence.
+      # This step is the half that can: it re-derives each closed enum's
+      # members from its own DEFINITION and fails on any registered dispatch
+      # site that carries a `default:`, omits a member, or has had its arming
+      # removed. It runs in THIS job rather than only in ctest so it grades
+      # every PR shape, including docs-only ones where no build runs at all —
+      # a change that deletes eshkol_require_exhaustive_dispatch() from
+      # CMakeLists.txt touches no source file the build matrix would notice.
+      - name: No dispatch switch over a closed enum carries a default (SW-70)
+        shell: bash
+        run: python3 scripts/gate_exhaustive_dispatch.py --no-trace
 
       # --offline deliberately: reading branch-protection settings needs
       # administration:read, which the default workflow token does not
--- a/.icc/abi-header-baseline.json
+++ b/.icc/abi-header-baseline.json
@@ -1,5 +1,5 @@
 {
-  "commit": "892622c0f628f61bc5c829fb63f1cac4606c8553",
+  "commit": "61935eef779b3f1764c36f72208c2bfccea23810",
   "counts": {
     "A_macro_api": {
       "inc/eshkol/eshkol.h": 32,
@@ -161,14 +161,14 @@
       "lib/core/dnc_api.c": 3,
       "lib/core/i128_runtime.cpp": 3,
       "lib/core/inference.cpp": 6,
-      "lib/core/introspection.cpp": 8,
+      "lib/core/introspection.cpp": 10,
       "lib/core/kb_persistence.cpp": 7,
       "lib/core/logic.cpp": 10,
       "lib/core/runtime_autodiff.cpp": 4,
       "lib/core/runtime_closure_alloc.cpp": 4,
       "lib/core/runtime_continuations.cpp": 1,
       "lib/core/runtime_deep_equal.cpp": 7,
-      "lib/core/runtime_display_hosted.cpp": 6,
+      "lib/core/runtime_display_hosted.cpp": 8,
       "lib/core/runtime_exceptions_hosted.cpp": 4,
       "lib/core/runtime_hash_table.cpp": 4,
       "lib/core/runtime_object_alloc.cpp": 20,
@@ -278,7 +278,7 @@
     "D_ir_const_offset": 45,
     "E_alloc_with_header": 182,
     "F_parallel_header_decl": 2,
-    "G_header_field_write": 164,
+    "G_header_field_write": 168,
     "H_wasm_glue": 34,
     "J_secondary_prefix_abi": 33,
     "K_raw_byte_offset": 10,
--- a/.icc/architecture-model.yaml
+++ b/.icc/architecture-model.yaml
@@ -815,3 +815,91 @@ invariants:
     evidence:
       trace_name_pattern: '^abi_layout_pin$'
       max_age_days: 45
+
+  # 14. Closed-enum dispatch exhaustiveness, encoded as intended-invariant with
+  #    a source-derived guard.  THE CLASS: a dispatch switch over a closed enum
+  #    whose `default:` encodes an ASSUMPTION instead of failing, in a language
+  #    where the enum can grow without the switch ever knowing.  PR #498 found
+  #    four tensor-valued AD node types falling into an early `default:` in
+  #    eshkol_tensor_backward_dispatch that did nothing at all — gradients
+  #    exactly 0.0, no diagnostic, exit 0 — behind a comment asserting the case
+  #    was impossible, reasoned from a numeric band of enum values that those
+  #    four node types had already left.  ESH-0214d is the same shape one
+  #    subsystem over.  The primary enforcement is COMPILE-TIME
+  #    (-Werror=switch-enum per translation unit, or
+  #    ESHKOL_EXHAUSTIVE_SWITCH_BEGIN/_END per switch); this invariant is the
+  #    source-derived second half, so the property is re-measured on lanes
+  #    whose toolchain has no such diagnostic (MSVC) and so REMOVING the
+  #    enforcement is itself caught.  Expected PASS; adding a `default:` to any
+  #    registered site, or an enum member no site names, FAILs it.
+  - id: INV-closed-enum-dispatch-exhaustive
+    kind: intended-invariant
+    incident: closed-enum-dispatch-default
+    severity: critical
+    description: >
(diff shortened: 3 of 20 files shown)


tsotchke:
Follow-up on the same push: registered one more dispatch site with the gate — the ESH-0214d watchlist switch in `evac_object`. It lives inside `#ifndef NDEBUG`, so the compiler never sees it in the release builds CI actually runs; `-Werror=switch-enum` on that translation unit cannot protect a switch the preprocessor removed first. That site is guarded by the source-derived gate alone, which is the concrete reason the two halves are not redundant. Red-proofed by dropping one subtype from the watchlist: `FAIL lib/core/runtime_regions.cpp:evac_object`. Gate now covers 8 sites.

PR #501, opened 2026-08-27
fix(gpu): close the vacuous GPU correctness gate (SW-67)

tsotchke:
## What was broken

The GPU correctness gate could not fail.

- `tests/gpu/gpu_correctness_gate.esk` printed only unlabelled `RESULT <label> <value>` lines and exited 0 unconditionally — no `PASS`, `FAIL`, `error`, or `mismatch`-shaped output in any form, and no way for a judge (marker-based or exit-code-based) to ever fail it.
- `tests/gpu/gpu_test.esk`, `sf64_simple_test.esk`, and `sf64_nonuniform_test.esk` had the same shape: they computed results and displayed `(expected N)` comments but never actually compared actual against expected. An audit of every `tests/gpu/*.esk` file's failure vocabulary found these four files (not just the one originally named) emitted zero failure-shaped output.
- The other 13 files (plus the already-compliant `ozaki_certification_test.esk`) did print well-formed `FAIL: <description>` markers on a failed check, but **none of them ever called `exit` with a nonzero status**. A failed check only ever produced text, never a process-level signal — so any judge that (correctly) treats exit code as authoritative saw green regardless of what a GPU kernel actually computed.

`scripts/lib/test_isolation.sh` already documents a prior near-miss of this exact defect class (an anchored `^FAIL:` regex that missed indented markers and `sf64_primitives_test.esk`'s colonless `FAIL (got ...`), fixed by making marker matching unanchored. That fix closed the marker-*quality* half of the problem; it could not close the exit-code half, which is what this PR does.

## The fix

**1. Structured contract, exit codes primary.** Every `tests/gpu/*.esk` file now aggregates a failure counter and calls `(exit 0)`/`(exit 1)` on a final `PASS:`/`FAIL:` verdict line, matching the strongest pre-existing convention in the suite (`ozaki_certification_test.esk`). `gpu_correctness_gate.esk` additionally gained ten backend-independent self-check invariants (finiteness of every RESULT value, exact transpose round-trip, an independent manual reduce-sum cross-check that never touches `tensor-sum`'s own reduction path, and algebraic identities like `sum(A+B) == sum(A) + sum(B)`), so it has real teeth even when run standalone through `scripts/run_gpu_tests.sh` — not just through the CPU-vs-GPU differential in `gpu_correctness_gate.sh`.

**2. Marker alignment as backstop.** `scripts/run_gpu_tests.sh` now documents ONE verdict grammar at its head (`PASS: <description>` / `FAIL: <description>`), and `scripts/lib/test_isolation.sh` gains an additive, opt-in `eshkol_test_output_has_verdict()` / `eshkol_test_output_has_success_marker()` pair — no change to any existing caller's behavior — wired into a new "NO VERDICT" grading branch: a test that exits 0 but never printed a single recognized PASS/FAIL line is graded **FAIL**, not PASS. Absence of evidence that a check ran is not evidence the check passed (the same principle #472's oracle fix used).

**3. The canary.** `tests/gpu/gate_canary_must_fail.esk` is a deliberate, permanent failure. `run_gpu_tests.sh` runs it on every invocation, outside the normal PASS-required loop, and requires it to fail (non-zero exit AND a `FAIL:` marker). If it ever passes, the whole run is forced red regardless of every other result — a real regression could be hiding behind exactly the same kind of silence that let the canary through. `gpu_correctness_gate.sh` separately gained a `--self-test` mode (mirroring `check_ledger_integrity.py`'s convention) that asserts its `SKIP` path writes **no** trace record to `scripts/icc_traces/gpu_execution.jsonl`, while its `FAIL`/`PASS` paths write distinguishable ones — closing the "a skipped lane must never read as a pass" gap at the trace-emission layer, on top of `.icc/completion-oracles.yaml`'s existing `severity: high` absence-is-FAIL handling for the `gpu-execution` target.

**4. Stale-comment / tolerance correction.** A comment in both `gpu_correctness_gate.esk` and `.sh` claimed `display` prints only ~6 significant digits, used to justify a loose `1e-4` tolerance. That claim is false: `display` calls `eshkol_dtoa_shortest` (`lib/core/runtime_display_hosted.cpp`), the same shortest-round-trip formatter used everywhere else — confirmed bit-exact by a direct round-trip probe (print a double, parse it back, compare bits: exact match on 8 varied test values including subnormal-adjacent and full-mantissa-precision cases). `GPU_GATE_TOL` is tightened from `1e-4` to `1e-9` on the strength of that correction plus a real measured Metal-vs-CPU divergence of **exactly 0** across all ten probes on this workload (see evidence below) — five to nine orders of magnitude looser than the corruption a real GPU-kernel bug produces (per `ozaki_correctness_test.esk`'s documented CRT-defect class).

**5. Ledger.** Filed and closed `SW-67` (bucket `VACUOUS-ASSURANCE`) in `.icc/silent-wrong-ledger.yaml`. Verified against every open PR branch on the remote before filing — `SW-65`/`SW-66` are taken by `fix/geometric-bridge-backwards` (#498) and `fix/evac-subtypes-dnc-sdnc-taylor` (#499). `check_ledger_integrity.py` passes (108 entries, no duplicate IDs).

## Local proof (Metal, Apple M2 Ultra)

Canary red-proof, then the full suite, run via `./scripts/run_gpu_tests.sh`:

```
Canary  gate_canary_must_fail.esk (must fail)              RED as expected (exit 1)
  >>> harness confirmed live: a failing kernel does turn this gate red
Canary  gpu_correctness_gate.sh --self-test                PASS

Testing all files in tests/gpu/ directory...

Testing cuda_host_sync_regression_test.esk                 PASS
Testing cuda_ozaki_correctness_test.esk                    PASS
Testing elementwise_correctness_test.esk                   PASS
Testing gpu_correctness_gate.esk                           PASS
Testing gpu_diagnostic_test.esk                            PASS
Testing gpu_large_matmul_test.esk                          PASS
Testing gpu_scale_correctness_test.esk                     PASS
Testing gpu_test.esk                                       PASS
Testing matmul_correctness_test.esk                        PASS
Testing ozaki_certification_test.esk                       PASS
    PASS: JIT and AOT both prove Accelerate mismatch>0 in BLAS mode (JIT:23 AOT:23) and fixed-N16 Metal exact mismatch=0
Testing ozaki_correctness_test.esk                         PASS
Testing ozaki_fast_test.esk                                PASS
Testing reduce_correctness_test.esk                        PASS
Testing sf64_nonuniform_test.esk                           PASS
Testing sf64_primitives_test.esk                           PASS
Testing sf64_simple_test.esk                                PASS
Testing softmax_normalize_test.esk                         PASS
Testing transpose_correctness_test.esk                     PASS

Total Tests: 18   Passed: 18   Failed: 0   Pass Rate: 100%
Ozaki exact-GEMM certification: VERIFIED (JIT+AOT, CPU-BLAS mismatch vs Metal exact mismatches=0)
Gate canary: confirmed RED, and gpu_correctness_gate.sh --self-test: PASS
```

Red-proof on both structurally-central files, corrupt-then-revert: injecting a wrong expected value into `sf64_primitives_test.esk` and `gpu_correctness_gate.esk` each produced a nonzero exit with a `FAIL:` line; reverted before commit.

The full CPU-vs-GPU differential (`gpu_correctness_gate.sh`, fresh GPU-enabled + CPU-only builds, not `run_gpu_tests.sh`'s standalone pass) was also run end-to-end at the new `1e-9` tolerance:

```
Diffing GPU vs CPU RESULT lines (relative tolerance 1e-9)...
  ok       matmul-checksum: GPU=135255870.36117578 CPU=135255870.36117578 rel_diff=0
  ok       matmul-c00: GPU=642.7586144782134 CPU=642.7586144782134 rel_diff=0
  ok       matmul-cmid: GPU=219.0769892244969 CPU=219.0769892244969 rel_diff=0
  ok       matmul-clast: GPU=52.81408417220773 CPU=52.81408417220773 rel_diff=0
  ok       add-checksum: GPU=526312.6631015609 CPU=526312.6631015609 rel_diff=0
  ok       mul-checksum: GPU=260851.59076217725 CPU=260851.59076217725 rel_diff=0
  ok       transpose-roundtrip-c00: GPU=642.7586144782134 CPU=642.7586144782134 rel_diff=0
  ok       transpose-roundtrip-clast: GPU=52.81408417220773 CPU=52.81408417220773 rel_diff=0
  ok       reduce-sum-A: GPU=263348.73874376045 CPU=263348.73874376045 rel_diff=0
  ok       reduce-max-A: GPU=2.0000999998864146 CPU=2.0000999998864146 rel_diff=0

Max relative diff across all probes: 0 (tolerance 1e-9)
PASS: GPU execution matches CPU reference within tolerance (1e-9) on Darwin
```

## Mesh / CUDA lane

`.github/workflows/ci-mesh.yml`'s advisory lane should pick this branch up on the self-hosted CUDA runner. Not yet confirmed at the time of opening this PR — will report the run result as a follow-up comment.

## Scope

The GPU harness/test contract only (`tests/gpu/`, `scripts/run_gpu_tests.sh`, `scripts/lib/test_isolation.sh`, the ledger). No dispatch-switch code touched — that is a separate, parallel lane's surface.

This closes a blocker for the engine-kernel certification program: a GPU correctness gate that cannot fail is not evidence a certification consumer can build on.

## Test plan

- [x] `./scripts/run_gpu_tests.sh` — 18/18 PASS, canary RED, self-test PASS, exit 0
- [x] `tests/gpu/gpu_correctness_gate.sh` — full GPU-vs-CPU differential PASS at the new `1e-9` tolerance, measured `max_rel_diff=0`
- [x] `tests/gpu/gpu_correctness_gate.sh --self-test` — PASS
- [x] Red-proof: corrupted-then-reverted checks in `sf64_primitives_test.esk` and `gpu_correctness_gate.esk` both produced nonzero exit + `FAIL:` line
- [x] `python3 scripts/check_ledger_integrity.py` — PASS, no duplicate IDs
- [ ] Self-hosted CUDA mesh lane (`ci-mesh.yml`) — pending confirmation

### DIFF
--- a/.icc/silent-wrong-ledger.yaml
+++ b/.icc/silent-wrong-ledger.yaml
@@ -5943,3 +5943,102 @@ entries:
       consumer can build on. Scope was the GPU harness/test contract only;
       no dispatch-switch code was touched (a parallel lane owns that
       surface).
+
+  - id: SW-71
+    bucket: VACUOUS-ASSURANCE
+    status: closed
+    closed_at: "fix/gpu-correctness-gate-vacuous"
+    closed_by: "branch fix/gpu-correctness-gate-vacuous"
+    title: "Windows judge: ^-anchored FAIL: marker regexes matched only line 1 of captured output; markers after a banner line were invisible on every Windows suite; discovered when the GPU gate canary was judged NO FAIL MARKER"
+    repro: >-
+      On windows-x64-cuda, tests/gpu/gate_canary_must_fail.esk (SW-67's
+      deliberate, permanent-failure canary) exits nonzero and prints a
+      banner line first, then its marker on the SECOND line:
+        === GPU Gate Canary (this test MUST fail) ===
+        FAIL: gate_canary_must_fail (deliberate: expected C[0,0]=999.0, got 19) - this failure is EXPECTED
+      scripts/run_all_tests.ps1's Invoke-SimpleCompileRunSuite grades the
+      canary path with `$FailRegex -and -not ($run.Output -match
+      $FailRegex)` for suites configured with an anchored FailRegex such as
+      "^FAIL:|Failed:\s+[1-9]" (gpu, and the same ^-anchored shape on the
+      bignum/rational/parallel/signal/error_handling/macros/io/benchmark/
+      migration/codegen/numeric suites; Invoke-TimedCompileRunSuite's
+      ASSERTION FAIL check and Invoke-XlaSuite's inline `$run.Output -match
+      "^FAIL:"` share the same construction). PowerShell's bare -match
+      calls .NET regex without RegexOptions::Multiline, so a bare `^`
+      anchors to the START OF THE WHOLE STRING, not each line: because the
+      banner line comes first, `^FAIL:` never matches, and the judge
+      reports "CANARY FAILED WITH NO FAIL MARKER" even though the marker
+      is right there on line 2. windows-x64-cuda failed three consecutive
+      PR #501 runs on this false negative before the anchoring defect
+      itself was identified as the root cause.
+    observed: >-
+      The canary's own FAIL: marker, present verbatim in stdout, was
+      reported as absent - "gate_canary_must_fail.esk (must-fail canary
+      exited nonzero but printed no FAIL: marker)" - purely because it was
+      not the first line of the captured output. The defect is not scoped
+      to the canary: the identical `$run.Output -match $FailRegex`
+      evaluation is also the ASSERTION FAIL path for every ordinary test in
+      every Windows suite listed above, so a genuine `FAIL:` marker printed
+      after any banner, header or diagnostic line was silently invisible to
+      the judge in the OTHER direction too - a real regression whose
+      marker isn't on line 1 would have been graded PASS, not FAIL.
+      -RequirePassMarker's unanchored "PASS" check and the
+      RuntimeErrorRegex "error:" checks carry no `^`/`$` anchor and were
+      not affected by this half of the defect.
+    correct: >-
+      A `^`-anchored marker regex passed to a Windows suite runner means "a
+      line that starts with this marker" - the same semantics the bash
+      judge gets for free from grep's line-by-line matching
+      (scripts/run_gpu_tests.sh, scripts/lib/test_isolation.sh) - not "the
+      first line of the entire captured stream is this marker." Every site
+      that evaluates a FailRegex/RuntimeErrorRegex/RequirePassMarker or ad
+      hoc marker regex against a test process's captured output must apply
+      RegexOptions.Multiline (or equivalent line-splitting) so that `^`/`$`
+      bind per line, defined in exactly one place so the semantics cannot
+      drift back out of sync at a future call site.
+    observed_after: >-
+      Added `Test-OutputMatches -Output $Output -Regex $Regex` (wraps
+      [regex]::IsMatch with RegexOptions.Multiline) as the single choke
+      point for marker-regex evaluation in scripts/run_all_tests.ps1, and
+      routed every FailRegex/RuntimeErrorRegex/RequirePassMarker check in
+      Invoke-SimpleCompileRunSuite and Invoke-TimedCompileRunSuite, plus
+      Invoke-XlaSuite's inline `^FAIL:`/`PASS`/`error:` checks, through it
+      - six call sites total; none of the suites' own FailRegex/
+      RuntimeErrorRegex string literals changed. Verified on the real
+      Windows 10 / PowerShell 5.1 host (desktop-jack-blupc) with a
+      csc.exe-built stub eshkol-run.exe standing in for the compiler: (a)
+      the real canary fixture (banner line, then a FAIL: marker on line 2,
+      exit 1) -> PASS (must-fail canary correctly failed) with 0 failed;
+      (b) an ordinary fixture (a couple of lines, exit 0) -> PASS; (c) the
+      canary forced to exit 0 -> FAIL "did not fail - verdict pipeline is
+      broken"; (d) an ordinary fixture that exits 0 but prints a `FAIL:`
+      marker on line 2 -> now correctly graded FAIL, proving marker
+      detection is live again for normal tests, not just the canary; (e)
+      `[System.Management.Automation.Language.Parser]::ParseFile` reports
+      no syntax errors under PowerShell 5.1.19041.
+    fixed_by: >-
+      branch fix/gpu-correctness-gate-vacuous - Test-OutputMatches helper
+      in scripts/run_all_tests.ps1, applied at every FailRegex/
+      RuntimeErrorRegex/RequirePassMarker/inline-marker evaluation site.
+    evidence: >-
+      scripts/run_all_tests.ps1 (Test-OutputMatches,
+      Invoke-SimpleCompileRunSuite, Invoke-TimedCompileRunSuite,
+      Invoke-XlaSuite) - real-host run on desktop-jack-blupc (Windows 10,
+      PowerShell 5.1.19041) against a csc.exe-built stub compiler plus two
+      fixture executables, results (a)-(e) above; local `pwsh -NoProfile`
+      parse check; python3 scripts/check_ps1_encoding.py PASS (the file
+      stays pure ASCII).
+    caught_by: [windows-x64-cuda-ci, direct-measurement]
+    missed_by: [run_all_tests.ps1 (pre-fix), icc-smoke, icc-readiness]
+    notes: >-
+      Discovered while fixing SW-67's own must-fail canary contract on
+      Windows (commit 0b4ce1ce): the canary went red for the right reason
+      (the must-fail file did fail) but the judge's *reason* for choosing
+      the FAIL branch it took was already wrong, one layer down. A judge
+      that cannot see a marker that IS there is the same shape of defect
+      as a gate that cannot go red at all (SW-67's VACUOUS-ASSURANCE
+      finding) - one degree closer to the reporting layer instead of the
+      test-content layer. Scope is the Windows PowerShell judge only;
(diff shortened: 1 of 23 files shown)


tsotchke:
Mesh/CUDA lane follow-up: \`CI (mesh, self-hosted)\` ran on this PR (run 33027528505) and completed in 6s via \`mesh-preflight\`, which correctly declined to fan out — repo variable \`ESHKOL_MESH_CI\` is currently unset (not \`on\`), so no mesh lane (including the CUDA execution lane) actually ran. This is the workflow's documented off-by-default safety behavior, not a defect introduced by this PR. No CUDA-runner evidence is available for this change until that variable is turned on / a runner carrying the \`eshkol\` label is confirmed online; the Metal local proof in the PR body stands as the real-hardware evidence for now.

tsotchke:
## windows-x64-cuda fix: the canary contract was known to one judge, not the other

Root cause confirmed exactly as flagged: \`gate_canary_must_fail.esk\` exited 1 as designed, but the Windows CUDA lane judges \`tests/gpu/\` through \`scripts/run_all_tests.ps1 -Mode gpu\`, not \`run_gpu_tests.sh\` — and \`run_all_tests.ps1\` had never been taught the canary contract, so it counted the deliberate red as a genuine suite failure (\`Tests Failed: 1\`, \`Failed Suites: gpu\`, exit 1).

### Fix: a declarative convention, not duplicated knowledge

Added \`Test-IsMustFailCanary\` / \`Get-MustFailCanaryVerdict\` to \`run_all_tests.ps1\`, keyed off the filename pattern **\`*_must_fail.esk\`** (matches both \`gate_canary_must_fail.esk\` and the extension-stripped form some suite drivers use) rather than a hardcoded literal name — the same shape of knowledge \`run_gpu_tests.sh\`'s \`CANARY_NAME\` already carries, now expressed as a rule instead of a name. Wired into both generic compile-run drivers (\`Invoke-SimpleCompileRunSuite\`, used by the \`gpu\` suite; \`Invoke-TimedCompileRunSuite\`, for future-proofing) so any future \`*_must_fail.esk\` in any suite is handled automatically:
- exit 0 → hard FAIL, named "verdict pipeline broken"
- nonzero exit with no FAIL-shaped marker → hard FAIL, named "marker grammar broken"
- nonzero exit + FAIL marker → the PASS condition for that file

Also hardened the \`gpu\` suite's \`-FailRegex\` from the old \`^FAIL:\` (a **single-line-anchored .NET regex matched against the whole captured output string** — effectively dead past the first line) to the same unanchored, word-boundary pattern \`scripts/lib/test_isolation.sh\` already uses, and added \`-RequirePassMarker\` so a test exiting 0 with no PASS/FAIL marker at all is graded FAIL (parity with \`run_gpu_tests.sh\`'s NO VERDICT branch).

### Verified without a Windows machine

No \`pwsh\` was available locally, so I installed it via a prebuilt Homebrew bottle (no compilation) and: (1) confirmed the full script still parses via \`[System.Management.Automation.Language.Parser]::ParseFile\`; (2) extracted the function definitions and unit-tested \`Get-MustFailCanaryVerdict\` against synthetic \`$run\` objects — **including the exact windows-x64-cuda failure shape** (exit 0, canary's own "THIS IS A BUG" text) — confirming it is now graded \`Ok = $false\` (forces red), and the correct canary shape (exit 1 + FAIL marker) is graded \`Ok = $true\`. All 9 assertions passed.

### Audit: every other judge of \`tests/gpu/*.esk\`

| Judge | Status |
|---|---|
| \`scripts/run_gpu_tests.sh\` | Already fixed (this branch, prior commit) |
| \`scripts/run_all_tests.ps1\` (\`Mode 'gpu'\`/\`'all'\`) | **Fixed by this commit** — root cause of the windows-x64-cuda red |
| \`scripts/run_all_tests.sh\` (meta-orchestrator) | Delegates to \`run_gpu_tests.sh\`; its own suite verdict is exit-code-primary already, and its log-detail-scanner never sees the canary's text (distinct "Canary" line prefix, no stdout dump on the expected-red path) — documented inline so this can't regress silently |
| \`tests/gpu/*_gate.sh\` (gpu_correctness_gate, ozaki_correctness_gate, ozaki_fast_gate, ozaki_certification_gate, cuda_ozaki_correctness_gate) | Each targets one named \`.esk\` file directly, never globs the directory — cannot encounter the canary |
| CMakeLists.txt / ctest | No registration references \`tests/gpu/*.esk\` at all |
| \`scripts/run_v1_2_edge_cases*.sh\` | No reference to \`tests/gpu\` — not a judge |
| \`scripts/language_coverage.py\` | Globs \`tests/gpu/*.esk\` for surface feature-usage counting only; does not execute or grade |
| \`.github/workflows/{ci-mesh,gpu-execution-gate}.yml\` | Invoke \`run_gpu_tests.sh\` / \`gpu_correctness_gate.sh\` directly — already covered above |

### SKIP-vs-PASS on the GPU-less windows-x64-cuda runner

Confirmed genuine, not vacuous: **zero** \`[GPU]\` init/dispatch banner lines appear anywhere in the full windows-x64-cuda job log, meaning every one of the 17 non-canary tests executed via CPU fallback, and every "PASS" reflects a real backend-independent correctness check (finiteness, exact identities, tolerance comparisons) — none of these files claim to prove GPU execution happened. \`cuda_host_sync_regression_test.esk\` does print an informational \`SKIP: no active GPU backend; validating CPU fallback only\` line when \`eshkol_gpu_init()\` reports 0 devices, but it still runs the real \`A*I=A\` check and exits on the actual result — the skip is advisory text, not a shortcut around verification. The distinct, correctly-isolated concern of *proving GPU hardware actually executed* is owned by \`gpu_correctness_gate.sh\`'s SKIP+no-trace mechanism (feeding \`.icc/completion-oracles.yaml\`'s \`gpu-execution\` severity:high criterion) — untouched here, and not invoked by this Windows lane at all.

Rebased onto current master (through #499) and pushed.

PR #505, opened 2026-08-28
ci: infrastructure-disclosure gate for PR text and commits

tsotchke:
The public repo must never contain private network access details or
credential material (private IPs, SSH key paths, tunnel/proxy recipes, MAC
addresses). Squash-merged commit `97869acd` carried two internal hostnames
in its body -- the second such leak this week -- because GitHub assembles a
squash-merge commit body out of the PR's individual branch commit messages,
and nothing in review treats commit prose as a reviewable surface the way a
diff hunk is.

**Scope, per maintainer ruling:** internal machine/hostnames by themselves
are public-safe in this repo and are not what this gate flags. It targets
the things that are load-bearing regardless of whose name is on them: a
route into a private network, or a credential a reader could act on.

## What this adds

`scripts/check_disclosure.py` scans three text surfaces for a PR's
`<base>..<head>` range: every commit message, the PR title/body, and the
ADDED lines of the diff over tracked text files. Two independent detection
layers:

1. **Generic patterns** (hardcoded, safe to publish -- they describe a
   structural shape, never a specific identifier): private IPv4 ranges
   (10.x, 172.16-31.x, 192.168.x, and the 100.64-127.x CGNAT range
   Tailscale also allocates from), ssh key file references, `-i
   ~/.ssh/...`, `ProxyCommand`, `tailscale nc`, and MAC addresses.

2. **An optional private denylist** of exact tokens, loaded from a file
   OUTSIDE this repository (`$ESHKOL_DISCLOSURE_DENYLIST`, default
   `~/.eshkol/disclosure-denylist.txt`), case-insensitive substring match,
   for a maintainer who wants to additionally flag specific known
   identifiers. Off by default is a normal, supported mode: with no
   denylist file, the gate prints one neutral line noting the layer isn't
   configured and runs the generic layer only.

`scripts/disclosure_allow.txt` is a tracked, public-safe allowlist
(reproducible-benchmark hardware names, CI capability labels) -- a match
fully inside an allowlisted phrase is not a hit.

Every finding is reported with the matched span redacted, never the literal
token -- the CI log a failing PR produces is itself public, and even a
generic-pattern hit can be a live secret. Denylist hits are labeled only
`denylist`, never which token matched.

`--self-test` builds in-memory fixtures for every generic pattern class, the
allowlist-suppression mechanic, diff line-number tracking, and both
denylist states (active, via a fixture under this repo's `.scratch/`, never
`/tmp`; and not-configured, asserting the neutral-note guarantee).

## CI wiring

Wired into `.github/workflows/ci.yml`'s `assurance-gates` job: the checkout
step gains `fetch-depth: 0` so the gate can diff the PR's own base/head SHAs
directly. A step optionally materializes the `DISCLOSURE_DENYLIST`
repository secret to a runner-local file when configured (PR title/body are
passed through env vars and a file, never interpolated into the shell
command); with no secret set, it prints a neutral note and the gate runs its
generic layer only. The real scan runs only for `pull_request` events,
failing closed if the base/head commits can't be resolved. Push events emit
a deliberate no-op PASS. Registered as a `severity: high` criterion
(`disclosure_clean`) in `.icc/completion-oracles.yaml`'s
`eshkol-compiler-readiness` oracle, following the pattern PR #503's
`check_ps1_encoding.py` used.

## Optional: a maintainer's own denylist

The `DISCLOSURE_DENYLIST` repository secret is optional. A maintainer who
wants to additionally flag specific known-sensitive tokens can populate a
local file outside this repo (default `~/.eshkol/disclosure-denylist.txt`,
one token per line) and set:

```
gh secret set DISCLOSURE_DENYLIST < /path/to/local/denylist.txt
```

Without it, CI runs the generic layer only -- that is expected, ordinary
operation, not a degraded state.

## Local demonstration

Ran the generic layer alone (no denylist configured) against
`694c3179..origin/master`: **0 findings** -- no private IPv4 literals, ssh
key-file references, ProxyCommand/`tailscale nc` recipes, or MAC addresses
anywhere in that range.


tsotchke:

PR #508, opened 2026-08-28
docs: v1.3.5-evolve release documentation audit

tsotchke:
Full documentation audit against the code at `97869acd`, ahead of the v1.3.5-evolve tag. Two commits: mechanical staleness and gate documentation first, then status/roadmap/navigation reconciliation.

Doctrine applied throughout: where a doc **lagged** shipped reality it was raised; where a doc **claims** something the code does not deliver, the claim stands and the gap is filed as a build item for the maintainer's ruling (listed at the end).

## Machine-checked staleness

- Regenerated `docs/api/` with `scripts/gen_api_docs.py`. Six public headers added this wave had no generated page at all: `abi_fingerprint.h`, `memory_abi_v2.h`, `exhaustive_dispatch.h`, `frontend/node_identity.h`, `core/symbol_syntax.h`, `backend/frechet_mean_core.h`. `--check` is now clean.
- Corrected 100+ stale source line counts across README, API_REFERENCE, ESHKOL_V1_ARCHITECTURE, DESIGN, the `docs/breakdown/` set, `examples/README.md` and the press sheet, each re-measured with `wc -l` at this commit. Aggregate rows (XLA backend, VM family, headers, test corpus, whole-tree totals) recomputed from an explicitly stated file set.
- Re-hashed the five `doc-claims-allowlist` entries whose `claim_id` had drifted again. `claim_id` is a hash over path+line+value, and both CHANGELOG.md and RELEASE_NOTES.md grew above them; each is now carried at **both** its master-branch and its release-branch line, so the gate is green whichever branch is measured. `check_doc_claims_residual.py` at this commit: unexplained wrong claims **27 to 0**.

## Capability claims raised to what shipped

- **Multi-shot re-entrant `call/cc` (#491).** COMPLETE_LANGUAGE_SPECIFICATION 16.1.1 still said "Single-shot capture via `setjmp`/`longjmp`" thirteen lines above 16.1.2's own multi-shot statement. API_REFERENCE described the continuation as an escape procedure. The guide and quick reference said nothing.
- **Exact VM `divergence`/`curl` (#487).** README, FEATURE_MATRIX and ROADMAP all still described these as `h=1e-7` central differences. `vm_native.c` cases 753/754 seed one forward dual per input variable with no step size anywhere.
[...]

PR #509, opened 2026-08-28
docs: complete the public surface — geometry builtins, AD tape, core.memory, ABI/find_package/registry/gate contracts

tsotchke:
Closes BI-33 and BI-34 of the v1.3.5-evolve documentation audit: public surface
that ships with no user-facing page, and mechanisms that refuse things without
telling anyone why.

Text only. No build was run; every claim is grounded in source, tests or the
committed machine sources of truth. Built on top of #508 (merged into this
branch, so it should land after it).

## New pages

**`docs/reference/stdlib/geometry.md`** — all **62** geometric builtin names,
native ids 804-861. Name, id, arity, argument types, return type, the closed form
the shipped build computes, and a runnable call shape for each. Every
name/id/arity triple was verified mechanically against `eshkol_vm.c`'s
`BUILTINS[]` table.

Three facts that were nowhere on record:

- They are **VM-only**. Absent from the native LLVM dispatch and the native
  closure table, so a compiled program calling one gets an unknown-function
  error, not a slow path.
- **No target in this repository defines `ESHKOL_GEOMETRIC_ENABLED`**, so every
  shipped build takes `vm_geometric.c`'s portable constant-curvature path. In
  that path several ops are flat Euclidean approximations that discard the
  curvature argument they accept. Each entry says exactly what its op does.
  `frechet-mean` is the one genuinely Riemannian op, and the only one that
  raises rather than returning `()`.
- Six spellings collide with the pure-Scheme `core.manifold` module and are
  **not** the same functions — `make-hyperbolic-manifold` differs in arity,
  `manifold-type` returns an integer here and a symbol there.

The page carries no stamped example output and says so: these ops run only on
the bytecode VM and no output was captured for this revision. What it asserts
instead is each op's closed form, which is a source fact.

**`docs/reference/ad/tape.md`** — the explicit reverse-mode tape surface,
**33 names across 30 ids**, both engines. Lifecycle, the fact that `ad-backward`
zeroes before seeding (so it is re-runnable and does not accumulate), the
unchecked recorders vs the checked accessors, `ad-pow`'s node exponent, the three
engine differences, and the instrumentation counters — including why
`(ad-note-finite-difference!)` is what makes `(= (ad-finite-difference-evals) 0)`
a measurement rather than a tautology.

## Extended pages

| Page | What was added |
|---|---|
| `docs/STDLIB_V1_2_API.md` | `core.memory` (15 exports) and `core.memory_store` (12) as Appendix B.16/B.17 — event and log shapes, the durability contract, the two-layer verify, the v1 payload contract |
| `docs/reference/agent/ffi.md` | The `eshkol_object_abi_v*` guard: what an undefined reference to it means, its four causes, how to diagnose it, and the three cases it does not cover |
| `docs/BUILD_INTEGRATION.md` | The `find_package(Eshkol)` consumer contract — result variables, `REQUIRED_VARS`, the imported target, hint variables, the compile-function signatures |
| `docs/reference/ad/architecture.md` | The 83-row AD node registry, what `UNREGISTERED` means, why the dispatcher has no `default:`, the exact geometric backwards for node types 33-40, where they refuse and why, and the Fréchet stationarity gate |
| `docs/breakdown/GPU_ACCELERATION.md` | The gate's payload contract, the trace contract, the two distinct "can this go red?" mechanisms, and the deliberate skip-clean vs fail-closed asymmetry with the AD VM gate |
| `docs/breakdown/RUNTIME_CONFIGURATION.md` | How the VM region variables are actually parsed, and the Stage-1 evacuator's five practical limits |
| `docs/reference/language/continuations.md` | Pinning marks **every** open region, is never lifted, leaks differently per engine with only the VM saying so, and is one of **two** independent native mechanisms |

## `cmake --install` — a gap, marked as such

`CMakeLists.txt` has no `install()` rule for `FindEshkol.cmake`,
`EshkolCompile.cmake`, the `eshkol-runtime` archive, or `lib/stdlib.esk`. All
four are installed only by the homebrew formula and the release workflow.
`stdlib.o` and `stdlib.bc` are installed, so a plain `cmake --install` prefix
satisfies two of the four `REQUIRED_VARS` and fails the other two.

No PR adding those rules is open on origin, so this is documented as
**"Planned for a v1.3.5 follow-up"** rather than as complete. If that lane lands
first, this line is the one to flip.

## Corrections found along the way

- `core.manifold` was listed in `STDLIB_V1_2_API.md` as not auto-loaded (it is,
  via `lib/stdlib.esk`) and described as "placeholders pending native VM ops"
  (it is a complete closed-form implementation).
- `docs/reference/stdlib/memory.md` still carried a Known-issues entry for the
  `memory-fold-lww` type warning. That is ledger **SW-63**, closed by #494.
- `event-content-hash` is a public export of `core.memory` that neither the page
  nor the stdlib INDEX listed.
- `ESHKOL_VM_REGION_QUIET` was documented as silencing a `region-close` note; it
  silences the `with-region` promote-whole note.
- `GPU_ACCELERATION.md` had two sections numbered 9. Renumbered; nothing links
  by anchor.
- The stdlib INDEX said 58 modules / 638 provided symbols; the table itself has
  59 / 679.
- `vector-transport` (843) is a geometric op that the surface generator files
  under `vector` by name prefix, so the `geometry` category reports 61 while the
  family is 62. Recorded on the page; coverage is unaffected.

## Gates

| Gate | Result |
|---|---|
| `scripts/check_surface_counts.py` | PASS — 10 registered docs agree |
| `scripts/gen_api_docs.py --check` | `docs/api/ is up to date.` |
| `scripts/check_doc_claims_residual.py` | PASS, unexplained 0 (was 6) |
| `scripts/gate_no_silent_wrong.py` | PASS |
| `scripts/check_ledger_integrity.py` | PASS |
| `scripts/check_oracle_schema.py` | PASS |
| `scripts/check_ps1_encoding.py` | PASS |
| `audit_oracle_false_green`, `check_evidence_staleness`, `gate_ad_shared_node_model`, `gate_exhaustive_dispatch`, `check_build_fingerprint`, `check_self_verdicts` | PASS |
| all 15 `--self-test` fixtures | PASS |
| link and anchor check across every changed file | 0 dangling |

`check_package_manifest.py` needs a built package and `check_diagnostic_corpus.py`
needs `build/eshkol-run`; neither is runnable under the no-build rule.

The six doc-claims entries added to the allowlist are all the family-sum
misattribution the allowlist's own `detector_fix_references` section already
describes (an aggregate line count across a list of files, bound by the detector
to one of them). Each aggregate was re-measured and is correct as written; the
figures are #508's, in files this branch does not edit.


tsotchke:
Consolidated into #513

PR #515, opened 2026-08-28
fix: v1.3.5 audit follow-ups — vm manifold-dim (SW-72) and installable find_package contract

tsotchke:
Code follow-ups from the v1.3.5 docs audit (report section 3, items BI-4 and BI-29).

## BI-4 — `vm_geometric_manifold_dim` silent-wrong shape (ledger SW-72)

`vm_geometric_manifold_dim`'s `ESHKOL_GEOMETRIC_ENABLED` branch
(`lib/backend/vm_geometric.c`) returned `0` unconditionally, ignoring the
manifold's real dimension — a plausible-looking wrong answer, not a loud
error. `docs/KNOWN_ISSUES.md` recorded it as prose only with no ledger
row, so `scripts/gate_no_silent_wrong.py` reported PASS.

- Fixed by mirroring the neighboring, already-correct
  `vm_geometric_manifold_curvature`: guard on `vm_manifold_has_value` and
  return `qllm_manifold_get_dim` on the manifold's opaque handle. The
  portable fallback branch (`VmFallbackManifold.dim`) was already correct
  and is unchanged.
- Filed `.icc/silent-wrong-ledger.yaml` **SW-72** (closed, with
  repro/root-cause/fix/evidence) — next free id, checked against
  `origin/master` and every open PR branch (`git branch -r` + grep; SW-71
  was the prior highest).
- Added VM-parity coverage for two manifolds' dimensions on both engines:
  `tests/vm/geometric_fallback_numeric_regression.esk` (VM native id 858
  via `manifold-dim`, hyperbolic + a newly added Euclidean handle of a
  different dimension) and `tests/manifold/manifold_test.esk` (native/AOT
  `core.manifold`'s `manifold-dimension`, the module's existing
  Poincare-ball and Euclidean manifolds).
- Removed the now-fixed prose bullet from `docs/KNOWN_ISSUES.md` (checked
  PR #508's diff on that file first — it doesn't touch this line).

Note: `ESHKOL_GEOMETRIC_ENABLED` is not wired to any CMake option in this
tree — it's a different macro from `ESHKOL_QLLM_ENABLED`, which only
controls the standalone `eshkol-qllm-run` demo's `USE_QLLM` matmul
backend — so that branch remains uncompiled by any build this repo's
`CMakeLists.txt` can produce today. Recorded in the ledger notes; wiring
a real build config for it is out of scope here.

**Test evidence:**
- `scripts/run_manifold_tests.sh` (native `eshkol-run`): 19/19 passed,
  including the two new `manifold-dimension` checks.
- `scripts/run_vm_surface_tests.sh` (VM, `--profile hosted-vm
  --emit-eskb` + `eshkol-vm-standalone-test`): 57/57 passed, including
  the new "manifold dim euclidean" check.
- `scripts/run_vm_parity.sh`: 191 passed, 0 failed, 0 infra.
- `python3 scripts/check_ledger_integrity.py --self-test` and (no flag):
  both PASS.
- `python3 scripts/gate_no_silent_wrong.py --no-trace`: PASS.
- `ctest --output-on-failure -j2` (built subset — `eshkol-run`,
  `eshkol-runtime`, `stdlib`, `eshkol-vm-standalone-test`, plus the extra
  targets the full install below needed): 139 passed, 0 failed, 78 not
  run (unbuilt optional test binaries outside this change's scope — no
  genuine `***Failed`/`***Exception`/`***Timeout` anywhere in the log).

## BI-29 — `cmake --install` did not produce a `find_package(Eshkol)`-consumable prefix

`cmake/FindEshkol.cmake` (from #496) and `docs/BUILD_INTEGRATION.md`
already document `find_package(Eshkol REQUIRED)` +
`include(EshkolCompile)` against `<prefix>/share/eshkol/cmake/` as the
one downstream discovery path, but `CMakeLists.txt` had no `install()`
rule for `FindEshkol.cmake`, `EshkolCompile.cmake`, the `eshkol-runtime`
target, or the `lib/*.esk` module sources — only the release workflow and
the homebrew formula staged those by hand.

Added:
- `install(TARGETS eshkol-runtime ARCHIVE ...)` to `lib/eshkol`
- `install(FILES cmake/FindEshkol.cmake cmake/EshkolCompile.cmake ...)`
  to `share/eshkol/cmake`
- `install(DIRECTORY lib/ ... FILES_MATCHING PATTERN "*.esk")` to
  `share/eshkol/lib`, preserving `lib/`'s relative layout so every
  `(require ...)`d module (`core.manifold` included) resolves from an
  installed prefix, not just `stdlib.esk` itself
- a generated `eshkol.pc` (`cmake/eshkol.pc.in`) for non-CMake consumers,
  mirroring `Eshkol::eshkol`'s `INTERFACE_LINK_LIBRARIES` link line
  (runtime archive + `stdlib.o` + the platform's system frameworks) —
  pkg-config and `find_package` stay one contract, two discovery paths.
  Kept `FindEshkol.cmake` as the *only* CMake discovery contract — no
  separate `EshkolConfig.cmake` export, to avoid two divergent paths.

Updated `docs/platform/BUILD_NOTES.md` with a "Consuming an installed
Eshkol" section describing the installed layout and both discovery paths.

**Verification (real, end to end):**
1. `cmake --install build --prefix .scratch/install-test`
2. `cmake -S tests/integration/system_package -B .scratch/consumer-build
   -DESHKOL_CMAKE_DIR=.scratch/install-test/share/eshkol/cmake
   -DEshkol_ROOT=.scratch/install-test` — `find_package(Eshkol REQUIRED)`
   resolved all four `REQUIRED_VARS` (`-- Found Eshkol: .../eshkol-run
   (found version "1.3.4")`).
3. `cmake --build .scratch/consumer-build` — `eshkol_compile_executable()`
   compiled+linked `hello.esk`.
4. Ran the resulting binary: printed `system-package-integration-ok`.

**Side note for the maintainer:** getting `cmake --install` to complete
end to end (step 1) required building every target with an unconditional
top-level `install()` rule, which surfaced that the *vendored* zlib
FetchContent subproject's own `cmake_install.cmake` ignores the parent
`--prefix` override for its shared-library variant and writes to the
CMake-configure-time `CMAKE_INSTALL_PREFIX` (`/usr/local` on this
machine) instead. That's a pre-existing defect in the vendored
`eshkol_zlib` integration, unrelated to this PR's changes (confirmed:
another concurrent build on the same host hit the same path
independently, before this branch's install run) — not fixed here, since
it's out of this ticket's scope, but worth its own follow-up. Four files
this session's `cmake --install` wrote under `/usr/local/lib`
(`libz.1.3.1.dylib`, `libz.1.dylib`, `libz.a`, `libz.dylib`) still need
manual cleanup on that host; I could not remove them myself (blocked
outside the repo/scratch sandbox).

## Scope

No changes to `CHANGELOG.md`, `.github/`, or any docs outside
`docs/KNOWN_ISSUES.md` and `docs/platform/BUILD_NOTES.md`.

### DIFF
--- a/.icc/silent-wrong-ledger.yaml
+++ b/.icc/silent-wrong-ledger.yaml
@@ -6042,3 +6042,97 @@ entries:
       scripts/run_gpu_tests.sh and scripts/lib/test_isolation.sh's
       bash-side grep-based marker matching already operate line-by-line
       and are unaffected.
+
+  - id: SW-72
+    bucket: SILENT-WRONG
+    subtag: numeric
+    family: vm-geometric/manifold-dim
+    status: closed
+    closed_at: "fix/audit-manifold-dim-install"
+    closed_by: "branch fix/audit-manifold-dim-install"
+    title: "vm_geometric_manifold_dim returned 0 unconditionally in the ESHKOL_GEOMETRIC_ENABLED configuration, regardless of the manifold's real dimension"
+    repro: >-
+      lib/backend/vm_geometric.c:340-348 (pre-fix): under
+      `#if defined(ESHKOL_GEOMETRIC_ENABLED)` the function ignored both of
+      its arguments (`(void)vm; (void)mv;`) and unconditionally `return 0;`,
+      with a doc comment admitting "returns 0 when ESHKOL_GEOMETRIC_ENABLED,
+      per the inline TODO." `vm_geometric_manifold_dim` backs VM native id
+      858 (`manifold-dim`/`manifold-dimension`) and is also called
+      internally by `manifold-metric-tensor` (case 829) and
+      `manifold-christoffel-tensor` (case 830) to size their output tensors,
+      so every consumer of a manifold's dimension in that build
+      configuration would have silently received 0 (and downstream
+      christoffel/metric tensors would have silently degenerated to
+      `dim<=0` NIL results) rather than the manifold's actual dimension.
+      Filed as docs/KNOWN_ISSUES.md prose only, with no ledger row, so
+      `scripts/gate_no_silent_wrong.py` reported PASS despite the defect
+      (v1.3.5 docs audit, section 3 BI-4).
+    observed: >-
+      `(manifold-dim (make-hyperbolic-manifold n c))` (or any manifold
+      constructor) would return `0` instead of `n` whenever the compiler
+      was built with `-DESHKOL_GEOMETRIC_ENABLED` linking the
+      semiclassical_qllm manifold backend — a plausible-looking wrong
+      answer (a valid dimension value, just always the same wrong one), not
+      a loud error.
+    correct: >-
+      `manifold-dim`/`manifold-dimension` must return the manifold's actual
+      dimension on every engine and every build configuration: the VM's
+      portable fallback (`VmFallbackManifold.dim`, already correct), the
+      VM's qLLM-linked configuration (`qllm_manifold_get_dim`, the sibling
+      accessor `vm_geometric_manifold_curvature` already forwards to
+      `qllm_manifold_get_curvature` the same way), and the native/AOT path
+      (`lib/core/manifold.esk`'s `manifold-dimension`, `(vector-ref m 1)` on
+      its `#(type dim)` representation) — all three must agree for
+      equivalent manifolds.
+    root_cause: >-
+      `ESHKOL_GEOMETRIC_ENABLED` is not wired to any CMake option in this
+      tree (`ESHKOL_QLLM_ENABLED` only controls the separate
+      `eshkol-qllm-run` demo executable's `USE_QLLM` matmul backend, a
+      different macro), so the enabled branch of vm_geometric.c has never
+      been compiled by any build this repository can produce — the
+      curvature getter was written correctly against the real qLLM header
+      (`qllm_manifold_get_curvature`) but the dimension getter was left as
+      an unimplemented stub, and with the branch permanently dead at
+      compile time no test, sanitizer or differential run could ever have
+      caught the stub returning the wrong thing.
+    fix: >-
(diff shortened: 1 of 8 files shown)


tsotchke:
Follow-up commit pushed: excluded the vendored zlib FetchContent subproject from Eshkol's own install manifest (`SKIP_INSTALL_ALL`, scoped to just that `FetchContent_Declare`/`MakeAvailable` call).

Root cause: zlib is a private build dependency (only its static archive is copied into `eshkol-agent-zlib.a`; `eshkol-runtime` never links it), but its own `install()` calls were still walked by `cmake --install` and, against the pinned revision, resolved their destinations from this project's *configure-time* `CMAKE_INSTALL_PREFIX` rather than a later `--prefix` override — landing under the default prefix (`/usr/local` here) regardless of what a downstream packager passed to `cmake --install --prefix`.

Re-verified end to end:
- `grep -ci zlib` on a full `cmake --install build --prefix .scratch/install-test` log: **0** (previously 4 files landed under `/usr/local/lib`).
- `find /usr/local -newer <marker>` taken immediately before and immediately after the install: both empty and identical — nothing written outside the given prefix.
- `tests/integration/system_package/` still configures/builds/links/runs, printing `system-package-integration-ok`, against the resulting prefix.
- `eshkol-agent-zlib.a` still builds correctly (unaffected — the fix touches only `install()`, not the build/copy step).
- `run_manifold_tests.sh` (19/19) and `run_vm_surface_tests.sh` (57/57) still green on the rebuilt tree.

Only pcre2/sqlite/tree-sitter/Yoga's own install() rules were checked and confirmed to already respect `--prefix` correctly — this defect was specific to the pinned zlib revision, so the fix is scoped to just that one dependency.

PR #516, opened 2026-08-28
fix(vm): compile the geometric path — Riemannian builtins no longer fall back to Euclidean stubs (SW-73)

tsotchke:
## What was wrong

`lib/backend/vm_geometric.c` carries two dispatch bodies for the same 62 builtin names (ids 804-861), selected by `ESHKOL_GEOMETRIC_ENABLED`. **No target defined that macro, and nothing in the build system could** — no option, no `target_compile_definitions`, no cache variable. So the portable body was the one every CI lane, every release binary and the WASM playground compiled, and the linked-library body was unreachable from every configuration of this repository.

What the shipped body computed was the **flat answer under the curved name**, for every value of the curvature argument:

| builtin | ids | what it computed |
|---|---|---|
| `hyperbolic-exp-map` / `manifold-exp-map` / `retraction` | 809, 842 | `base + tangent` |
| `hyperbolic-log-map` / `manifold-log-map` | 810 | `point - base` |
| `spherical-log` / `spherical-log-map` | 822 | `point - base` |
| `spherical-exp` / `spherical-exp-map` | 821 | `normalize(base + tangent)` — a retraction at the wrong arc length |
| `geodesic-distance` / `poincare-distance` | 811, 816 | the L2 distance |
| `parallel-transport` / `vector-transport` | 812, 843 | the identity, endpoints popped and discarded |
| `mobius-add` | 814 | `x + y` |
| `mobius-scalar-mul` | 815 | `r * x` |
| `manifold-project` | 813 | a copy |
| `riemannian-grad` | 841 | a copy, point discarded |
| `riemannian-sgd-step` | 839 | `point - lr*grad` |
| `riemannian-adam-step` / `riemannian-adam-step!` | 840, 861 | `point + delta`, an ambient addition |
| `geodesic-attention-scores` / `-forward` | 844, 847 | scored by the ambient L2 distance |
| `exterior-derivative` | 835 | zeros of the input's shape |
| `hodge-star` | 836 | its argument, unchanged |

Every one of those except the two exterior-calculus ops **accepted a curvature argument and dropped it**. That is exactly the case `inc/eshkol/backend/frechet_mean_core.h` already names for the Fréchet mean: *"A result that carries a curvature parameter it ignores is the plausible-wrong-number case: nothing in the output shows the argument was dropped."*

It is not an approximation either. Two points of the unit Poincaré ball a tenth of a unit inside the boundary:

```scheme
(geodesic-distance #(0.9 0.0) #(-0.9 0.0) -1.0)
;; was      1.8                  (the L2 chord)
;; correct  5.888877958332881    (= 4 artanh 0.9)
```

The chord is bounded by the ball diameter; the geodesic distance diverges at the boundary, so the error is unbounded. No diagnostic, exit 0.

## Why "just turn the enabled path on" was never the fix

Measured, not assumed. A syntax-only pass of `lib/backend/eshkol_vm.c` with `-DESHKOL_GEOMETRIC_ENABLED` against a current `libsemiclassical_qllm` include directory reports **19 errors inside `vm_geometric.c`**: moved arities across `qllm_hyperbolic_exp_map` / `_log_map` / `_distance` / `_parallel_transport` / `_mobius_add` / `_mobius_scalar`, `qllm_hyperbolic_project` now taking a `qllm_tensor_t*` where the call passes a `float*`, and the SO(3)/SE(3) arms naming types (`qllm_so3_algebra_t`, `qllm_so3_t`) the library no longer declares.

The unreachable body is **stale against the API it names**, not a working implementation waiting for a switch. And it is fp32, which `frechet_mean_core.h` already documents as unable to satisfy the exact-derivative gate. So the shipped body is made to deliver the mathematics instead.

## What this does

**`inc/eshkol/backend/riemannian_core.h`** — the closed forms of constant-curvature geometry in f64: Möbius addition and scalar multiplication, exp and log maps, geodesic distance, parallel transport through the gyration `(λ_x/λ_y)·gyr[y,−x]v`, projection, and the Euclidean→Riemannian gradient conversion. A header for the same reason `frechet_mean_core.h` is one: `vm_geometric.c` is a unity-build include and `eshkol-vm-standalone-test` compiles it as a single TU, so there is no object file to call into.

The curvature argument is read as the **sectional curvature K** throughout, which is what every in-tree call site already passes (`(make-hyperbolic-manifold 2 -1.0)`, `(poincare-distance … -1.0)`) and what `eshkol_frechet_mean_compute` already takes: `K < 0` is the Poincaré ball of radius `1/√−K`, `K = 0` is Euclidean, `K > 0` is the sphere of radius `1/√K`. **Every op reduces exactly to its flat form at `K = 0`** — the old bodies were right for that one value of the argument and wrong for every other.

The formulas are the ones `lib/bridge/qllm_bridge.cpp` computes for the AD tape (`ad_hyperbolic_distance`, `ad_poincare_exp_map`, `ad_poincare_log_map`, `ad_geodesic_attention`), so the VM engine and the AD bridge now agree on what these operations mean.

**Domain violations raise** a catchable condition naming the builtin, the reason and the curvature, rather than returning a number: a point outside the ball, a point off the sphere, a log with no finite value, `mobius-add` at positive curvature. `log` refuses rather than clamping `√c|u|` to `1−1e−12`, for the reason `eshkol_frechet_log_map` already gives — a clamp substitutes a fabricated magnitude for one that does not exist.

**`exterior-derivative` and `hodge-star` raise `not implemented on the VM engine`.** Neither can be computed from what it is handed: a form's coefficient array at a single point carries no derivative information, and a flat array does not record the degree the star depends on. Returning zeros asserted that every form is closed; returning the argument asserted that the star is trivial. Listed as build items below.

**`CMakeLists.txt` gains `ESHKOL_GEOMETRIC_QLLM`** — the first way any configuration of this repository can select the linked-library body at all. OFF by default, and a `FATAL_ERROR` at configure time when `ESHKOL_QLLM_ROOT` is unset or its headers are absent, so the option cannot silently degrade to the portable body and report success.

## Tests

`tests/vm/geometric_riemannian_surface_regression.esk` — **42 assertions, each chosen so the flat answer fails it**: the near-boundary geodesic distance, Möbius non-commutativity, exp/log mutual inversion, the Riemannian length identity `λ_x‖log_x(y)‖ = d(x,y)`, transport-is-not-the-identity together with transport-preserves-Riemannian-length, the `K = 0` reductions, the sphere branch, and every refusal path. Picked up automatically by `run_vm_surface_tests.sh`'s `*_surface_regression` glob and by `language_coverage.py`'s `CI_VM_SURFACE_GLOBS`.

Two committed regressions had **encoded the wrong answers as expectations** and are repaired:

- `geometric_fallback_numeric_regression.esk` asserted `(riemannian-sgd-step p p 0.25 -1.0) = 0.75` with `p = (1.0, 1.0)` — a point of norm `√2`, not in the Poincaré ball of radius 1 at all;
- `riemannian_adam_state_regression.esk` asserted the ambient Adam update at curvature `-1.0` for the same out-of-ball point.

Both held only because the ops ignored the curvature. They now assert the flat update at `K = 0`, where it is exactly right, plus curved cases the flat answer fails. `geometric_surface_regression.esk` gets in-domain arguments for the spherical ops (their base must lie **on** the sphere) and guards around the two refusing ops.

**No `PARITY.tsv` row changes.** These 62 names are on the VM surface only — absent from `llvm_codegen.cpp` and from `eshkol_compiler.c`'s native builtin table — so the native-vs-VM differential never compares them and the ratchet governs none of them. That is also why this survived: the gate that had to catch it is the self-checking one, a probe asserting against mathematics inside a single run, which is the property `run_vm_surface_tests.sh`'s own header calls out.

## Gates

```
scripts/run_vm_surface_tests.sh          Passed: 58   Failed: 0
scripts/run_vm_parity.sh --audit-only    1 passed, 0 failed
scripts/check_ledger_integrity.py        ledger_integrity_clean: PASS (118 entries)
scripts/check_ledger_integrity.py --self-test   PASS
scripts/gate_no_silent_wrong.py --no-trace      no_open_silent_wrong: PASS
```

The full `run_vm_parity.sh` reports **189 passed, 2 failed** — both failures are `tests/vm_parity/corpus/42_iota_srfi1.esk`, a **pre-existing** native-side divergence where native `iota` ignores the SRFI-1 optional start/step and the VM honours them. `iota` lives in `lib/core/list/generate.esk` and the VM prelude, neither of which this branch touches.

## Ledger

**SW-73**, bucket `SILENT-WRONG`, **status `closed`**, evidence the new regression.

This is the **forward half** of the surface SW-70 records the backward half of: `AD_NODE_HYPERBOLIC_DISTANCE`, `AD_NODE_POINCARE_EXP_MAP`, `AD_NODE_POINCARE_LOG_MAP` and `AD_NODE_GEODESIC_ATTENTION` propagated exactly `0.0` through `eshkol_tensor_backward_dispatch`. The same four operations on the VM engine were also returning the wrong *value*, so a program using them got a Euclidean answer and, if it differentiated, a zero gradient of it.

## Coordination with #509

`docs/reference/stdlib/geometry.md` does not exist on master — it is #509's page. It is carried here **with the corrections applied on top of #509's content**, so whichever PR lands second resolves a conflict at exactly the lines that changed rather than merging a page that states `geodesic-distance` is the L2 distance. #509's page was honest about the old behaviour; it is now wrong in the other direction, which is why it could not be left alone.

## Build items (not fixed here)

1. **`exterior-derivative` (835)** — needs the form's coefficients as differentiable functions of position, not values at a point. Currently raises.
2. **`hodge-star` (836)** — needs the form's degree, which a flat coefficient array does not record. Currently raises.
3. **The `ESHKOL_GEOMETRIC_ENABLED` body is stale** — 19 compile errors against the current qLLM ABI, and it is fp32 throughout. Bringing it back (or deleting it) is its own change; no gate in this repository covers it.
4. **`ad_poincare_log_map` in `lib/bridge/qllm_bridge.cpp` clamps** `√c|u|` to `1−1e−12` instead of refusing — the same fabricated-magnitude defect `eshkol_frechet_log_map` refuses and `eshkol_rm_log_map` now refuses. Left for the AD lane rather than changed from the VM side.
5. **`spherical-project` (823)** returns the zero vector for a zero-norm input, which is not a point of the sphere. Minor, pre-existing, untouched.
6. **`curvature-gradient` (852), `curvature-hessian` (855), `adaptive-curvature-step` (856)** remain heuristics (a gradient sum, a constant `0.0`, a fixed-rate update). They do not carry a curvature argument they discard, so they are not in this defect's class, but they are not measurements either.
7. **`tests/vm_parity/corpus/42_iota_srfi1.esk`** is red on master — native `iota` drops the SRFI-1 optional arguments. Separate lane.

### DIFF
--- a/.icc/silent-wrong-ledger.yaml
+++ b/.icc/silent-wrong-ledger.yaml
@@ -6042,3 +6042,197 @@ entries:
       scripts/run_gpu_tests.sh and scripts/lib/test_isolation.sh's
       bash-side grep-based marker matching already operate line-by-line
       and are unaffected.
+
+  - id: SW-73
+    bucket: SILENT-WRONG
+    status: closed
+    closed_at: "fix/vm-geometric-enabled"
+    closed_by: "branch fix/vm-geometric-enabled"
+    title: "the VM's Riemannian geometry builtins computed their EUCLIDEAN counterparts on every shipped build: geodesic-distance was the L2 chord, mobius-add was vector addition, parallel-transport was the identity, exp/log maps were +/-, each accepting a curvature argument it discarded"
+    mechanism: >-
+      lib/backend/vm_geometric.c carries two dispatch bodies for the same 62
+      builtin names (ids 804-861), selected by ESHKOL_GEOMETRIC_ENABLED. The
+      `#if defined` body calls out to libsemiclassical_qllm, an out-of-tree
+      library. The `#if !defined` body is a self-contained "portable
+      constant-curvature fallback".
+
+      NO TARGET IN THE REPOSITORY DEFINED THAT MACRO, and until this branch
+      nothing in the build system could: there was no option, no
+      target_compile_definitions, no cache variable — `grep -rn
+      ESHKOL_GEOMETRIC_ENABLED` found hits only inside vm_geometric.c itself.
+      MEASURED while writing this entry: the linked-library body does not even
+      COMPILE against a current libsemiclassical_qllm. A syntax-only pass of
+      lib/backend/eshkol_vm.c with -DESHKOL_GEOMETRIC_ENABLED and that library's
+      include directory reports 19 errors inside vm_geometric.c — changed
+      arities across qllm_hyperbolic_exp_map/_log_map/_distance/
+      _parallel_transport/_mobius_add/_mobius_scalar, qllm_hyperbolic_project
+      now taking a qllm_tensor_t* where the call passes a float*, and the
+      SO(3)/SE(3) arms naming types the library no longer declares. So the
+      unreachable body was not a working implementation waiting to be switched
+      on; it was stale against the API it names, which is why "just define the
+      macro" was never the fix.
+      So the portable body was the one every CI lane, every release binary and
+      the WASM playground compiled, and the linked-library body was unreachable
+      from every configuration of this repository. That is a defensible
+      arrangement for an out-of-tree dependency; what made it a silent-wrong is
+      what the portable body actually computed.
+
+      It computed the FLAT answer under the curved name, for every value of the
+      curvature argument:
+
+        hyperbolic-exp-map / manifold-exp-map / retraction (809, 842)
+            base + tangent
+        hyperbolic-log-map / manifold-log-map (810)
+            point - base
+        spherical-log / spherical-log-map (822)
+            point - base
+        spherical-exp / spherical-exp-map (821)
+            normalize(base + tangent) — a retraction, right geodesic, wrong arc
+            length (atan|v| instead of |v|)
+        geodesic-distance / manifold-distance / poincare-distance (811, 816)
+            the L2 distance
+        parallel-transport / manifold-parallel-transport / vector-transport
+        (812, 843)
+            the identity on v, with the two endpoints popped and discarded
+        mobius-add (814)
+            x + y
+        mobius-scalar-mul (815)
+            r * x
+        manifold-project (813)
+            a copy
+        riemannian-grad (841)
+            a copy of the Euclidean gradient, with the point discarded
+        riemannian-sgd-step (839)
+            point - lr*grad
+        riemannian-adam-step / riemannian-adam-step! (840, 861)
+            point + adam_delta, an ambient vector addition
+        geodesic-attention-scores / geodesic-attention-forward (844, 847)
+            scored by the ambient L2 distance
+        exterior-derivative (835)
+            a zero tensor of the input's shape
+        hodge-star (836)
+            its argument, unchanged
+
+      Every one of these ops except 821/822/835/836 ACCEPTS A CURVATURE
+      ARGUMENT AND DROPPED IT. That is the same class the Frechet mean carried
+      before it was fixed, and inc/eshkol/backend/frechet_mean_core.h states it
+      exactly: "A result that carries a curvature parameter it ignores is the
+      plausible-wrong-number case: nothing in the output shows the argument was
+      dropped." A Euclidean number returned under the name geodesic-distance is
+      not an approximation of the geodesic distance — on the Poincare ball the
+      chord is bounded by the ball diameter while the geodesic distance
+      diverges at the boundary, so the two disagree without bound.
+
+      This is the FORWARD half of the surface SW-70 records the backward half
+      of. SW-70 found that AD_NODE_HYPERBOLIC_DISTANCE, AD_NODE_POINCARE_EXP_MAP,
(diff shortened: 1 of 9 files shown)


tsotchke:
Held (auto-merge disarmed): an adversarial mathematical audit of the new Riemannian formulas returned findings that must be fixed before merge (parallel transport gyration closed form, intrinsic Adam second moment, Adam state identity, curvature-sign branching shared with the AD bridge, near-boundary handling). The fixes land in a superseding PR built on this branch.

tsotchke:
Superseded by the geometry-completion PR, which carries this branch plus the fixes from the adversarial audit.

PR #523, opened 2026-08-28
test(guard): R7RS guard coverage across five axes; => clauses and test-only clauses fixed on all three code generators

tsotchke:
## What

R7RS `guard` had two kinds of coverage and neither was complete.

**One engine.** `tests/error_handling/*.esk` are the `guard` tests, and
`scripts/run_error_handling_tests.sh` compiles each one AOT and runs the binary.
No `guard` program in the repo had ever executed under `-r` or on the bytecode
VM.

**Differential, not golden.** The harnesses that *do* span engines — the P1
differential and the P5 parity ratchet — compare the engines to *each other*, so
a clause form both backends get wrong the same way compares byte-equal and
passes. The ledger already had a name for that blind spot (`SW-06`,
"shared-defect blindness").

This adds `scripts/run_guard_coverage.sh`: every case in
`tests/error_handling/guard_coverage/` is graded against a **hand-authored R7RS
golden** — not against another engine — on five axes (`jit`, `aot-o0`, `aot-o2`,
`vm-src`, `vm-eskb`). Two optimization levels are separate axes on purpose: the
differential finding this corpus inherits was an `-O1`+-only crash.

## Defects fixed (4)

All four were wrong on native **and** on the VM, which is why no differential
had ever seen them.

| | |
|---|---|
| **SW-78** | `(test => receiver)` was unimplemented in all three clause readers. Native code-generated the literal identifier `=>` as a variable reference; both bytecode compilers compiled it as an ordinary body expression. The VM's `cond` had the same hole — native gained `=>` when ESH-0109 closed and neither bytecode compiler was brought along. |
| **SW-79** | The test-only clause `(test)`, whose value R7RS defines as the test's own value, left the result unset: native substituted `'()`, the VM returned whatever was on the stack. The ESKB emitter's `cond` was *also* still dropping no-body clauses that `vm_compiler.c` had already been fixed for — the two bytecode compilers had drifted. |
| **SW-82** | **An implicit re-raise destroyed the condition's payload.** `eshkol_raise()` honours a caller-supplied value only while `g_raised_value_set_by_user` is set and clears that flag on *every* raise, so the guard's own fall-through re-raise took the fallback branch and overwrote the payload with the exception struct. `(guard (o (#t o)) (guard (i ((number? i) "n")) (raise "payload")))` answered `#<exception>`, exit 0. |
| **SW-83** | **A closure created inside a guard clause pointed at the wrong code.** The handler is inlined into its parent chunk, but only the handler's own entry pc was relocated; a nested `lambda`'s pc constant was copied verbatim. `(define f (guard (e (#t (lambda (x) (+ x 100)))) (raise 1)))` then `(f 1)` answered `1`. Adding `=>` is what made it loud — an arrow clause calls its receiver immediately, so a lambda receiver re-entered the top of the program until `STACK OVERFLOW`. |

Three code generators carry the clause-reader fixes and are kept
instruction-for-instruction aligned so they cannot drift again:
`llvm_codegen.cpp` (native), `vm_compiler.c` (`vm-src`), and the ESKB emitter's
own copy in `eshkol_compiler.c` (`vm-eskb`).

## Defects found and filed (6)

Each has a minimal repro in `tests/error_handling/guard_coverage/found/` with
expected-vs-observed in its header, and — where it is a single check — a case of
its own that keeps it **measured every run** on the axes that are correct, with
the divergent axis declared and justified in `ENGINES.tsv`. Nothing was merely
deleted.

| repro | axes | bucket |
|---|---|---|
| `weh_handler_return_swallows_condition.esk` | **all five** | `SW-84` SILENT-WRONG |
| `vm_setbang_global_from_guard_clause_lost.esk` | vm-src, vm-eskb | `SW-85` PARITY-RATCHET |
| `vm_internal_defines_in_guard_body.esk` | vm-src, vm-eskb | `SW-86` PARITY-RATCHET |
| `native_guard_variable_not_a_binding.esk` | jit, aot-o0, aot-o2 | `SW-87` LOUD-ERROR |
| `guard_wind_multi_module_dominance.esk` | jit, aot-o0, aot-o2 | `SW-88` LOUD-ERROR |
| `vm_eskb_wind_order_vs_clause.esk` | vm-eskb only | `SW-89` PARITY-RATCHET |

### `SW-84` blocks `assurance-gates` — ruling needed

`SW-84` is filed **open and in the gating bucket**, so
`scripts/gate_no_silent_wrong.py` reports `no_open_silent_wrong: FAIL` and this
PR cannot merge until it is fixed or waived. That is deliberate: a
`with-exception-handler` handler that returns from a non-continuable `raise`
swallows the condition entirely — a wrong answer, no diagnostic, exit 0 — which
is what SILENT-WRONG means. The waiver is a maintainer decision
(`owner`/`reason`/`expires`) and is **not** taken by the lane that found the
defect. Fixing it is a `with-exception-handler` change, not a `guard` change, and
is out of this branch's scope: `guard` never relies on its handler returning.

It is also the entry that makes the case for the gate. Being wrong *identically*
on every engine, it is invisible in principle to every differential harness the
project has.

## Coverage

18 cases / 3 fail-closed probes, graded on 5 axes — 93 checks required and
green, 10 `(case, axis)` pairs declared not-required with justification. Clause shapes (`(test body
…)`, `=>`, test-only, `else`), first-match-wins, the no-clause-matched re-raise
(including across two non-matching guards and an explicit re-raise from a clause
body), nesting and callee-owned guards, the guard variable's binding discipline,
`guard` × `dynamic-wind` escape ordering and after-thunk counts, `guard` ×
`with-exception-handler` × `call/cc`, `error` with the error-object accessors,
every raised value kind, and `guard` in each non-tail expression context.

Tail position is deliberately out of scope — that is `SW-58`, pinned by
`tests/tco/guard_tail_context/`.

Two structural stages keep the gate from going vacuous: no golden may contain a
`FAIL:` token or drift from its case's `chk` names (so a red case cannot be
"fixed" by editing its expectation), and a fail-closed probe that dies at
*compile* time is rejected rather than credited as failing closed.

## Docs

`SW-80` corrects three standing claims in
`docs/reference/language/error-handling.md` and `INDEX.md`: `raise-continuable`
is unimplemented on **every** substrate (not "VM-only" — it appears in no
builtin table, no dispatch and no prelude); the error-object accessors **do**
work natively (that limitation notice was stale, and the VM is the side with the
`-irritants` stub); and the "differential `guard` findings" section attributed
two closed findings to `ESH-0101`/`ESH-0102`, which denote a recursion-depth
stack-guard item and mutual tail-call optimization respectively. `ESH-0101` is
**not** closed by this branch.

## Wiring

CI `pillars-fast` (which already builds the three targets it needs),
`docs/TESTING.md`, and the `guard_coverage_gate` criterion in
`.icc/completion-oracles.yaml`.

## Evidence

Built and measured on a Linux host matching ci.yml's `linux-x64-lite`
configuration (LLVM 21, XLA/GPU off). Gate summary and per-axis results in the
run log; `SW-82` and `SW-83` additionally verified by direct probe before and
after the fix (`#<exception>` → `payload`; `1` → `101`).


tsotchke:

PR #527, opened 2026-08-28
fix(ad): geometric bridge ops refuse off-manifold inputs and branch on the sign of curvature (SW-76)

tsotchke:
The AD bridge no longer clamps sqrt(c)*|u| toward the ball boundary (which fabricated finite log-map magnitudes); inside the ball the stable asinh/Möbius-numerator forms shared with the VM (inc/eshkol/backend/riemannian_core.h) are exact, and on or outside the ball the op refuses with a message naming the op and norm. Curvature is branched on sign: c = -K only for K<0; the Poincaré-only producers refuse K>=0; geodesic attention gains flat and spherical branches matching the VM and validates every row before scoring instead of producing NaN. Golden fixtures byte-identical. The conformal-factor convention is unchanged pending the maintainer's ruling. Ledger SW-76 closed.


tsotchke:

PR #529, opened 2026-08-28
ci: the mesh is the primary CI executor — ESHKOL_MESH_PRIMARY routes Linux contexts to self-hosted runners

tsotchke:
With the repository variable ESHKOL_MESH_PRIMARY=on, the Linux x64 contexts (xla, asan/ubsan, wasm, lite, guard, assurance-gates, surface-manifest) run on the self-hosted linux-mesh runners and linux-x64-cuda on the CUDA node; job names are unchanged so the required checks keep working. Mesh-routed jobs use a fail-loud toolchain preflight (no apt-get), per-job build directories cleaned at the end, and no sudo. Reserved variables route macOS, Windows and ARM64 lanes to the mesh once such runners are registered; hosted runners remain only for platforms the mesh lacks. Advisory mesh lanes are deduplicated against primary ones. Fork-PR approval requirements documented in SELF_HOSTED_RUNNERS.md.

### DIFF
--- a/.github/workflows/ci-mesh.yml
+++ b/.github/workflows/ci-mesh.yml
@@ -187,7 +187,7 @@ jobs:
     runs-on: ${{ fromJSON(matrix.labels) }}
     timeout-minutes: ${{ matrix.timeout }}
     env:
-      MESH_JOB_BUILD_DIR: ${{ github.workspace }}/.mesh-build-${{ github.run_id }}-${{ github.run_attempt }}-${{ matrix.name }}
+      MESH_JOB_BUILD_DIR: .mesh-build-${{ github.run_id }}-${{ github.run_attempt }}-${{ matrix.name }}
     strategy:
       fail-fast: false
       # One runner per node; don't let two lanes fight over the same box's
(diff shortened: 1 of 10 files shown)


tsotchke:
Pushed `b13de1d6`: the three coordination jobs that gate everything downstream were still on `ubuntu-22.04` — `changes` (ten jobs `needs:` it), `mesh-preflight` (mesh-matrix is `needs:` it and gated on its output), and `docs-only-required-context-stubs` (on a docs-only PR it is the *only* producer of the required contexts).

Measured on master, hosted jobs currently die in 2–6 seconds with `steps = 0` and no log. So as this PR stood, the lanes would have been correctly routed to the mesh and never dispatched, and a docs-only PR would have been left permanently unmergeable with nothing to re-run — the failure #455 closed, reopened through a different door.

All three are coordination only, so they take a broad `["self-hosted", "eshkol"]` selector rather than a lane node's labels. The hosted fallback is unchanged for fork PRs and mesh-off.

Two prerequisites outside the diff are now in place: the `linux-mesh` label this workflow asks for existed on no registered runner (the linux lite/xla/asan lanes would have queued forever), and `ESHKOL_MESH_MACOS_LABELS` was read but never set. `ESHKOL_MESH_ARM64_LABELS` stays unset on purpose — no linux-arm64 runner is registered and the empty-string guard falls back to hosted rather than queueing.

`ESHKOL_MESH_PRIMARY` remains unset, so none of this is live yet.

tsotchke:
Closing for now: moving the primary CI executor to self-hosted runners is deferred until after the v1.3.5-evolve release, so the runner topology does not change during the final release dry run. The branch is kept and will be revisited as a fresh change against master after the tag.

PR #530, opened 2026-08-28
fix(runtime): stack overflow in user recursion reports a diagnostic instead of SIGILL (ESH-0101, SW-81)

tsotchke:
Deep non-tail recursion past the stack now produces 'eshkol: stack overflow ...' naming the stack size and ESHKOL_STACK_SIZE, with a nonzero exit, on JIT and AOT and inside parallel-map workers (sigaltstack per thread); with a larger ESHKOL_STACK_SIZE the same program completes. The former SIGILL was the trap on the guard page with no handler. tests/stress/found reproducer promoted to a hard gate (8/8), task ESH-0101/ESH-0112 verified, ledger SW-81 closed, runtime docs updated.


tsotchke:

PR #533, opened 2026-08-28
fix(exceptions): a handler returning from a non-continuable raise now raises a secondary exception on every engine (SW-84)

tsotchke:
R7RS 6.11: when a with-exception-handler handler returns from a non-continuable raise, a secondary exception is raised in the handler's dynamic environment. Native codegen, the VM source path and the ESKB path (new OP_RAISE_SECONDARY) all do this now, with the original condition's message in the secondary diagnostic; raise-continuable is unchanged. Tests: nested handlers, handler-raises-inside-handler, VM parity corpus and found cases. Ledger SW-84 closed with five-axis evidence.


tsotchke:

PR #541, opened 2026-08-28
fix(tco): self-recursive tail calls inside guard bodies run in constant stack with exact handler semantics on both engines (SW-58)

tsotchke:
Makes the FEATURE_MATRIX claim true: a self-recursive tail call in a guard body is constant-stack while the guard's handler still catches a raise from the next iteration, with handler-chain and dynamic-wind state identical to the non-TCO program. Native guard replay semantics verified and retained; the VM gains direct self-tail OP_TAIL_CALL under guards, collapsible catch-all guards, growable handler chains, per-handler value snapshots/restoration, and continuation/region-evacuation root handling. Tests: 1e6-iteration recursion inside guard (flat stack/RSS), raise on iteration N caught, nested guards, guard+dynamic-wind, escape_matrix and VM parity. tail-calls.md and FEATURE_MATRIX updated; ledger SW-58 closed.


tsotchke:

PR #544, opened 2026-08-28
fix(vm): load path resolution, top-level predeclaration, tail propagation through when/unless/and/or/let, first-class accessors, and native parameter evacuation (LE-05, LE-08, LE-13, SW-39, IF-08)

tsotchke:
VM load uses the shared resolver (nested and path-literal loads, ESKB source context); top-level VM procedures are predeclared without reordering intervening expressions (mutual recursion grouping); when/unless/and/or and local-binding bodies propagate tail position via OP_TAIL_CALL_POPN; first-class use of string-length/string-ref/vref covered with the regenerated prelude cache; native HEAP_SUBTYPE_PARAMETER evacuation deep-walks converter, current and dynamic values. Native JIT/AOT and VM/ESKB fixtures, CTest entries, docs and ledger closures included.


tsotchke:

PR #557, opened 2026-08-29
fix(tensor): malformed shapes are rejected with diagnostics on every engine instead of crashing, returning junk or hanging (SW-110)

### LINKED ISSUE #550: [v1.3.4-evolve] malformed tensor shapes can crash, return junk, or hang
Observed with clean `tsotchke/eshkol@90cbd7130f47b8184bcc77b8d5c1b0026da980de`, `Eshkol Compiler v1.3.4-evolve`, compiler SHA-256 `caa295b19a6e9388963aa0def99dade63656d2dcbffccad421bd1daaa1db3750` on Linux x86-64. Upstream HEAD was not rebuilt.

Three minimal malformed-shape programs do not fail safely:

```scheme
;; incompatible broadcast: AOT and JIT exit 139
(display (tensor-add (reshape #(1.0 2.0 3.0 4.0 5.0 6.0) 2 3) #(10.0 20.0)))

;; wrong reshape element count: AOT/JIT exit 0 and print junk
(display (reshape #(1.0 2.0 3.0) 2 2))
;; observed: #((1 2) (3 1e-323))

;; negative dimension: AOT exits 0 and reaches the acceptance marker;
;; JIT does not finish within 90 seconds
(define malformed (reshape #(1.0 2.0 3.0) -1 3))
(display "MALFORMED_INPUT_ACCEPTED") (newline)
```

Commands:

```bash
eshkol-run repro.esk -o repro -L /path/to/build && ./repro
eshkol-run -r repro.esk -L /path/to/build
```

Expected: a bounded, actionable shape/dimension error in both engines before any invalid allocation or access. Actual: SIGSEGV, silent junk padding, or nontermination depending on the case/engine.

tsotchke:
Fixes #550. Unchecked dimension products, reshape count mismatches, ignored broadcast failures and VM shape coercion to truncated values let malformed shapes crash, return junk or hang. A central checked shape/product validator with resource-limit enforcement now guards native LLVM/JIT/AOT, the VM and the XLA boundary; every malformed shape is rejected with a diagnostic. Reported reproducer and adversarial variants covered on both engines; ledger SW-110 closed.


tsotchke:

PR #558, opened 2026-08-29
fix(ml): cross-entropy validates its targets and computes the documented loss for indexed and one-hot forms (SW-112)

### LINKED ISSUE #552: [v1.3.4-evolve] indexed cross-entropy target returns wrong loss instead of rejecting
Observed with clean `tsotchke/eshkol@90cbd7130f47b8184bcc77b8d5c1b0026da980de`, `Eshkol Compiler v1.3.4-evolve`, compiler SHA-256 `caa295b19a6e9388963aa0def99dade63656d2dcbffccad421bd1daaa1db3750` on Linux x86-64. Upstream HEAD was not rebuilt.

```scheme
(define logits (reshape #(2.0 0.0 0.0 2.0) 2 2))
(define indices #(0 1))
(display (cross-entropy-loss logits indices))
```

AOT and JIT both exit 0 and print `2.8200751916029176`. For indexed targets `[0,1]`, the mean softmax cross-entropy for these logits is approximately `0.1269280110429726`.

If `cross-entropy-loss` supports indexed class targets, the returned value is wrong. If it accepts only one-hot targets, the rank/shape mismatch should be rejected explicitly rather than interpreted as another layout.

tsotchke:
Fixes #552. A shared validated cross-entropy kernel now serves native, VM and the bridge AD: it supports normalized probability/one-hot targets and indexed targets, validating shape, finiteness, normalization, integrality and range, and rejects anything else instead of returning a wrong loss; the VM exposes cross-entropy-loss. Backward matches forward on both engines; ledger SW-112 closed.


tsotchke:
Superseded by #633 (same commits; this branch stopped receiving workflow runs).

PR #559, opened 2026-08-29
fix(tensor): one validation layer for shape products, metadata, broadcasting and indexing; adjacent defects closed (SW-114+)

tsotchke:
Closes the class behind the recent tensor issues rather than the instances: a shared tensor validation API (checked shape products, metadata, broadcasting, per-axis indexing) used by every constructor and op entry on native and VM, plus the adjacent defects a user would hit next — attention shape relations beyond rank, layer-norm parameter shape contract (no modulo wrap), tensor-get bounds instead of aliasing, Box-Muller log(0) guard. Adversarial fixtures per defect on both engines with VM parity; ledger entries SW-114 upward.


tsotchke:

PR #561, opened 2026-08-29
feat(ad): exact-coefficient Taylor and user-number AD re-cut on the registry; the VM returns exact derivatives for exact inputs (P6/P11)

tsotchke:
Re-cut of the deferred exact-coefficient/user-number AD work on the current dispatch registry: exact rational VM dual halves and a K+1 Taylor carrier, VM taylor and derivative-n, so exact rational inputs yield exact results on the VM as they do natively. Acceptance: the exact suite and reverse-taylor suite pass together with the existing AD/Taylor gates and exhaustive dispatch.


tsotchke:

PR #562, opened 2026-08-29
feat(gpu): WebGPU backend for the WASM target inside the ordinary GPU dispatch (with explicit CPU fallback and refusal)

tsotchke:
Adds an ESHKOL_GPU_WEBGPU dispatch seam selected by the same runtime dispatch that picks Metal and CUDA, with Emscripten backend selection, a WebGPU C backend, WGSL kernels for the ops the GPU correctness gate covers, and explicit CPU fallback or refusal where a kernel cannot meet the gate tolerance. JS glue stubs for every new runtime function; docs and feature matrix updated to the verified status.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1363,6 +1363,14 @@ jobs:
           set -euxo pipefail
           python3 scripts/check_wasm_imports.py --build-dir "$BUILD_DIR"
 
+      # Browser-independent WebGPU glue contracts belong in the ordinary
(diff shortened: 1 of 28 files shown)


tsotchke:
Superseded by #715, merged into master as 2630d3a23: this PR's four commits were carried onto current master, with the double-float kernels replaced by exact IEEE f64 WGSL kernels verified bit-identical to the CPU path in a real browser. The browser VM path follows in a separate pull request.

PR #589, opened 2026-09-01
test(ml): exercise the canonical dataloader-has-next? predicate (language-surface coverage 1108/1108)

tsotchke:
Since #512 made dataloader-has-next? the canonical name, tests/ml/dataloader_test.esk only exercised the alias, so the language-surface coverage gate stood at 1107/1108 and the quantum-macos lane (which runs that gate) has been red on master. The test now calls the canonical predicate and keeps exactly one alias call so both names carry execution evidence. Verified with the focused test on the mesh; the coverage run is part of the quantum lane.


tsotchke:
Folded into #592 (cherry-picked as 72ffa99b) because the coverage ledger there has to be regenerated at 1109 constructs with this test's evidence included; this PR can close once #592 lands.

PR #588, opened 2026-09-01
fix(release): remove post-merge conflict artifact

tsotchke:
Safety rollback for the current release train. The initial marker-only candidate 948021c7 was rejected by the full Enki suite: 13/246 failures including a real SIGSEGV in the 31-capture parallel closure gate, FFI diagnostic regression, and ABI-ratchet failure. Current candidate 34d92169 (tree a7d42b28) reverts the entire unsafe #565/#548 codegen tranche while preserving the previously verified train through #566, plus three whitespace cleanups. Do not merge until complete ICC mesh builds and full CTest pass on macOS ARM and Linux x64, followed by platform-focused CUDA/Windows/WASM/sanitizer gates.

PR #592, opened 2026-09-01
fix(codegen,vm): restore the llvm_codegen.cpp hunks dropped by the codegen extraction; VM manifold builtins stop shadowing core.manifold; defined shifts in the packed-string emitters

tsotchke:
The codegen extraction (#548, commit bca3a6d2) carried a stale copy of lib/backend/llvm_codegen.cpp: the extraction was cut from an older revision of the file, so every change that had landed in that file since then was silently undone while the sibling modules and runtime headers kept the newer contracts. The affected changes are 893c7747 (shared Unicode classifier for the char predicates), 90203a8e (module privacy guards), 5e79cf31 (source-level symbol names in arity diagnostics), a324f35a (length-explicit string interning), d9e0740c (fatal-error propagation), 53b8d264 (eshkol_arity_error_current call sites), eba46f59 (32-bit closure capture-count encoding), f34d846c (continuation state flags) and 6a721b16 (file-stat return type). The most visible consequence was the 31-capture parallel closure crash: the lambda emitter wrote the 16-bit capture layout while the parallel dispatcher and CLOSURE_ENV_GET_NUM_CAPTURES read the 32-bit one, so a closure with 31 captures was dispatched as if it had 65567 and called through the wrong ABI.

This PR restores every dropped hunk in place (audited commit by commit against the extracted modules), and fixes two further defects the gate surfaced on the same master:

- The VM builtin table aliased public core.manifold names (manifold-type, make-hyperbolic-manifold, manifold-parallel-transport, ...) to lower-level native ops with different representations, and the VM bootstrap rule that prefers a VM builtin over a stdlib wrapper silently discarded the module's real definitions. The raw ops now carry their own names (the -handle family), core.manifold owns its public API on every engine, the two probe tests that exercised the raw ops are updated, and the VM prelude cache and language-surface manifest are regenerated.
- The packed-string constant emitter in the VM compiler accumulated bytes into a signed 64-bit value with shifts up to 56 bits; it now packs into an unsigned accumulator and converts once, with identical constants.
- The object-header layout comment that the extraction copied into inc/eshkol/backend/llvm_codegen.h now refers to eshkol_object_header_t instead of restating the layout, which is what the ABI header-inventory ratchet asks for.

Verification (Linux x64, LLVM 21, Release, tests on): the five previously failing tests pass (0/5 on the unpatched tree); ctest -R 'closure|parallel|ffi_boundary|abi_header' 16/16; full CTest 251/253 against 247/253 unpatched; wasm-execute-diff 94 passed / 0 failed (was 92/2); VM surface tests 58/58 (was 57/1); SW-74 and file-stat suites 8/8, 10/10, 27/0, 8/0; ASan+UBSan build of the VM/parity/string subset with the vm_compiler.c:3469 report gone; ABI header inventory check (830 sites) and selftest, surface counts, prelude-cache builtins (732/732), ledger integrity and silent-wrong gates all pass. The one remaining CTest failure on this base, v1_3_jit_cache_runtime_smoke, fails identically on the unpatched tree and is tracked separately.

Carry audit note: e156c92d's codegen half (AOT IR telemetry) was removed by 893c7747, not by the extraction, and is left as a separate decision; the generated docs/api files were already stale on this base and are not touched here.

Supersedes #588: the revert would remove #565 and #548 wholesale to cure a defect that lives in one carried file.

Follow-up commits on this branch: the canonical dataloader-has-next? test from #589 (72ffa99b); the execution-backed language-coverage floor ratcheted to 1109/1109 with a regenerated ledger from a full traced run including the quantum corpus (f86782ff); and a committed VM surface check for the process-spawn-argv-with-options alias, which previously had no execution evidence anywhere in the tree (df47eba8).


tsotchke:
Mesh gate evidence for df47eba8 (hosted Actions unavailable): macOS arm64, wasm (94/0), ASan+UBSan, surface-manifest and the Linux x64 CUDA lane pass; Linux x64 252/253 with the only failure being the pre-existing JIT-cache smoke (byte-heuristic check, addressed in #593); Linux arm64 251/252 with the known sweep timeout. On a fresh Linux x64 node the full linux-x64 battery passes outright (252/252 plus every run_all_tests suite), and a fresh Linux arm64 node reports 252/252. The readiness probes that fail on this tree fail identically on master a619b41c (VM heap-budget watchdog, p8 axis-6, and the VM letrec even/odd divergence in recursion_test.esk); this branch reduces engine-parity divergence from 7 programs to 6 and VM-only failures from 3 to 1. Fixes for those master-level defects follow as separate PRs.

PR #595, opened 2026-09-02
build(deps): fetch the pinned Eigen commit from the official GitLab repository where it is reachable

tsotchke:
FetchContent pins Eigen at bc3b39870ecb690a623a3f49149a358b95c5781d (the 5.0.1 tag). The GitHub eigenteam mirror the pin was fetched from does not advertise that commit (its tags stop at 3.3.7), so a fresh node's FetchContent clone fails; the existing mesh nodes only build because a seeded dependency cache hides it, and the new Windows lane node had to be seeded by hand. The pin now fetches from https://gitlab.com/libeigen/eigen.git, where the commit is reachable from the 5.0 branch and the 5.0.1 tag; the tag hash and GIT_SHALLOW FALSE are unchanged. Verified by resolving the commit on the GitLab remote; the mesh gate exercises the fetch on fresh dependency trees.


tsotchke:
Superseded by merged PR #610 (master commit 5ce74beeac25aca56c0c8129083e6db77f96f7dc). Verified that master contains the complete Eigen repository change and the unchanged pinned commit bc3b39870ecb690a623a3f49149a358b95c5781d. Closing this duplicate; the fix is already on master.

PR #596, opened 2026-09-02
docs(test): specify ESKM v1 and add golden compatibility fixtures

Gabriel-Kahen:
## Problem

ESKM v1 had implementations but no normative byte-level contract or historical golden corpus. This PR makes the existing format independently checkable, preserving its bytes and documenting backend restrictions separately from wire validity. The final acceptance update adds the roadmap's missing bounded large positive case.

## Scope

- `docs/reference/tensors/eskm-v1.md` and its link in `docs/reference/tensors/INDEX.md`: header/record layout, little-endian integer widths, binary64 bits, CRC-32 coverage, count/size rules, versions, names, and historical native/VM restrictions.
- `tests/core/fixtures/eskm-v1/`: six accepted historical files (scalar, zero extent, ordinary, rank 8, multi-tensor, and 32×32) and six existing malformed files; a manifest with SHA-256 digests, exact metadata/payload bits, provenance, and structured mutation recipes.
- `scripts/check_eskm_v1_fixtures.py`: standard-library-only corpus checker and checksummed-payload negative self-test.

The corpus totals 9,160 bytes. The added 8,237-byte `large-32x32.eskm` has 1,024 values `(i - 512) / 8`, for signed flat index `i` from 0 through 1023. All eleven previously reviewed fixture files are unchanged. This changes no loader, public API, wire byte rule, resource limit, or product behavior. The specification distinguishes ESKM from native compiler tensor-save/tensor-load dispatch's separate ESKT format.

## Provenance and revisions

Accepted files are emitted by `write_checkpoint` from annotated tag `v1.2.4`:

- Tag object: `4e07f166a7a0da28d24c78fb1c1af4258c4c1845`
- Producer commit: `b98dc8b32399de739a037e9fa0a470bf0426eca9`
- Tagged `lib/core/model_io.cpp` SHA-256: `53e10801ef30701705e643462418bde4649dfc975f0c8731f7762dc8527de889`
- PR comparison base: `ea81a85448300858f92a33bbc676dc71f91ce06b`
- Verified head: `5ad058b2bc416c97a76c21f246d6b3bb4044b48f`
- Latest fetched upstream/master: `5ce74bee` (the model-I/O implementation, header, and native test are unchanged from the comparison base).

The historical writer was compiled from an unmodified `git archive v1.2.4` source/header extraction with a small positive-input driver. It reproduced all six accepted files byte-for-byte. Every malformed fixture remains a checker-reconstructed mutation of `ordinary-2x3.eskm`.

## Verification

Refreshed on 2026-09-07, Linux x86-64, at the exact head above.

PASS:

- `python3 scripts/check_eskm_v1_fixtures.py`: 6 accepted, 6 rejected; all 12 SHA-256 digests match.
- `python3 scripts/check_eskm_v1_fixtures.py --self-test`: accepts the corpus and rejects a changed scalar payload even after its CRC and manifest hash are refreshed.
- Historical v1.2.4 producer identity/source digest and 6/6 byte-identical writer reproductions.
- Native `model_io_test`: exact-head `tests/core/model_io_test.cpp` and `lib/core/model_io.cpp` freshly compiled with C++20/LLVM 21 headers and linked to the existing read-only LLVM 21 build's runtime support archive. The support build's `lib/` and `inc/` sources match this head. Test passes, including its existing 1024×256 round trip; isolated `TMPDIR` avoids other tasks' files. This is a focused rebuild, not a full fresh CMake build.
- `python3 scripts/check_test_coverage.py`: 46 suites.
- Python syntax, manifest JSON, relative documentation links, and `git diff ea81a854 HEAD --check`.
- Independent read-only contract review against the pinned writer/reader sources.

FAIL (unchanged base inputs; no generated files modified):

- `python3 scripts/gen_api_docs.py --check`: stale `docs/api/INDEX.md`, `docs/api/README.md`, and `docs/api/backend/llvm_codegen.md`. Reproduced the same three failures in a clean `git archive ea81a854` extraction.
- `python3 scripts/check_surface_counts.py`: eight stale references to 1108 where the canonical count is 1109; checker and all affected inputs are unchanged from the comparison base.

NOT RUN:

- Full CMake build/CTest battery and release gate on this final head. The maintainer's earlier full-battery report applies to `6708c063`, not this SHA.
- Native JIT/AOT and VM source/bytecode fixture matrix here; dependent #597 owns and is updating that acceptance evidence for the sixth positive file.
- macOS, Windows, and big-endian hosts.

## Dependencies and remaining acceptance

This behavior-neutral corpus PR has no loader-hardening dependency and remains first in the maintainer's stated sequence. Dependent #597 owns runtime CTest coverage, the four-engine matrix, and its release-oracle integration; #601 owns its separate bounded-resource gate. Neither is duplicated here. The #597 owner has the exact new fixture/head and is incorporating it.

The normative contract and historical corpus checks pass locally. Final runtime-matrix evidence belongs to #597, and full-battery acceptance for this exact head remains outstanding. PR is open; filing and local checks do not constitute merge or maintainer acceptance.


tsotchke:
Thank you for this, and apologies for the silence — six PRs and five issues landed while the release train was head-down on a separate correctness wave, and you deserved an answer well before now.

I reviewed this one properly rather than taking the report on trust, and re-ran the parts that are independently checkable:

```
check_eskm_v1_fixtures.py self-test:
  [OK] committed corpus passes
  [OK] checksummed payload drift is rejected
self-test: PASS

ESKM v1 fixtures: PASS (5 accepted, 6 rejected)
```

The provenance block reproduces exactly, which is the part that makes the corpus worth having:

| | claimed | reproduced |
|---|---|---|
| tag object | `4e07f166…c1845` | matches |
| peeled commit | `b98dc8b3…6eca9` | matches |
| tagged `model_io.cpp` SHA-256 | `53e10801…7de889` | matches |

Pinning the fixtures to a tagged writer and publishing the digests is what lets a later reader tell "the format changed" from "the corpus rotted", and it is the reason this can be a base for the gates in #597 and #601 rather than another set of magic bytes. The ESKM/ESKT distinction is a genuinely useful thing to have written down; that ambiguity has cost time before.

### Sequencing

The intended order, with the caveat that merges are the maintainer's call and the queue is currently behind an in-flight release:

1. **#596** first — it is the base the rest of the ESKM work stands on, and it is behavior-neutral.
2. **#597** after it, at which point it reduces to the two focused commits you describe.
3. **#598** and **#600** are independent of that chain and can go whenever they are gated.
4. **#601** and **#602** stay stacked on #555 as you have them; once #555 lands they want the rebase and retarget to master you already called out. Keeping them stacked was the right call — it kept the loader hardening out of both review diffs.

### One ask

Neither gate is currently visible to our release oracle. `.icc/completion-oracles.yaml` has no ESKM criterion, so `eskm_v1_model_load_engine_parity` and `eskm_model_fuzz_smoke` will pass in CTest and still contribute nothing to the readiness verdict that decides whether a release is done. Our standing rule is that every new pillar gets wired in, and these two qualify.

The mechanism is a JSON-L event under `scripts/icc_traces/` plus a criterion of this shape:

```yaml
      - runtime_event:
          event_kinds: [eshkol_smoke]
          event_names: ["eskm_v1_model_load_engine_parity"]
          event_values: ["PASS"]
        severity: high
        label: 'ESKM v1: all 16 producer->consumer engine pairings agree byte-for-byte'
        action: ./scripts/run_eskm_v1_model_load_parity.sh
```

Happy for that to be a follow-up rather than growing this PR — say which you prefer and I will either wire it myself or leave it to you.

### On CI

GitHub Actions is not currently running for this repository, so nothing will report on your PRs no matter how long they sit; that is not a comment on the work. We gate on our own runners instead, and PR heads are fetchable there, so I can put your branches through the full battery and post the verdicts here. I will start with this one.

One more thing, from #598. You reported three failures in the v1.2 edge-case suite as unrelated baseline noise — `channels_test` passing every assertion and then taking SIGSEGV during teardown, `parallel_ad_worker_init_test` losing an externally submitted task, and `parallel_hash_table_mutation_test` on parallel insertion and key visibility. Reproducing them against a pre-filing run is what establishes they are not yours, and that part holds.

What does not hold is the "baseline" framing, and that is on us rather than on you: I went looking and **none of the three is recorded anywhere** — not in `docs/KNOWN_ISSUES.md`, not in our internal handoff notes. They have been failing without a record, which is how a real defect gets absorbed as ambient noise. A crash during teardown after every assertion passes is the kind of thing that is easy to wave through and is often a genuine lifetime bug. I am filing all three so they stop being folklore. Thank you for writing them down precisely enough to act on.

tsotchke:
Full battery on this PR's head (`6708c063`), `linux-x64` lane on our own gate node — LLVM 21, Release, 253 CTest entries:

```
98% tests passed, 6 tests failed out of 253
```

The six:

```
 30  parallel_closure_capture_boundaries_native_aot
 94  v1_3_jit_cache_runtime_smoke
 99  ffi_boundary_fail_open_test
135  abi_header_inventory_ratchet
136  abi_header_inventory_semantic_ratchet
137  abi_header_inventory_selftest
```

**None of them is yours.** I checked each against the gate's own evidence for master at `a619b41c`, which this branch is based on — master fails exactly these same six on its own, in three independent runs. They are the known red set from a stale copy of `llvm_codegen.cpp` that the codegen extraction carried, and #592 is the fix-forward for all of them. Your diff is docs, fixtures and a standard-library-only checker; it does not touch any of those paths, and the other 247 tests pass.

So the honest reading is: this PR is clean, it is sitting on a red base, and it goes green the moment #592 lands underneath it. I will re-run it then so the record shows a clean battery under its own SHA rather than an inherited one.

Gabriel-Kahen:
Thank you. I handled the release-oracle wiring and kept it inside the existing gate PRs rather than opening separate micro-PRs: #597 now gates the 16-cell parity matrix, and #601 gates the bounded fuzz smoke. Both use fail-closed evidence and negative controls. I also rebased this PR onto current master after #592/#593 and reran the focused LLVM 21 checks; the updated evidence is in the PR description.

PR #597, opened 2026-09-02
test(model-io): add ESKM v1 cross-engine parity gate

Gabriel-Kahen:
Depends on #596 at `5ad058b2bc416c97a76c21f246d6b3bb4044b48f`. This branch is rebased onto that exact dependency head; it remains an open PR, not merged or accepted.

The repository lacked literal ESKM model producer-to-consumer evidence across native JIT, native AOT, VM source and VM bytecode. This PR executes all 16 combinations and compares semantic metadata and complete serialized bytes. During verification, a wrong final payload expectation exposed an additional test defect: naming the callback `values` bypassed it on native engines. The callback is now named `payload-check`, and every engine must detect that payload mismatch in the self-test.

## Scope and behavior

Focused files above #596: `CMakeLists.txt`, `scripts/run_eskm_v1_model_load_parity.sh`, `tests/core/eskm_v1_model_load_parity.esk`, `docs/reference/tensors/eskm-v1.md`, plus the existing five-line smoke integration and eight-line `.icc/completion-oracles.yaml` criterion.

- Four producers write the ordered two-tensor model; all four outputs must match historical v1.2.4 ESKM v1 bytes.
- Each consumer reads all four outputs, checks names/order, shapes/rank, f64 dtype and payload, then rewrites exactly. Direct count assertions are native-only; VM shape and byte comparisons also constrain count.
- Every consumer reads/rewrites ordinary, rank-8, multi-tensor and the new bounded 32×32 fixture. All 1,024 large-fixture values are checked with bounded recursion depth.
- Six existing malformed fixtures must be refused per consumer. Three negative controls per consumer cover wrong model structure, unexpected malformed-file acceptance and a wrong final payload expectation.
- Only a successful normal matrix run supplies the exact PASS event required by the smoke criterion. Self-test success cannot satisfy it.

No loader, writer, allocation policy, public API or wire-format implementation changes.

## Verification — 2026-09-07

Head: `fc22ceb200de5919c0476a76d577878a47ad74c6`.
Current PR base / fetched upstream master: `5ce74beeac25aca56c0c8129083e6db77f96f7dc`.
Dependency stack original base: `ea81a85448300858f92a33bbc676dc71f91ce06b` (current master not incorporated; dependency stack retained).

PASS:

- Fresh private Linux x86-64 Release build with LLVM 21.1.8; optional GPU, BLAS and agent FFI disabled; at most two compiler jobs.
- `ctest --test-dir build-pr597 --output-on-failure -R 'model_io|eskm_v1|vm_canonical_stdlib|vm_prelude_cache'`: **5/5**, including **16/16 model pairs**, **4/4 producer golden comparisons**, **12/12 negative controls**, model-I/O unit tests and VM stdlib/cache checks.
- A temporary copy restoring the faulty `values` callback name makes the new self-test fail for both native engines (exit 1), proving this regression is detected.
- Actual isolated smoke `probe`/`emit_event` code emits INFRA for missing ESKM execution; the actual configured exact filters exclude missing, INFRA, FAIL, self-test-only and unrelated evidence. A private ICC evaluator was not used for this check.
- Fixture checker (6 accepted/6 rejected) and its payload-drift self-test; outcome taxonomy 17/17; test inventory; oracle schema/bindings/false-green checks and self-tests; shell syntax and `git diff --check`.

BASELINE FAIL:

- API-doc check reports stale generated INDEX/README/backend llvm_codegen pages.
- Surface-count check finds 1108 references against canonical 1109.
- Relevant inputs and checker scripts for both failures are unchanged from the dependency stack original base; no generated files were edited.

NOT RUN / acceptance limits:

- Private ICC CLI, full smoke/release pipeline, full repository suite, sanitizers, macOS and Windows.
- Scalar and zero-extent fixtures remain structural-corpus cases because VM materialization does not support them. This is explicitly a backend-common model matrix, not all-valid-fixture parity.
- Native/current tensor APIs use separate ESKT files. This ESKM model matrix does not establish the roadmap's tensor-file 16-way acceptance; that requires separate evidence.

## Merge compatibility follow-up

Commit `fc22ceb200de5919c0476a76d577878a47ad74c6` changes only CMake registration placement. The ESKM block now follows the standalone VM test section with the same test commands/properties and equivalent explicit build/platform/target guards, avoiding adjacent insertions from sibling PRs. Normal fast-forward push; dependency unchanged.

PASS: fresh configure and the same focused CTest selection **5/5** after relocation. `git merge-tree --write-tree` is clean against master `5ce74beeac25aca56c0c8129083e6db77f96f7dc`, #600 `83467dde5fc6ac427436d457d821af094238a93b`, and #615 `44b366442f1d60d39dfdb63b2033e7ebbbb0672d`. These are merge-conflict checks, not builds or acceptance of the combined trees.


tsotchke:
Answered at length on #596, so just the part specific to this one.

Executing all 16 producer→consumer pairings rather than inferring compatibility transitively is the right instinct, and the two details that make it worth having are easy to miss: requiring each consumer to **rewrite** every producer output byte-identically, and the negative controls. The rewrite requirement is what catches signed zero, infinity and NaN payload bits, which numeric equality cannot prove and which are precisely where a checkpoint format quietly loses information. The negative controls are what stop the matrix from passing because it silently tested nothing.

Sequencing: this goes in behind #596. The one ask from that thread applies here — `eskm_v1_model_load_engine_parity` currently reports to CTest but not to the release oracle, so it cannot contribute to the readiness verdict. Template is in the #596 comment; happy for it to be a follow-up.

Gabriel-Kahen:
Resolved the post-#596 documentation conflict by merging master `99773a83` into this PR at `43923f47`. The resolution retains the complete matrix documentation and self-test descriptions while incorporating the merged corpus and loader changes. No parity script or Scheme fixture changes in this merge.

Verification: fixture integrity checker PASS (6 accepted / 6 rejected expectations checked offline); shell syntax PASS; fresh CMake configure PASS with both parity CTests registered; oracle schema PASS (347/347); public API docs and coverage inventory PASS; whitespace check PASS. GitHub now reports MERGEABLE.

The runtime model matrix and malformed-input tests were not rerun for this documentation conflict resolution; this update does not claim fresh runtime or CI acceptance. A separate follow-up aligns the public tensor matrix with #555's now-merged ESKM dispatch, leaving this model matrix's scope intact.

tsotchke:

PR #598, opened 2026-09-03
test(v1.2): harden forward-reference diagnostic runner

Gabriel-Kahen:
## Problem

The retained Bug W regression used fixed `/tmp/fwd_t*.esk` paths, so concurrent runs could overwrite one another and successful runs left fixtures behind. It also reported a missing compiler as a successful skip and gave limited evidence when its diagnostic oracle failed.

## Scope

This is a one-file hardening of `tests/v1_2_edge_cases/forward_ref_named_test.sh`. It does not change compiler or runtime behavior.

## Change

- create a unique per-run workspace under the parent test-isolation directory, durable evidence root, or configured temp root
- clean up normally while honoring durable/debug artifact-retention modes
- preserve cancellation exit semantics for HUP, INT, and TERM
- accept an explicit runner path and POSIX, macOS, or MSYS-style absolute and relative `BUILD_DIR` values
- classify missing runners and temp-directory failures as infrastructure errors
- capture stdout/stderr per case and report the expected diagnostic, actual exit status, and bounded stderr on failure
- require both the exact unresolved function name and the driver's clean diagnosed-rejection status (`1`), so a diagnostic followed by a crash cannot pass
- add deterministic negative controls proving the oracle rejects a missing function name, a false-success exit, and a signal-style exit

## Verification

Refreshed against `upstream/master` at `ea81a854` with LLVM 21.1.8:

PASS:

- `bash -n tests/v1_2_edge_cases/forward_ref_named_test.sh scripts/run_v1_2_edge_cases_tests.sh`
- standalone real-runner regression: 3/3
- `tests/v1_2_edge_cases/forward_ref_named_test.sh --self-test` (missing-name, exit-0, and exit-139 negative controls)
- two simultaneous real-runner invocations: 2 x 3/3, with no workspace collisions or leaks
- configured temp-root, durable-artifact placement, missing-runner, temp-setup failure, and MSYS `BUILD_DIR` resolution checks
- focused CTest selection: 3/3 (`forward_ref_scan_boundary_test`, `undefined_variable_refusal_is_clean_nocache`, `undefined_variable_refusal_is_clean_cached`)
- `python3 scripts/check_test_coverage.py --self-test`
- `python3 scripts/check_test_coverage.py` (46 suites)
- `git diff --check upstream/master...HEAD`
- independent review plus final full-diff review

The earlier full v1.2 edge-case run finished with 101 passed and four failures also reproduced on the pre-filing commit. That establishes that they were not caused by this PR; it does not classify them as an accepted baseline. The maintainer has since filed the first three as defects:

- `channels_test`: all assertions pass, then SIGSEGV during teardown
- `parallel_ad_worker_init_test`: external submission task-loss assertion
- `parallel_hash_table_mutation_test`: parallel insertion/key visibility assertions
- `repl_named_let_capture_jit_test`: JIT SIGSEGV

NOT RUN: the full repository-wide test matrix, macOS, or Windows.

## Out of scope

- compiler diagnostic wording or source-stack improvements
- fixes for the four unrelated pre-filing failures

### DIFF
--- a/tests/v1_2_edge_cases/forward_ref_named_test.sh
+++ b/tests/v1_2_edge_cases/forward_ref_named_test.sh
@@ -1,74 +1,144 @@
 #!/usr/bin/env bash
-# forward_ref_named_test.sh — Bug W regression (2026-04-25)
-#
-# Asserts that calling a forward-referenced function that was never
-# defined produces an error message that NAMES the function and
-# exits non-zero.
+# Bug W regression (2026-04-25).
+# Assert that unresolved forward calls exit nonzero and name the function.
+# Usage: forward_ref_named_test.sh [--self-test] [path/to/eshkol-run]
+
 set -u
 
-ROOT="$(cd "$(dirname "$0")/../.." && pwd)"
-RUN="$ROOT/${BUILD_DIR:-build}/eshkol-run"
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+SCRIPT="$ROOT/tests/v1_2_edge_cases/forward_ref_named_test.sh"
+SELF_TEST=0
+if [ "${1:-}" = "--self-test" ]; then SELF_TEST=1; shift; fi
 
-if [ ! -x "$RUN" ]; then
-    echo "SKIP: $RUN not built"
-    exit 0
+if [ -n "${1:-}" ]; then
+    RUN="$1"
+else
+    BUILD_PATH="${BUILD_DIR:-build}"
+    case "$BUILD_PATH" in
+        /*|[A-Za-z]:/*) RUN="$BUILD_PATH/eshkol-run" ;;
+        *) RUN="$ROOT/$BUILD_PATH/eshkol-run" ;;
+    esac
 fi
 
-mk() { cat > "$1"; }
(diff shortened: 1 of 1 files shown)


tsotchke:
Merged-in-spirit already: fixed `/tmp/fwd_t*.esk` paths are a real collision hazard, and a missing compiler reported as a successful skip is the worse of the two bugs — a test that cannot run should never look like a test that passed. Classifying that as infrastructure rather than success is the right call.

On the three failures you listed as unrelated baseline noise: the "not yours" part holds, and reproducing against a pre-filing run is what establishes it. The "baseline" part does not, and that is on us. None of the three — `channels_test` SIGSEGV during teardown, `parallel_ad_worker_init_test` task loss, `parallel_hash_table_mutation_test` key visibility — is recorded in `docs/KNOWN_ISSUES.md` or anywhere else. They have been failing as unattributed noise rather than as tracked defects. I have filed all three.

`channels_test` is the one I would look at first: every assertion passes and then it crashes on teardown, which is the classic shape of an object-lifetime bug on the shutdown path and exactly the kind of failure that gets absorbed because the assertions were green. Two of the three are in the parallel subsystem, which is also where a separate parity failure aborts inside glibc malloc, so they may be one cluster rather than three independents. Thank you for writing them down precisely enough to act on.

tsotchke:

PR #599, opened 2026-09-03
fix(vm): patch each letrec upvalue from the sibling slot it actually captured

tsotchke:
## Problem

The `engine_semantic_parity` readiness probe was red on master. Reduced:

```scheme
(letrec ((is-even? (lambda (x) (if (= x 0) #t (is-odd?  (- x 1)))))
         (is-odd?  (lambda (x) (if (= x 0) #f (is-even? (- x 1))))))
  (list (is-even? 6) (is-odd? 6)))
```

The native engine returns `(#t #f)`. The VM returned `(#t #t)`.

## Root cause

`letrec` creates each binding's closure before its siblings are assigned, so a closure that captures a not-yet-initialised sibling captures a placeholder by value. A second pass exists to re-snapshot those upvalues once every binding holds its final value, but it walked the closure's upvalue array and the binding list in lockstep: upvalue `j` was patched from binding `j`. That is only correct when a closure captures every sibling, in declaration order, starting from the first.

`is-even?` captures exactly one sibling, `is-odd?` at binding index 1, as its only upvalue at index 0. It was therefore patched from binding index 0, which is itself, and every call it made to what should have been `is-odd?` re-entered `is-even?`. Calling `is-even?` from the top still reached its base case either way, which is why only the second value of the pair ever looked wrong.

## Change

The compiler already knew, per upvalue, which enclosing slot it must read. That index is now threaded through explicitly. Each capture that lands inside an elaborating `letrec`'s own slot range is queued as a closure-slot, upvalue-index, sibling-slot triple, and the queue is emitted only after every binding has been stored, so a self-recursive binding's own patch cannot fire before its own store. The native call takes the index and the slot instead of reconstructing them positionally, and it keeps snapshotting by value rather than aliasing the frame, because a `letrec`-bound closure outlives the frame it was defined in.

Overflowing the patch queue now refuses the compile. A dropped entry is a silently miscompiled mutual recursion, which is the failure this pass exists to prevent, and it would appear only in the largest `letrec` forms, where it is hardest to see.

The prelude bytecode cache is regenerated because the emitted sequence changed.

## Test

`tests/vm/letrec_mutual_recursion_surface_regression.esk`, picked up automatically by the VM surface-regression glob, pins four shapes:

- the discovery shape at both parities and at zero,
- a three-way round robin in which no closure captures the sibling at its own binding index, so a position-based patch cannot get lucky,
- a single-binding self-recursive `letrec`, guarding the fix's timing,
- an ordinary nested closure that must keep its existing by-value capture.

## Evidence

Full `linux-x64` lane on a gate node, evidence in a comment below.


tsotchke:
## Lane evidence

`linux-x64` on a gate node (gcc-15.2.0, LLVM 21), full battery at `78beac66`:

```
252/253 tests passed
The following tests FAILED:
   94 - v1_3_jit_cache_runtime_smoke
Total Test time (real) = 2571.63 sec
```

The single failure is the pre-existing gcc-15 defect this branch does not touch: `jit_cache_test.sh` stage 1 asserted the object-ABI tag with a `strings`/`grep` heuristic over `.rodata`, and gcc-15 does not keep that literal contiguous. #593 replaces the heuristic with `eshkol-run --abi-fingerprint`, the same code path the run-cache key folds in. It reproduces on this node on master and is unrelated to `letrec`.

Every other suite is green, including the new `letrec_mutual_recursion_surface_regression` case and the whole VM surface-regression glob.

tsotchke:
## letrec* had the same defect, in a harder form

Probing `letrec*` after the `letrec` fix: R7RS `letrec*` initializes bindings left to right, but an initializer's lambda may still name a binding declared later — only the *call* has to happen after that binding is initialized. So a forward-referencing closure still captures the placeholder by value, and `compile_form_letrec_star` did no patching at all.

On the pushed branch head, VM vs native:

```
=== VM ===
letrec* even-odd 6 =>
ERROR: VM terminated on a fatal runtime error (pc=25563 sp=1165 fp=1162)

=== NATIVE ===
letrec* even-odd 6 => (#t #f)  (want (#t #f))
letrec* mod3 1 => b  (want b)
letrec* mod3 2 => c  (want c)
```

The VM called the NIL placeholder — a fatal error rather than `letrec`'s silent wrong value. `91b66f49` opens the same slot range and drains the same patch queue around `letrec*`'s initializer loop. Sequential reads of an already-stored earlier binding are unaffected: the drain only rewrites upvalues captured from this form's own slots, and it snapshots what is by then the final value either way. The prelude bytecode cache is byte-identical after regeneration (no prelude form forward-references), so this commit does not touch it.

## Evidence at 91b66f49 (same gate node)

- `letrec_mutual_recursion_surface_regression`: 19/19 checks, `Failed: 0` — the letrec cases plus `letrec*` counterparts of each, a `letrec*` sequential-initializer case pinning the semantics the patch must not disturb, and a self-recursive `letrec*`.
- `scripts/run_vm_surface_tests.sh`: `Passed: 59  Failed: 0`.
- `scripts/run_vm_parity.sh`: `228 passed, 3 failed, 0 infra`.

The three parity failures are pre-existing and unrelated. Rebuilding this node at the PR base `df47eba8` reproduces both crashes byte-for-byte:

```
BASE df47eba8 56_factor_graph_inference   rc=139   (SIGSEGV, vm-src)
BASE df47eba8 76_parallel_map_large_apply rc=134   (glibc malloc assertion, vm-src and vm-eskb)
```

Both are recorded as separate build items; neither involves `letrec`.

tsotchke:

PR #600, opened 2026-09-03
feat(model-io): publish checkpoint saves atomically

Gabriel-Kahen:
## Behavior

Tensor (ESKT) and model (ESKM v1) saves publish through a shared atomic-save helper in native and VM paths. Saving over an existing checkpoint leaves the old bytes in place until the complete replacement has been written, flushed, and closed successfully. A same-directory rename is the commit point; handled failures remove the temporary file and preserve the destination.

This guarantees process-level atomic replacement within one filesystem. It does not claim power-loss durability: neither the file nor its parent directory is fsynced. Existing wire bytes and capability checks are preserved. POSIX replacements preserve regular-file permission bits; new files use 0600. Destination symlinks are replaced without modifying their referents.

The PR includes focused failure-injection and four-engine tests, production-archive isolation checks, CI wiring, the documented contract, and required generated VM/site artifacts. Failure hooks exist only in the dedicated test executable.

## Revision and dependencies

- Head: `83467dde5fc6ac427436d457d821af094238a93b` (`gabe/atomic-checkpoint-saves`).
- PR base / merge base: `ea81a85448300858f92a33bbc676dc71f91ce06b` (`master`).
- Current fetched upstream master during this verification: `5ce74beeac25aca56c0c8129083e6db77f96f7dc`; the tests below ran at the exact PR head, without rebasing or adding unrelated compiler changes.
- No implementation dependency on #596, #597, or #602. The maintainer explicitly confirmed independence from #596/#597. Overlapping VM files with #602 were coordinated; this verification made no source edits.
- The existing implementation and maintainer feedback were reviewed again. No required atomic-publication gap was found, so no cosmetic commit was added.

## Exact-head verification

PASS on a fresh private Linux build using LLVM 21.1.8 and Clang 22, Release, with BLAS/GPU/agent FFI disabled and at most two compiler jobs:

```sh
ctest --test-dir .scratch/pr600-build --output-on-failure -j1 \
  -R '^(model_io_test|atomic_checkpoint_file_test|model_io_production_archive_test|generated_artifact_validator_test|runtime_core_boundary_test|atomic_checkpoint_save_test|vm_prelude_cache_is_current)$'
```

- **PASS:** 7/7 focused CTests; four-engine atomic-save matrix completed in 93 seconds. Exact ESKT/ESKM bytes, destination replacement, missing-parent and directory-destination failure, permissions, symlink referent preservation, concurrent writers, and temporary-file cleanup all passed.
- **PASS:** injected open/write/Nth-write/flush/close/rename failures and controlled interruption preserve the old checkpoint and clean up; deferred-signal cleanup passed.
- **PASS:** freshly compiled helper regression with both `mkostemp` and no-`mkostemp` fallback paths; the same existing helper regression also passed ASan+UBSan.
- **PASS:** capability-policy suite, 50/50.
- **PASS:** Emscripten 4.0.22 compilation of the atomic helper, plus the Emscripten-shaped host compile. Checked-in VM bundle Node smoke produced the expected sqrt and gradient results.
- **PASS:** generated-artifact validator (13 tests), VM prelude-cache check, production-archive and runtime-core boundaries, website verifier, shell syntax, Actionlint, and `git diff --check`.

The old Ubuntu-linked cached executable could not launch against the available Arch LLVM target set; it was not used as current runtime evidence. The fresh private build above resolved that environment mismatch.

**FAIL:** none in the completed current acceptance checks.

**NOT RUN:** native macOS/Windows execution, full repository test battery, fresh full WASM/site regeneration (sources unchanged), or the required GitHub CI matrix at this head.

## Remaining acceptance gap

GitHub reports this PR as OPEN and MERGEABLE, with merge state BLOCKED and an empty status-check rollup at the verified head. Required CI acceptance remains unverified. This PR has not been merged or declared accepted; native macOS/Windows coverage remains a platform gap.


tsotchke:
Answered the program as a whole on #596; the specific thing worth saying here is about the claim you did **not** make.

"This guarantees process-level atomic replacement; it does not claim power-loss durability" is exactly right, and it is the sentence most atomic-save patches get wrong. Write-to-unique-temp-in-destination-directory then `rename` gives replacement atomicity within the filesystem; it says nothing about whether the bytes survive a power cut without an `fsync` on the file and its parent directory. Scoping the guarantee to what the mechanism actually provides is worth more to a reader than a stronger sentence would be.

Keeping the temp file in the destination directory rather than a system temp dir is the other detail that matters — it keeps the rename within one filesystem, which is what makes it atomic at all.

Independent of #596/#597, so it can be gated whenever.

tsotchke:

PR #601, opened 2026-09-03
test(model-io): add deterministic ESKM fuzz gate

Gabriel-Kahen:
Depends on #555. This PR remains deliberately stacked on fix/issue-549-payload-validation so its review diff contains only the fuzz/resource and release-oracle work.

## Problem

The ESKM v1 loader had no deterministic, resource-bounded campaign that could detect crashes, hangs, hostile allocation declarations, or incorrect materialized bits. A CTest-only fuzz smoke would also be invisible to the release readiness verdict.

## Scope

- add a deterministic ESKM generator/mutator and native probe
- exercise headers, counts, dimensions, names, payloads, checksums, truncation, later-record corruption, scalar/empty/rank-32/large-name and multi-record cases
- compare accepted materialization against an independent FNV-1a oracle
- isolate every input in a bounded subprocess with timeout, address-space, artifact, and cleanup limits
- provide fixed-seed smoke/full modes and exact replay
- register the 70-case smoke in CTest
- emit provenance-bearing JSON-L release evidence only from the canonical bounded campaign
- add the high-severity v1.3.5-evolve completion-oracle criterion and ensure the readiness workflow builds and runs the dedicated probe

No loader, VM, public API, or wire-format implementation change is included beyond the #555 base.

## Fail-closed release evidence

Replay, reduced-count, relaxed-timeout, disabled-self-test, uncapped, and non-POSIX runs cannot write a release-satisfying PASS event. Events include timestamp, probe digest, and Git SHA. Missing, wrong, partial, stale, or deliberately failing evidence does not satisfy the criterion.

## Verification

PASS:

- LLVM 21.1.8 focused build
- model_io_test and eskm_model_fuzz_smoke: 2/2
- canonical 70-case smoke: zero failures
- deliberately failing campaign emitted FAIL
- replay did not overwrite release evidence
- harness negative controls
- completion-oracle schema: 346/346 criteria
- action bindings: 236 tests and 277 bindings
- false-green and evidence-staleness checks
- test inventory: 46 suites
- VM prelude and generated API checks
- Python, shell, workflow syntax, and git diff --check
- independent final review

NOT RUN:

- the external icc readiness binary, which is unavailable on this host
- macOS or Windows runtime execution
- full repository test battery
- dedicated TSan, MSan, or standalone LSan builds

## Out of scope

- changing ESKM v1 bytes or loader acceptance policy
- automatic delta debugging
- cross-engine parity, which remains in #597
- resolving or modifying #555

## Assignment verification update — 2026-09-07

Head reviewed: `c148e32243803a231ac2fb22700e19b3311516a9`; dependency #555 remains open. No source changes were needed for this verification update.

PASS (existing completed local logs, inspected after execution):

- Focused native `model_io_test` and ASan+UBSan `model_io_test`.
- Ordinary fixed-seed smoke: 70 cases, zero failures; full: 700 cases, zero failures.
- ASan+UBSan fixed-seed smoke: 70 cases, zero failures; full: 700 cases, zero failures. Leak detection was enabled for this run; this is not a standalone LSan build.
- All 13 harness self-checks in each completed smoke/full log.

These are local diagnostic results, not canonical release evidence. Ordinary runs used the existing 256 MiB address-space limit; sanitizer runs omitted that address-space cap for shadow-memory compatibility and retained the existing per-input timeout, arena and artifact budgets. Do not infer a sanitizer address-space-bound result.

The task then stopped with a platform security restriction. No further campaign execution was attempted after that restriction. Remaining verification was not completed; in particular, the earlier verification section above is historical and is not a claim that every generated/oracle check or the full repository battery was rerun in this update. Exact-source private build logs and six result logs are retained locally under `/tmp/eskm-601-verification-b29a/`. Maintainer review and #555 integration remain outstanding.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1192,6 +1192,17 @@ jobs:
           # consumes — without it architecture-verify false-FAILs.
           cmake --build build --parallel
 
+          # Keep the ESKM release oracle isolated from the main product build:
+          # configure fuzz support separately and build only its bounded probe.
+          cmake -S . -B build-fuzz -G Ninja \
+            -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+            -DESHKOL_REQUIRED_LLVM_MAJOR="${LLVM_MAJOR}" \
+            -DLLVM_CONFIG_EXECUTABLE="$LLVM_CONFIG" \
+            -DESHKOL_BUILD_TESTS=ON \
+            -DESHKOL_BUILD_AGENT_FFI=OFF \
+            -DESHKOL_ENABLE_FUZZ=ON
+          cmake --build build-fuzz --target eskm_model_fuzz_probe --parallel
+
       # B3 (2026-08-26 audit): scripts/run_language_coverage.sh hard-requires
       # a quantum-enabled runner (agent.quantum/agent.pqc are part of "100%
       # language coverage" — see run_language_coverage.sh's own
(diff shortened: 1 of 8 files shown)


tsotchke:
Answered the ESKM program on #596. Two things here specifically.

**The resource discipline is the part I want to call out.** An 8 MiB per-run artifact budget, retaining only the exact failing input plus JSON metadata, per-input subprocess isolation, a two-second timeout and a 256 MiB address-space cap — that is the whole checklist, and it was written into the harness rather than bolted on after an incident. We learned this the expensive way: a previous fuzz harness of ours filled a build node's disk because it had no cap and no cleanup. Pre-empting that is appreciated more than it might appear.

**Independent oracle.** Comparing materialized output against an FNV-1a oracle rather than against the loader's own view is what makes the campaign able to catch a wrong answer, not just a crash or a hang. A fuzzer that only detects crashes would pass a loader that cheerfully materializes incorrect bits.

Stacking on #555 rather than duplicating the loader hardening in the diff was the right call. Once #555 lands this wants the rebase and retarget to master you already flagged. Same oracle-wiring ask as #597: `eskm_model_fuzz_smoke` reports to CTest but is invisible to the release readiness verdict.

PR #602, opened 2026-09-03
fix(vm): reject malformed ESKM before materialization

Gabriel-Kahen:
## Problem and change

VM checkpoint loading could create strings, tensors, and list entries before discovering a rejected later record or trailing bytes. Preflight now validates every record and exact payload consumption before creating VM objects. Native and VM cursor readers use subtraction-based bounds checks. The final correction also refuses rank-zero and zero-element records during preflight because the existing VM constructor already rejects these shapes during materialization.

Scope: `lib/backend/vm_model_io.c`, `lib/core/model_io.cpp`, `tests/backend/vm_model_io_fail_closed_test.c`, `tests/core/model_io_test.cpp`, and focused test registration in `CMakeLists.txt`.

## Dependency and revisions

- Base: `fix/issue-549-payload-validation` at `e737c10a28c34099d502b0b788dba09c89d06875`.
- Head: `e588e1f10c2c4b02d85efd26f943b83e4c05f861`.
- Depends on maintainer PR #555, still OPEN at verification. Keep this PR stacked until #555 lands; retarget/rebase afterward. This PR does not take over #555's registration, format, or payload validation work.
- Related: #596 documents the v1 compatibility contract; #597 addresses the cross-engine fixture matrix.

## Compatibility and limits

ESKM wire version 1 (the v1.2 checkpoint contract) is unchanged. Duplicate names remain accepted in order with their original payloads; native and VM positive controls cover this. Native scalar/empty acceptance is unchanged. VM rank-zero scalars and zero-element tensors such as shape `[0]` remain unsupported: this PR moves their existing refusal before VM allocation, and does **not** establish scalar/empty native/VM acceptance parity.

Preflight uses one fixed-size record view and iteration bounded by validated file contents. It adds no global file-size limit, changes no public API, and does not promise rollback for allocator failure after successful validation. The whole-file temporary buffer still exists before preflight.

## Verification — Linux x86-64

Private copies of the clean prior-head build caches were used; affected VM translation units and executables were rebuilt/relinked against LLVM 21.1.8, with at most two compiler jobs. The native implementation and native positive test are unchanged from the verified prior-head cache. These are incremental builds, not a fresh full build. Tests ran in private working directories. `ESHKOL_PATH` was set to this checkout's `lib` and `LD_LIBRARY_PATH` to the cached LLVM21 library directory.

- PASS: `ctest --test-dir <normal> --output-on-failure -R '^(model_io_test|vm_model_io_fail_closed_test)$'`. Native round trips and both native/VM duplicate-name ordering/payload controls pass. VM rejection checks preserve heap slot and arena usage for trailing bytes, an incomplete later record, and unsupported scalar/empty later records.
- PASS: `ctest --test-dir <normal> --output-on-failure -R '^payload_validation_engine_parity$'` — native JIT, native AOT, and hosted VM bytecode, 91.71 seconds. The initial run without `ESHKOL_PATH` failed module discovery; the corrected environment passed.
- PASS: `ctest --test-dir <sanitize> --output-on-failure -R '^(model_io_test|vm_model_io_fail_closed_test)$'` with ASan + UBSan, `ASAN_OPTIONS=detect_leaks=1:halt_on_error=1`, and `UBSAN_OPTIONS=halt_on_error=1`; 2/2 pass with no sanitizer findings.
- PASS negative control: the two new bounded scalar/empty allocation-state assertions fail against the preceding `b691fe530a587a1e78e2e5fb32e02941f1587892` implementation and pass with the preflight correction. Test commit: `a715a4f3`; correction: `e588e1f1`.
- PASS: `python3 scripts/gen_api_docs.py --check`, `python3 scripts/check_test_coverage.py` (46 suites), and `git diff --check`.
- FAIL, inherited from #555: `vm_prelude_cache_is_current`. The generator emits 5329 instructions / 768 locals / 945 constants, while the committed header records 5340 / 770 / 947. A separately compiled generator using exact base `e737c10a` sources produces byte-identical output to this head. No generated source was changed in this PR.
- NOT RUN: `vm_canonical_stdlib` (no matching test registered at this stacked base), full suite, full sanitized compiler/engine matrix, macOS, Windows, ARM, WASM, GPU/XLA, and resource/fuzz campaigns.

## Remaining acceptance gaps

#555 must land and the inherited prelude-cache failure must be resolved before this stack can satisfy the complete acceptance gate. Native/VM scalar and empty-tensor parity remains a separate compatibility gap. This PR is filed and locally verified as stated above; it is not merged or accepted. No releases, infrastructure, unrelated compiler changes, or generated artifacts are included.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3078,6 +3078,27 @@ if(ESHKOL_BUILD_TESTS AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/tests/core/model_i
 endif()
 
 if(ESHKOL_BUILD_TESTS AND NOT WIN32 AND
+   EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/tests/backend/vm_model_io_fail_closed_test.c")
+    add_executable(vm_model_io_fail_closed_test
+        tests/backend/vm_model_io_fail_closed_test.c)
+    target_include_directories(vm_model_io_fail_closed_test PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/inc
+        ${CMAKE_CURRENT_BINARY_DIR}/generated)
+    eshkol_apply_vm_limit_definitions(vm_model_io_fail_closed_test)
+    if(LLVM_LIBRARY_DIRS)
+        target_link_directories(vm_model_io_fail_closed_test PRIVATE ${LLVM_LIBRARY_DIRS})
+    endif()
+    target_link_libraries(vm_model_io_fail_closed_test PRIVATE
+        eshkol-static
+        ${ESHKOL_EXTRA_LINK_LIBS}
+        ${LLVM_LIBS_LIST}
+        ${LLVM_SYSTEM_LIBS_LIST})
+    add_test(NAME vm_model_io_fail_closed_test COMMAND vm_model_io_fail_closed_test)
+    set_tests_properties(vm_model_io_fail_closed_test PROPERTIES
+        PASS_REGULAR_EXPRESSION "PASS: VM ESKM fail-closed preflight")
+endif()
+
+if(ESHKOL_BUILD_TESTS AND NOT WIN32 AND
    EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/tests/toolchain/payload_validation_test.sh" AND
    EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/tests/v1_3_edge_cases/payload_validation_test.esk")
     add_test(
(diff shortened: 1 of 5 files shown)


tsotchke:
Answered the program on #596; the two specifics here.

Preflighting every record **before** creating VM strings, tensors or list entries is the correct shape for a fail-closed loader. Validating as you materialize means a malformed later record leaves partial allocations behind, and in a VM that is not merely a leak — it is reachable, partially-initialized state, which is worse than the rejection it was trying to perform.

Subtraction-based bounds checks instead of offset arithmetic is the right fix and not just a style preference: `cursor + len > end` is the formulation that overflows on a hostile declared length and quietly passes, which for a format whose lengths come from the file is the whole attack surface. `end - cursor < len` cannot.

Stacked on #555 as intended; rebase and retarget to master once that lands.

PR #603, opened 2026-09-03
fix(vm,native): close the P8 axis-3 arity divergences and the axis-6 surface disagreements

tsotchke:
## Problem

Readiness on the gated ref reported two P8 escape-matrix probes red on master, both pre-existing:

- **axis-3 (`arity_sweep_native_vm_parity`)** — wrong-arity calls refused by one engine and silently mis-executed by the other.
- **axis-6 (`five_way_surface`)** — nine surface disagreements between the documentation, the two engines and the module system.

## Change

**Native.** The `exact->inexact`/`inexact` and `numerator`/`denominator` handlers indexed `call_op.variables[0]` with no argument-count check, so a zero-argument call read an unused slot and returned a bogus tagged value instead of failing the build. Every sibling math intrinsic has refused this since ESH-0362; these two were missed. They now raise the same arity error.

**VM.** `emit_builtin_preamble()` binds each `BUILTINS[]` entry to a closure whose body is exactly `def->arity` local loads, but the generic call path never compared a call's argument count against that arity, so `(ceiling)` compiled clean and read an uninitialised local at run time, printing `0`.

The new check keys on the **binding**, not the name. Many `BUILTINS[]` entries are internal fixed-arity primitives that the Scheme prelude immediately rebinds under the same public name at a different, often variadic, arity: the raw one-argument `error` op backs the public `(error msg irritant...)`. A prelude rebinding adds a later top-level local, so a call resolving to it is not a call to the raw op and carries none of its arity. Because the preamble runs first on a fresh chunk, builtin `b` occupies top-level local index `b`, so a resolved index below the builtin count whose name still matches is the preamble's own binding and nothing else. Resolving the head the way the call itself will therefore makes the check exact for **every** builtin, with no name allowlist: a prelude wrapper, a user redefinition and a lambda parameter each resolve elsewhere. Prelude compilation is excluded, because the prelude reaches these names through its own wrappers while it is still being compiled. The check refuses only calls with **too few** arguments — the memory-unsafe direction, and the one the sweep probes.

## The minimum is a third thing from the operand count

`BUILTINS[].arity` is how many operands `emit_builtin_preamble()` loads into the opcode — the shape of the call, not the caller's obligation. Reading it directly as the minimum was wrong, and it failed loudly twice.

First on `hash-ref`, which shares opcode 661 with the three-argument `hash-table-ref/default`, whose third operand is the value to return on a miss: `(hash-ref table key)` legitimately omits it. A full lane refused every two-argument `hash-ref` in the corpus and took 13 VM source tests down.

Then on `gcd` and `lcm`, found by the **full** axis-3 sweep. `codegenGCD`/`codegenLCM` answer a zero-argument call with the R7RS identity (`0` and `1`) and then fold over `num_vars` — native is genuinely variadic, so `(gcd 3)` is a documented-legal call that this branch had started refusing.

`BuiltinDef` therefore carries `min_arity` in three states: `0` is unset (the minimum is `arity`), a positive value is the real minimum, and a **negative** value declares the builtin variadic. `vm_builtin_arity_at_index` returns `-1` for that case, which is the channel the caller already uses for "not a preamble binding, make no claim", so no call site changes.

That claim is **checked, not pinned**. `DECLARED_VARIADIC` in `scripts/check_builtin_min_arity.py` maps each name to the `llvm_codegen.cpp` handler that makes it true, and the gate confirms the handler really has a `num_vars == 0` case. It discriminates: `codegenGCD` and `codegenLCM` pass, `codegenRound` does not, a missing handler is an error. Wired into ctest as `builtin_min_arity_surface_agreement`.

## Where the VM was right and native was not

Six probes diverged because the VM began refusing a zero-argument call that native still accepted, returning `0`. Two shapes, both fail-open:

- `inexact->exact`/`exact` and `bytevector-length` read `variables[0]` with no check at all — the same class closed for `exact->inexact`/`inexact` and `numerator`/`denominator`, and these are the siblings that pass missed.
- `round` and `hashCount` (which also serves `hash-table-size` and `hash-table/count`) do have the guard, but raise it with `eshkol_warn()`, which leaves the build succeeding and the swallowed `nullptr` surfacing as `0`.

All four now use `eshkol_arity_error_current()`. No handler's accepted arities change; only whether a wrong one is fatal. Both engines now refuse these calls, which is a better agreement point than both silently answering `0`.

## A defect this branch introduced in its own tooling

`scripts/gen_language_surface.py` matched `BUILTINS[]` rows with a three-field-only pattern, so the four-field `{"hash-ref", 661, 3, 2}` row matched nothing and `hash-ref` silently dropped out of the VM table — the generated surface claimed it had no VM backend at all (`builtins_in_vm_table` 727 → 726). `check_builtin_min_arity.py` had the mirror flaw: its `min_arity` group was `(\d+)`, so a negative row would have vanished from the gate's own view. A row these parsers cannot read is dropped in **silence**, so both patterns now track `BuiltinDef`. With the fix the manifest regenerates byte-identical to the committed one, which also shows it was not otherwise stale.

## Axis-6

Four of the nine were the checker misreading the source: the driver's `define` pattern only accepted a lowercase first letter and never matched `define-syntax`, so the uppercase terminal constants in `lib/agent/terminal.esk` and the `expose-hidden` macro in `lib/test/modules/dd10_privacy_scope.esk` were reported as provide orphans. Widening the pattern can only shrink the disagreement set. The other five follow commit `f1a85c64`, which renamed the VM-only manifold operations to a `-handle` family so they stop shadowing the public API in `lib/core/manifold.esk`, and `PARITY.tsv` row 201, which already accepts `dataloader-has-next?` as a VM-fidelity gap. The baseline is corrected in place: six keys that no longer name anything are removed and their five renamed successors added. The accepted scope is unchanged; only the spelling is.

## Evidence

Mesh gate, `linux-x64` lane on a dedicated gate node (LLVM 21, Release), pasted in a comment below.

Every before/after number was measured on real binaries, not read off the source: the "before" side is a build carrying none of these commits. The eight arity divergences the full sweep reported all agree afterwards.

**`--quick` is not evidence about the axes it did not sample.** The sampled default tests 30 of the 204 candidate builtins and reported 8/8 axes PASS on the very commit whose full sweep found ten divergences.

## Known gaps, not addressed here

- **`string-contains` means three different things**, and it is pre-existing — reproduced on a build with none of these commits. Native returns a boolean (the name reaches the `string-contains?` lowering), the VM returns an index or `-1`, and `eshkol_compiler.c:4690` documents a third, SRFI-13 contract ("index or `#f`"). Converging them changes the value a public builtin returns, so it needs a ruling rather than a drive-by fix. Filed as a build item.
- A wrong-arity call to a **user-defined** VM closure is still accepted where native refuses it. Closing that needs real per-binding arity tracking in the VM compiler, not a name lookup. Filed as a build item.


tsotchke:
## Evidence

### Full axis-3 sweep — the measurement that found this

The CI default is `--quick`, which samples **30 of the 204** candidate builtins. It reported 8/8 axes PASS on the commit whose **full** sweep reported ten divergences. Quick-mode green is not evidence about the axes it did not sample.

Full sweep on this branch, before the baseline was touched:

```
axis-3 arity sweep: tested=204 builtins, divergences=220 (baseline=366), NEW=2
NEW native-vs-VM divergences (not in baseline):
    string-contains::correct
    string-contains::wtype
```

All eight divergences this branch had opened are gone; the only remaining new keys are the two proven pre-existing ones, so nothing else had drifted in the two weeks since the baseline was last regenerated.

After recording those two:

```
axis-3 arity sweep: tested=204 builtins, divergences=220 (baseline=368), NEW=0
axis-3 gate: PASS
```

Axis-6 at the same commit:

```
axis-6 five-way surface: 589 disagreements (baseline=638), NEW=0, resolved=49
axis-6 gate: PASS
```

Both re-measured on the gate node's own Linux build at the tip (`518b0b08`), against the same baselines:

```
axis-3 arity sweep: tested=204 builtins, divergences=221 (baseline=368), NEW=0
axis-3 gate: PASS
axis-6 five-way surface: 589 disagreements (baseline=638), NEW=0, resolved=49
axis-6 gate: PASS
```

The one-key difference in the axis-3 total (221 on Linux, 220 on macOS) is platform-specific and inside the baseline on both; `NEW=0` on both.

### Per-probe, before and after

"Before" is a build carrying **none** of these commits, so these are measurements, not readings off the source.

| probe | before | after |
|---|---|---|
| `gcd::warity` | agree `3`/`3` | agree `3`/`3` — refusal of legal code gone |
| `lcm::warity` | agree `3`/`3` | agree `3`/`3` |
| `round::warity` | agree `0`/`0` (both fail open) | agree REJECTED/REJECTED |
| `inexact->exact::warity` | agree `0`/`0` | agree REJECTED/REJECTED |
| `bytevector-length::warity` | agree `0`/`0` | agree REJECTED/REJECTED |
| `hash-count::warity` | agree `0`/`0` | agree REJECTED/REJECTED |
| `hash-table-size::warity` | agree `0`/`0` | agree REJECTED/REJECTED |
| `hash-table/count::warity` | agree `0`/`0` | agree REJECTED/REJECTED |
| `string-contains::warity` | **DIVERGE** | agree REJECTED/REJECTED |

Six of these now agree by **both engines refusing** rather than both silently answering `0`, which is a better agreement point than the one the baseline recorded. `string-contains::warity` was a divergence before this branch and is fixed by it.

### The surface manifest

With the generator's row pattern corrected, `tests/coverage/language_surface.json` regenerates **byte-identical** to the committed file (`builtins: 1043, vm 727`). Before the fix it regenerated with `vm 726` and `hash-ref` missing its `vm` backend.

### Mesh gate

`linux-x64` lane on a dedicated gate node, LLVM 21, Release, at `85c06699`:

```
lane=linux-x64  status=fail  stage=done  rc=8
uname=Linux/x86_64  duration_seconds=2679
detail=lane battery exited 8
```

```
94/254 Test #94: v1_3_jit_cache_runtime_smoke ...***Failed  Regex=[FAIL:]  0.23 sec
99% tests passed, 1 tests failed out of 254
```

Build had **zero compile errors**. `run_all_tests.sh` reported no suite failures. The single ctest failure is `v1_3_jit_cache_runtime_smoke` — the gcc-15 `strings` heuristic that #593 fixes, which fails identically on the parent branch and on the previous head of this one.

The lane ran at `85c06699` rather than the tip. The only later commit changes `tests/escape_matrix/arity_parity_baseline.json`, and no ctest consumes that file, so the battery result carries unchanged; the axis-3 and axis-6 numbers below were taken at the tip.

tsotchke:

PR #605, opened 2026-09-04
fix(gpu): recognize current CUDA dispatch logs

### LINKED ISSUE #604: GPU correctness gate skips current cublasDgemm dispatch logs
## Problem

On current `master`, the real-device gate in `tests/gpu/gpu_correctness_gate.sh` can return `SKIP` after CUDA has actually executed the workload.

The gate recognizes the older `-> CUDA cuBLAS` wording, but the current CUDA selector emits a concrete operation:

```
[GPU] matmul 512x512 @ 512x512 -> cublasDgemm (cublasDgemm (default))
```

Because that line does not match the evidence regex, the gate exits before the GPU/CPU differential and writes no PASS evidence.

## Reproduction

Environment:

- NVIDIA GB10, compute capability 12.1
- CUDA 13.0
- LLVM 21.1.8
- Linux ARM64

With GPU and CPU builds supplied through `REUSE_BUILDS=1`, the gate reports:

```
Running GPU-enabled binary with dispatch forced...
SKIP: GPU-enabled binary built and ran, but no GPU device announced itself at runtime
```

Running the generated payload directly shows the `cublasDgemm` dispatch above. Its ten RESULT probes agree with the CPU build; the largest observed relative difference is `8.843665037e-16`, below the gate tolerance of `1e-9`.

## Expected behavior

The current `cublasDgemm` and optional `INT8-Ozaki` selector messages should count as CUDA execution evidence, while a `-> CPU` fallback must not.

## Proposed regression coverage

Centralize the accepted evidence expression and extend `--self-test` to cover:

- Metal initialization;
- legacy CUDA wording;
- current `cublasDgemm`;
- CUDA INT8-Ozaki;
- sf64 dispatch;
- explicit rejection of CPU fallback.

RichardHoekstra:
## What

Fixes #604.

The real-GPU correctness gate recognized the older `-> CUDA cuBLAS` log wording but not the selector messages emitted by the current CUDA backend. On a live CUDA device, the workload therefore executed through `cublasDgemm` and then the gate returned `SKIP` before comparing its output with the CPU reference.

This centralizes the runtime-evidence matcher and accepts the current concrete CUDA operations:

- `cublasDgemm`;
- `INT8-Ozaki`;
- the existing Metal, legacy CUDA, and sf64 evidence forms.

A `-> CPU` fallback remains explicitly rejected.

## Regression coverage

`gpu_correctness_gate.sh --self-test` now pins all five accepted evidence families and a negative CPU-fallback case. The self-test and live-device path call the same matcher, so their accepted formats cannot drift independently.

## Real-device verification

Verified on NVIDIA GB10 (compute capability 12.1), CUDA 13.0, LLVM 21.1.8, Linux ARM64.

Before:

```
Running GPU-enabled binary with dispatch forced...
SKIP: GPU-enabled binary built and ran, but no GPU device announced itself at runtime
```

After:

```
GPU device confirmed live:
  [GPU] matmul 512x512 @ 512x512 -> cublasDgemm (cublasDgemm (default))
...
Max relative diff across all probes: 8.843665037e-16 (tolerance 1e-9)
PASS: GPU execution matches CPU reference within tolerance (1e-9) on Linux
```

All ten RESULT probes were compared. Eight were bit-identical; the remaining two pairs differed by at most `8.843665037e-16`.

## Validation

- [x] `bash -n tests/gpu/gpu_correctness_gate.sh`
- [x] `tests/gpu/gpu_correctness_gate.sh --self-test`
- [x] Real CUDA-vs-CPU gate with reused clean Release builds
- [x] `ctest -R 'verify_gpu_backend_test|cuda_runtime_link_args_test'`
- [x] No runtime/compiler code or public API changed

`shellcheck` reports only the pre-existing SC2164 warning at the repository-root `cd`; this change introduces no new shellcheck finding.

## Review scope

One test/gate script only. The production GPU dispatch and numerical tolerance are unchanged; this fixes evidence recognition so an already-executed CUDA workload reaches the existing differential comparison.

### DIFF
--- a/tests/gpu/gpu_correctness_gate.sh
+++ b/tests/gpu/gpu_correctness_gate.sh
@@ -97,6 +97,19 @@ log()  { printf '%s\n' "$*"; }
 skip() { log "SKIP: $*"; exit 0; }
 fail() { log "FAIL: $*"; emit_trace FAIL "$*"; exit 1; }
 
+# Runtime evidence must name a live backend signal, not merely prove that a GPU
+# library was compiled in. Keep the accepted spellings in one place so the
+# self-test and the real-device gate cannot drift apart as backend logs evolve.
+GPU_DISPATCH_LOG_RE='\[GPU\] (Metal: |.*-> CUDA cuBLAS|'
+GPU_DISPATCH_LOG_RE+='matmul [0-9]+x[0-9]+ @ [0-9]+x[0-9]+ -> (cublasDgemm|INT8-Ozaki)'
+GPU_DISPATCH_LOG_RE+='|sf64 (dispatch|completed):)'
+gpu_dispatch_log_present() {
+    grep -qE "$GPU_DISPATCH_LOG_RE" "$1"
+}
+gpu_dispatch_log_lines() {
+    grep -E "$GPU_DISPATCH_LOG_RE" "$1"
+}
+
 # ─────────────────────────────────────────────────────────────────
 # --self-test: proves the SKIP/PASS/FAIL-vs-trace contract directly,
 # independent of whether this host actually has a GPU. `skip()` is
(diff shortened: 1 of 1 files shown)


tsotchke:
Thanks, this is correct and the self-test change is the important part: the gate now rejects the CPU-fallback line explicitly rather than only accepting the old wording, so a future selector rename fails loudly instead of skipping. Reviewed the regex against the current selector output; it matches cublasDgemm and INT8-Ozaki and still the legacy wording. This will go through the mesh gate on our CUDA nodes (CUDA 12) and merge with #607; your GB10 run is the CUDA 13 evidence, so if you can paste the gate's PASS line from that box on the PR it becomes part of the record.

tsotchke:

PR #607, opened 2026-09-04
fix(cuda): exclude SM72 from CUDA 13 defaults

### LINKED ISSUE #606: CUDA 13 default GPU configure fails on removed compute_72 architecture
## Environment

- Eshkol master at `a619b41c6e0cb59fa2002534c20e7ccdddfed155`
- Ubuntu 24.04, Linux ARM64
- CMake 3.28.3
- CUDA 13.0.88 (nvcc)
- NVIDIA GB10 (SM 12.1)

## Reproduction

```sh
cmake -S . -B build -G Ninja \
  -DCMAKE_BUILD_TYPE=Release \
  -DESHKOL_GPU_ENABLED=ON \
  -DESHKOL_REQUIRE_GPU_BACKEND=ON
```

Configuration fails during CMake's CUDA compiler check:

```text
nvcc fatal: Unsupported gpu architecture 'compute_72'
```

The generated probe command includes `compute_72` through `compute_90`, sourced from the default `ESHKOL_CUDA_ARCHITECTURES` value.

## Root cause

CUDA 13 removed offline compilation support for architectures below compute capability 7.5, while Eshkol's portable default still starts at SM72. The architecture list is installed into `CMAKE_CUDA_ARCHITECTURES` before `enable_language(CUDA)`, so even compiler identification cannot run.

The build-graph verifier also unconditionally requires SM72, which would reject a CUDA-13-compatible default even after configuration is repaired.

## Expected behavior

The default architecture policy should be toolkit-aware:

- preserve SM72 for older CUDA toolkits, including the verified Jetson Xavier/CUDA 11.4 path from #68;
- omit SM72 when CUDA 13 or newer is selected;
- keep explicit user-provided `CMAKE_CUDA_ARCHITECTURES` values authoritative;
- make the build-graph verifier enforce the appropriate portable floor for the detected toolkit.

An explicit `-DCMAKE_CUDA_ARCHITECTURES=121` configures and builds successfully on this machine, confirming the backend itself is healthy.

RichardHoekstra:
## Summary

- keep the existing portable CUDA architecture defaults for toolkits through CUDA 12
- omit SM72 when CUDA 13 or newer is detected, because offline compilation below compute capability 7.5 was removed
- expose the detected toolkit major to the backend verifier
- cover both accepted CUDA 13 defaults and rejected SM72 configurations

Fixes #606

## Verification

```sh
python3 -m unittest discover -s tests -p 'test_verify_gpu_backend.py'
```

All 6 focused verifier tests pass. The change was also validated with CUDA 13 on an SM121 GB10 GPU: configuration, build, GPU smoke test, and backend verification all passed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1566,10 +1566,19 @@ if(ESHKOL_GPU_ENABLED)
         # Try CUDA on Linux/Windows
         find_package(CUDAToolkit QUIET)
         if(CUDAToolkit_FOUND)
-            set(ESHKOL_HOST_CUDA_MAJOR "${CUDAToolkit_VERSION_MAJOR}")
+            set(ESHKOL_HOST_CUDA_MAJOR "${CUDAToolkit_VERSION_MAJOR}"
+                CACHE INTERNAL "CUDA toolkit major used by this build" FORCE)
             if(NOT DEFINED CMAKE_CUDA_ARCHITECTURES OR
                "${CMAKE_CUDA_ARCHITECTURES}" STREQUAL "")
-                set(CMAKE_CUDA_ARCHITECTURES "${ESHKOL_CUDA_ARCHITECTURES}"
+                set(_eshkol_cuda_architectures "${ESHKOL_CUDA_ARCHITECTURES}")
+                # CUDA 13 removed offline compilation for architectures below
+                # compute capability 7.5.  Keep SM72 in the portable default
+                # for older toolkits (notably the verified Jetson Xavier path),
+                # but do not pass it to a compiler that no longer accepts it.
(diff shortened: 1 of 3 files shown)


tsotchke:
Thanks. Reviewed: the version gate is the right place (CMake toolkit version, mirrored into the verifier through ESHKOL_HOST_CUDA_MAJOR), the CUDA 12 default list is untouched, and the tests cover both the accepted CUDA 13 graph and the rejected SM72 case. Our mesh has CUDA 12 only, so non-regression is what we can prove here; CUDA 13 configure success on your machine is the other half, and the configure output line listing the resolved architectures would be worth attaching. Queued with #605 for the mesh gate and merge.

tsotchke:

PR #609, opened 2026-09-06
fix: v1.3.5 corrections from the XLA program (LLVM 18-24 compat, broadcast, Metal f64, Adam pool)

tsotchke:
Corrections found while bringing the XLA backend up against StableHLO's pinned LLVM and running the tensor paths on new hardware. Every item ships with a test that failed before and passes after.

Compiler and build
- LLVM header overlay no longer discards include roots; vendored PJRT C ABI header tracked and on the XLA include path
- Compile against LLVM 18 through 24 through the version-guarded compatibility layer: intrinsic signature check spelled per version, single-operand loop hints for LLVM 23+, and BasicBlock terminator queries through terminatorOrNull (LLVM 24 changed getTerminator to assert; every block-closed guard fired on empty entry blocks)
- Driver and JIT link path use lld on AArch64 Linux only when ld.lld is on the PATH
- WASM import glue declares eshkol_broadcast_shape_f64
- Diagnostics: intrinsic mismatch reason on stderr; ESHKOL_DUMP_IR_ON_VERIFY_FAIL

Runtime and tensors
- Broadcast allocates the true result size (heap overflow) and reports incompatible shapes
- XLA elementwise path broadcasts instead of reading past the smaller operand; call site passes all ten arguments (the dispatch had been silently dead)
- Metal soft-float: sf64_div quotient placement and sf64_from_int shift (all f64 divides and axis means on Metal returned wrong values); regression tests for both
- VM implicit Riemannian-Adam state pool is growable; the seventeenth distinct shape no longer overwrites slot 0

Local gates: memory 25/25, autodiff 61/61, stdlib 22/22, VM surface 116/116, GPU 20/20 on Metal, XLA elementwise 11/11.


tsotchke:

PR #613, opened 2026-09-08
docs: propose ESKM v2 metadata envelope (GK-SER-05)

Gabriel-Kahen:
## Problem and proposed decision

GK-SER-05 needs a reviewed byte-level evolution decision before public format implementation. ESKM v1 has no safely extensible framing: historical readers ignored its reserved word. This design-only PR proposes wire version 2 with a 24-byte header, bounded optional TLVs, advisory bytewise annotations, a separate required-feature bitmap (initial supported mask zero), and the unchanged v1 records/CRC algorithm. Existing ESKM writers remain v1; native language ESKT dispatch remains unchanged. No v2 reader, writer, public API, fixture generator, or runtime behavior is implemented here.

**Status: Proposed.** Please record a byte-level decision on the header/CRC, optional-versus-required semantics, annotation grammar and permitted metadata loss, proposed admission caps, compatibility boundary, and implementation gates. Acceptance requires a maintainer review linked to the accepted document commit; filing or merging this Proposed document alone does not authorize implementation or complete GK-SER-05. Public opt-in API signatures/native IDs need their own review before the writer slice.

## Scope and provenance

Only `docs/design/ESKM_V2_FORMAT_DECISION.md` and its navigation link in `docs/README.md` change. The existing local draft and navigation addition were reproduced in an isolated worktree and technically reviewed; both original user files remain byte-identical. The general assistant-goals policy is not this assignment. No release, generated, runner, compiler, or unrelated files change.

- Base: `master` at `5ce74beeac25aca56c0c8129083e6db77f96f7dc`.
- Head: `2d85dc648d3088ebb75dcfd35f076c9efd7daedc`.
- V1 specification/corpus dependency: #596 at `edb614e57904229fc626b23801c9bffbbb248b25`, linked with immutable commit URLs. A clean master base avoids duplicating that unmerged PR's corpus diff.
- Reader prerequisite: GK-SER-02 / #602 after #555, including all-record preflight, nonzero v1 reserved-word refusal, and documented backend resource accounting.
- Writer prerequisite: GK-SER-04 / #600 atomic checkpoint publication.
- Assignment: GK-SER-05 from #587's proposed ownership roadmap; its roadmap file is not on this base and is not copied into this focused PR.

## Technical review

- Independently verified the exact 28-byte empty container: the preceding 24 bytes have CRC `0xc3e87005`; footer is `05 70 e8 c3`.
[...]

PR #614, opened 2026-09-08
docs: propose ESKM v1 VM scalar and empty materialization

Gabriel-Kahen:
## Problem

The historical ESKM v1 `scalar.eskm` (rank 0) and `empty-0x3.eskm` (shape `[0, 3]`) fixtures are valid and native-materializable, but the VM serialization path cannot construct them. #597 explicitly excludes this gap; #602 deliberately moves the existing VM refusal before persistent allocation. This proposal describes how a later reviewed I/O-local adapter could preserve their metadata and exact bytes across native JIT, native AOT, VM source, and VM bytecode. It does not complete GK-SER-01 or GK-SER-03 acceptance.

## Scope and decision requested

**Status: Proposed — maintainer review required before implementation.**

Only `docs/design/ESKM_V1_VM_MATERIALIZATION.md` changes. The explicit review decision is whether to approve or revise D1–D5: rank-0 inline dimension/stride representation, non-null arena storage for empty tensors, checked empty strides, existing object boxing/region lifetime, and the narrow I/O observation surface. Approval of this design would permit planning the separate implementation slice; this PR contains no implementation.

The adapter would remain private to VM ESKM I/O and preserve `VmTensor` layout, general constructors, public signatures/native IDs, and default v1 bytes. The document calls out that the resulting values can reach other tensor operations and requires further review if integration needs general tensor or memory-rule changes.

## Evidence and dependencies

- Source-read baseline: current `upstream/master`, `5ce74beeac25aca56c0c8129083e6db77f96f7dc`.
- Read the actual roadmap on `upstream/docs/gabe-kahan-roadmap` at `10301403f6ae434321108751bc46a0c3bad7ca62` and `CONTRIBUTING.md`.
- Read #596 fixture metadata at `5ad058b2bc416c97a76c21f246d6b3bb4044b48f` and #597's documented exclusions.
- Read #602's preflight refusal at `e588e1f10c2c4b02d85efd26f943b83e4c05f861`; it depends on #555 at `e737c10a28c34099d502b0b788dba09c89d06875`. Both were open when inspected. The implementation must follow review and coordinate with their landed behavior, adjusting preflight and materialization together while preserving the other admission guarantees.
- This is an existing-v1 VM representation decision, independent of #613's proposed v2 wire-format decision.

## Follow-on acceptance

Keep all six valid #596 fixtures unchanged. Every engine must inspect metadata and rewrite each fixture byte-for-byte; every rewrite must then pass all 16 producer/consumer combinations per fixture. Add VM ESKM single-tensor-route coverage and bounded positive lifetime checks after function/region return. New scalar/empty acceptance assertions must fail against pre-adapter behavior. The document also requires an expected-metadata/payload negative control for the future test logic.

## Verification

PASS (Linux x86-64, documentation-only):

- Pinned source and fixture metadata inspection; independent read-only design review found no actionable inaccuracies or scope issues.
- `python3 scripts/gen_api_docs.py --check` — generated API docs up to date; no generated changes.
- `python3 scripts/check_test_coverage.py` — 46 suites.
- `git diff --cached --check` and seven relative documentation/source links resolve.

NOT RUN:

- Runtime builds, four-engine executions, the future acceptance gate, full suite, and other platforms. No ESKM producer/consumer engines were exercised in this design-only PR.
- Negative runtime control: not applicable to this PR because it changes no test logic; the future control is specified, not claimed as passing.

## Compatibility and exclusions

No current product behavior, wire-format byte, compatibility/resource-limit policy, loader implementation, public tensor rule, or object ABI changes. No malformed-input/resource campaigns, generated artifacts, `.icc/` or release evidence, roadmap status changes, or unrelated cleanup. Scalar/empty acceptance remains unmet until a reviewed implementation passes the complete gate.

### DIFF
--- a/docs/design/ESKM_V1_VM_MATERIALIZATION.md
+++ b/docs/design/ESKM_V1_VM_MATERIALIZATION.md
@@ -1,0 +1,177 @@
+# ESKM v1 scalar and empty-tensor VM materialization
+
+- **Status:** Proposed; requires maintainer review before implementation.
+- **Decision owners:** ESKM maintainers and VM memory/runtime maintainers.
+- **Scope:** A serialization-local adapter supporting the remaining scalar/empty
+  compatibility gate in GK-SER-01 and GK-SER-03.
+- **Implementation and acceptance:** Not started; this document supplies no
+  runtime acceptance evidence and does not complete either work packet.
+
+## Problem and intended behavior
+
+ESKM wire version 1, used by the v1.2 checkpoints, can represent a scalar with
+rank 0 and one binary64 element, and an empty tensor with shape `[0, 3]` and no
+elements. The historical valid fixtures in
+[PR #596](https://github.com/tsotchke/eshkol/pull/596) include both. Native model
+loading can materialize them; current VM model loading cannot.
+[PR #597](https://github.com/tsotchke/eshkol/pull/597) explicitly excludes this
+gap from its cross-engine gate.
+
+After a reviewed implementation, loading either fixture through ESKM model I/O
+would produce a tensor with its original rank, dimensions, dtype, and payload.
+Native JIT, native AOT, VM source, and VM bytecode would inspect and rewrite that
+tensor without changing the checkpoint bytes. Rank 0 must remain rank 0; an
+empty tensor must remain empty. This is an I/O compatibility promise, not a new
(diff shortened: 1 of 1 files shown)


tsotchke:
Retriggering required repository checks for the merge train.

PR #615, opened 2026-09-08
test(io): add ESKT tensor-file four-engine cross-reader matrix

Gabriel-Kahen:
The public ESKT tensor-file path lacked a literal cross-reader matrix: #600 checks tensor writers and same-engine reads, while #597's 4×4 matrix covers ESKM models. This adds each native JIT, native AOT, VM source, and VM bytecode producer to every consumer, with three small f64 tensors per pair.

Each engine asserts tensor identity, rank, dimensions, element count, and f64 dtype. An independent Python oracle compares all 12 producer files and 48 consumer rewrites against exact expected host-endian ESKT v1 bytes, including signed zero. The runner also verifies that each consumer leaves its input copies unchanged. CTest runs the matrix and 48 negative controls that change only Python's expected bytes; no malformed file is given to any reader.

## Revision and dependencies

- Head: `44b366442f1d60d39dfdb63b2033e7ebbbb0672d` (`codex/eskt-cross-engine-matrix`).
- Base and fetched upstream master: `5ce74beeac25aca56c0c8129083e6db77f96f7dc`.
- No dependency on #596, #597, #600, or their fixtures/runtime changes. Current source inspection confirms public VM tensor I/O IDs 1820/1821, distinct from old model-I/O helper IDs 802/803.
- Five files only: fixture, runner, CTest registration, and reference documentation/index. No reader/writer, policy/API, generated artifacts, ICC, or release-workflow changes.

## Verification

PASS at initial implementation head `0e3c54c2648b7ef7f570e0b5062d87b40a6078b0` on a private Linux Release build with LLVM 21.1.8 / Clang 22.1.6, BLAS/GPU/agent FFI disabled, maximum two compiler jobs:

```sh
cmake --build .scratch/eskt-build --target eshkol-run eshkol-vm-standalone-test \
  model_io_test eshkol-vm-prelude-cache-gen --parallel 2
ctest --test-dir .scratch/eskt-build --output-on-failure \
  -R '^(eskt_tensor_engine_parity|model_io_test|vm_canonical_stdlib_surface_smoke|vm_prelude_cache_is_current)$'
```

- **PASS:** 4/4 focused CTests. Matrix: 16/16 pairs, 12 exact producer files, 48 exact consumer rewrites, 48/48 wrong expected-byte refusals.
- **PASS:** direct runner with `--self-test --keep`, Python compilation, test-coverage inventory, API-doc freshness, `git diff --check`, and read-only harness review.
[...]

PR #617, opened 2026-09-09
test(io): align public tensor parity with ESKM dispatch

Gabriel-Kahen:
## Problem

#555 merged public native/VM tensor save/load onto ESKM v1, but #615's exact-byte matrix still expected legacy ESKT. On a fresh build of master `99773a83`, the old runner fails at the first JIT vector: actual 85 bytes, expected 68. These are ordinary valid tensor outputs, not malformed-input tests.

## Change

- Rename the runtime gate to `eskm_tensor_engine_parity` and its scripts/fixture to match the current public API contract. No production runtime code or dispatch changes.
- Encode the normative single-record ESKM v1 layout independently: empty name, little-endian fields, f64 payload bits, CRC-32 footer. Do not infer expected bytes from producer output or accept both formats.
- Preserve all four engines, three shapes, 16 producer/consumer pairs, metadata checks, 12 exact producer outputs, and 48 exact rewrites. Wrong-expectation controls modify only Python expectations, never files supplied to Eshkol.
- Add three offline oracle unit tests: four immutable v1.2.4 single-record fixtures (including signed zero and a specific NaN payload), explicit wire fields, and wrong-expectation refusal.
- Correct the dedicated parity documentation and stale public-dispatch statements. Preserve the historical documentation URL and clearly label the handoff's older ESKT evidence. Both design proposals remain Proposed.

## Verification

Fresh Linux x86-64 Release build, GCC 16.1.1 / LLVM 21.1.8; GPU, BLAS, quantum, and agent FFI disabled.

```sh
cmake --build build --target eshkol-run eshkol-vm-standalone-test --parallel 2
ctest --test-dir build --output-on-failure -R '^eskm_tensor_(engine_parity|oracle_self_test)$'
python3 scripts/check_public_api_docs.py
python3 scripts/check_test_coverage.py
git diff --check
```

- Focused CTest: 2/2 PASS. Runtime matrix: 16/16, 12 producer files, 48 rewrites, 48/48 wrong-expectation controls. Offline oracle: 3/3 unit tests PASS.
- Public API docs, coverage inventory, and whitespace checks: PASS.
- Baseline negative control: the unchanged pre-fix ESKT runner FAILS on the same fresh runtime build with the exact-byte mismatch above.
- Textual merge checks: clean against #597 at `43923f47`, #598 at `7c88d864`, and the maintainer's refreshed atomic-save branch at `8434e2eb`. The original #600 head remains conflicted with master; this PR does not repair or add to that inherited conflict set.

Not run: full suite, macOS/Windows, big-endian execution, malformed-input runtime tests, fuzz/resource campaigns. Runtime scalar/empty materialization and v2 implementation remain outside scope. This does not restore or promise legacy ESKT compatibility, nor add a release-oracle claim.


tsotchke:
Superseded: the identical change (same patch-id) landed on master through #620 (merge commit a1157d42). Closing without merge.

PR #612, opened 2026-09-07
release: v1.3.5-evolve integrated candidate

tsotchke:
## What changed

Integrates the v1.3.5-evolve release candidate: guarded VM loading and resolver fixes, geometry/AD corrections, Unicode and checkpoint safety fixes, XLA S0-S7 verification, Rosette oracle binding, disclosure enforcement, regenerated VM prelude data, and release readiness checks bound to the exact checkout.

## Verification

- disclosure scan: PASS
- oracle schema: 357/357
- surface counts: PASS (1109 constructs, 1050 builtins)
- CI cost policy: PASS
- release readiness guard tests: 5/5
- VM prelude cache check: PASS
- release workflow surface test: PASS
- full CTest receipt: .scratch/ctest-v135.log (316/337; remaining failures are tracked in the release handoff)

The release tag remains gated on required GitHub checks and final release evidence.

PR #619, opened 2026-09-09
fix(gpu): recognize current CUDA dispatch logs

### LINKED ISSUE #604: GPU correctness gate skips current cublasDgemm dispatch logs

tsotchke:
Carries #605 by @RichardHoekstra rebased onto current master so it runs current CI. Original commits and authorship preserved. Supersedes #605 on merge.

Closes #604

### DIFF
--- a/tests/gpu/gpu_correctness_gate.sh
+++ b/tests/gpu/gpu_correctness_gate.sh
@@ -97,6 +97,19 @@ log()  { printf '%s\n' "$*"; }
 skip() { log "SKIP: $*"; exit 0; }
 fail() { log "FAIL: $*"; emit_trace FAIL "$*"; exit 1; }
 
+# Runtime evidence must name a live backend signal, not merely prove that a GPU
+# library was compiled in. Keep the accepted spellings in one place so the
+# self-test and the real-device gate cannot drift apart as backend logs evolve.
+GPU_DISPATCH_LOG_RE='\[GPU\] (Metal: |.*-> CUDA cuBLAS|'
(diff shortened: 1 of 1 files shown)


tsotchke:

PR #623, opened 2026-09-09
feat(model-io): publish checkpoint saves atomically

tsotchke:
Carries #600 by @Gabriel-Kahen rebased onto current master so it runs current CI. Original commits and authorship preserved. Supersedes #600 on merge.

This carry drops the PR's `.github/workflows/ci.yml` hunk (an added "Atomic checkpoint save contract" ctest step) per maintainer ruling — the underlying tests it would have gated (model_io_test, atomic_checkpoint_file_test, model_io_production_archive_test, atomic_checkpoint_save_test, etc.) are still added by this PR and still run via normal ctest discovery; only the new named CI step is excluded.

Rebase also dropped a now-dead ESKT-format standalone `tensor-save`/`tensor-load` implementation the PR's first commit reintroduced in `lib/core/system_builtins.c` and `lib/backend/vm_native.c` (VM opcodes 1820/1821) — master already removed that duplicate/shadowed path in favor of the single validated ESKM-v1 path (VM opcodes 802/803 in `lib/backend/vm_model_io.c`, native `eshkol_tensor_save_tagged`/`eshkol_tensor_load_tagged` in `lib/core/model_io.cpp`), which already carries this PR's atomicity changes cleanly.


tsotchke:

PR #634, opened 2026-09-10
fix(numeric): exact expt over the rational tower, vector literals keep exact elements, and the VM reads, serializes and prints bignum rationals exactly

tsotchke:
## Summary

- `(expt base exp)` silently overflowed to inexact `0` for an exact rational base (`(expt 1/3 50)` printed `0`), and degraded to inexact on a negative-exponent bignum-overflow case (`(expt 10 -30)` printed `1e-30`), on both the native LLVM backend and the bytecode VM. Root-caused to `eshkol_bignum_pow_tagged`'s exact-path gate excluding a rational base entirely (falling into a fallback that reinterpreted the rational's field layout as a bignum's) and to the VM's `expt` dispatch never checking `VAL_RATIONAL` or a negative exponent at all. Fixed on both engines with an exact numerator/denominator repeated-squaring path (SW-152).
- A numeric `#(...)` vector literal containing an exact-rational, bignum-magnitude, or non-numeric element was silently promoted to an f64-backed tensor, reinterpreting the element's heap payload as raw double bits — a memory-unsafe read, not just a precision loss (`(vector-ref #(1/2 1/3) 0)` read back garbage). Fixed at the parser level: such a literal now stays a genuine vector with every element preserved exactly, while a rectangular nest of plain numeric literals keeps Eshkol's documented tensor-literal flattening unchanged (SW-153).

## Test plan

- [x] `tests/core/exact_expt_test.esk` — 34/34 checks pass on native `-r` (JIT) and `-o` (AOT), wired into CTest (`exact_expt_jit_smoke`/`exact_expt_aot_smoke`)
- [x] `tests/core/vector_literal_exactness_test.esk` — 30/30 checks pass on native `-r`/`-o`, wired into CTest (`vector_literal_exactness_jit_smoke`/`_aot_smoke`)
- [x] `tests/vm_parity/corpus/77_exact_rational_expt.esk` — output matches byte-for-byte between native `-r` and `eshkol-vm-standalone-test`
- [x] `tests/vm_parity/corpus/78_vector_literal_exactness.esk` — output matches byte-for-byte between native `-r` and `eshkol-vm-standalone-test`
- [x] Task repro reproduced and confirmed fixed on JIT, AOT, and VM
- [x] `ctest --test-dir build -R 'exact|rational|vector|vm_parity|literal'` green (19/19)
- [x] `./scripts/run_all_tests.sh` green
- [x] `./scripts/run_vm_parity.sh` green
- [x] `python3 scripts/gen_api_docs.py --check`, `gen_language_surface.py --check`, `check_surface_counts.py --no-trace` clean
- [x] `python3 scripts/gen_silent_wrong_ledger.py --check` and `scripts/check_ledger_integrity.py` clean (142 entries)


tsotchke:
Carried verbatim by #650 (fix/bignum-rational-memory-reclamation), which was rebased on top of it. Closing without merge; the work lands through #650.

PR #635, opened 2026-09-10
feat(stdlib): a Navier-Stokes residual oracle over AD and the exact tower

tsotchke:
## Summary

- Adds `core.pde.ns-residual`, a residual oracle for incompressible Navier-Stokes: turns a candidate flow into its residual force `R = d_t u + (u . grad) u - nu Lap u + grad p` and divergence, computed entirely from AD partials of the flow's own procedures (cylindrical axisymmetric and Cartesian representations, cross-checked exactly against each other at a rational point on a polynomial field), rational Simpson quadrature for energy/dissipation (exact for a low-degree polynomial field), a similarity-coordinate flow constructor, a tau-series/singular-order report per component so a candidate ansatz can be scored mechanically, and a finite-difference-free smoothness probe for a proposed force/cutoff. 24 exported symbols.
- `docs/reference/stdlib/ns-residual.md`, `docs/reference/stdlib/INDEX.md` and `CHANGELOG.md` updated.
- `tests/stdlib/ns_residual_test.esk` (22 checks, native JIT + AOT) and `tests/vm_parity/corpus/78_ns_residual.esk` (the module's VM-portable surface; the residual/divergence functions themselves are native-only, inheriting the existing documented VM boundary on higher-order `derivative-n`/`taylor`).
- Found and fixed one bug during development: the axial residual component's pressure-gradient term used `d_r` instead of `d_z`.
- Found a compiler limitation (not fixed, routed around): nesting a second order-2-or-higher AD pass (`taylor`/`derivative-n`) around this module's own order-2 Laplacian pass fails, and in one specific shape returns a silently wrong answer instead of erroring. `ns-residual-tau-series`/`ns-force-smoothness-probe` cap their `order` argument at 1 and raise a clear error above that.

## Dependency note

This branch is built on `origin/fix/ad-exact-rational-and-nesting` (#630) with `origin/feat/exact-rational-linalg` (#632) merged in, since this module relies on both (the nested-AD/exact-rational-seed fix, and follows `core.exact_linalg`'s conventions). Both are still open as of this PR. Until they merge, this diff includes their commits; once they land on master this branch will be rebased and the diff will shrink to just this module.

## Test plan

- [x] `cmake --build build -j8` clean
- [x] `tests/stdlib/ns_residual_test.esk` passes on native JIT (`-r`) and AOT: 22/22
- [x] `tests/vm_parity/corpus/78_ns_residual.esk` agrees between native and the bytecode VM
- [x] Full `scripts/run_stdlib_tests.sh`: 24/24 (including this test)
- [x] `scripts/check_surface_counts.py --no-trace`: PASS
- [x] `scripts/gen_api_docs.py --check`: up to date


tsotchke:
Consolidated into #656 (feat/ns-mechanization-wave), which carries this PR's commits verbatim on top of #654 so the stack costs one verification run. Closing without merge; the work lands through #656.

PR #638, opened 2026-09-10
feat(stdlib): symbolic polynomial and truncated power-series values over the exact tower

tsotchke:
Adds core.symbolic: sparse multivariate polynomials and truncated multivariate power series with exact rational coefficients, Laurent leading-order queries, exact Taylor transcendentals (exp, log1p, sin, cos, sqrt1p), and symbolic evaluation of quoted expressions. 74 checks on JIT and AOT, VM parity corpus 28_symbolic. Depends on #630 and #632 (their commits are carried on this branch); rebase after they land.


tsotchke:
Consolidated into #656 (feat/ns-mechanization-wave), which carries this PR's commits verbatim on top of #654 so the stack costs one verification run. Closing without merge; the work lands through #656.

PR #645, opened 2026-09-10
fix(codegen): quoted bignum and bignum-rational literals are the same values as evaluated ones

tsotchke:
## Summary

SW-163 (SILENT-WRONG). Native's quote lowering (`lib/backend/llvm_codegen.cpp`, `codegenQuotedAST()` — what `(quote datum)`/`'datum` compiles through, and what `codegenQuasiquote()` falls through to for a bare atom or any non-splice list element) had no case for `ESHKOL_BIGNUM_LITERAL`, the AST type `parser.cpp` gives an integer literal too large for `int64_t`. Falling into `default:` returned a tagged null:

```scheme
(display (number->string '123456789012345678901234567890))   ; printed "0"
(display (exact? '123456789012345678901234567890))            ; was #f
(display (car '(123456789012345678901234567890 1)))           ; was ()
```

A related gap: `n/d` (including a bignum-magnitude numerator or denominator) desugars at parse time to a `(make-rational n d)` `CALL_OP` — the same AST shape a literally written `(make-rational n d)` call has. `codegenQuotedOperation()`'s `ESHKOL_CALL_OP` case quoted every call as list data unconditionally, so `'1/123456789012345678901234567890` quoted to the 3-element list `(make-rational 1 123456789012345678901234567890)` instead of the rational number.

**Fix:**
- `codegenQuotedAST()` gained `case ESHKOL_BIGNUM_LITERAL`, constructing the same bignum heap object the evaluated path builds for the identical literal via `eshkol_bignum_from_string()` — the exact runtime entry point the evaluated case already calls.
- `codegenQuotedOperation()` gained `isQuotedRationalLiteralDesugar()` — the identical narrow-shape whitelist `autodiff_codegen.cpp`'s `towerSafeExpr()` already applies for the analogous reason (`make-rational` applied to exactly two int64/bignum *literal* operands — the only shape the parser's desugar can produce; a genuine `(make-rational x 3)` call with a variable operand is unaffected) — and `codegenQuotedRationalLiteral()`, which routes through `eshkol_rational_make_tagged()`, the runtime entry point the evaluated `(make-rational n d)` call already uses.

`ESHKOL_QUOTE_OP` (including nested `''x`), quoted vectors/dotted pairs, and `codegenQuasiquote()`'s fallback path all share these two functions, so quasiquote and nested quoted structures inherit the fix with no separate change.

The bytecode VM was already correct here (fixed alongside SW-155 in `fix/vm-bignum-rational-reader-printer`, #640, merged).

**Dependency note:** branched from `fix/exact-rational-expt-and-vector-literals` (#634, still open) at its head `fbfa99bb` — this fix builds on that PR's `make-rational` desugar and `#(...)` exactness plumbing — but opened against `master` per instruction rather than against #634's branch. Will need a rebase once #634 merges.

**Discovered, left open (VM-side, out of scope for this PR):** the VM's own quote lowering (`lib/backend/vm_parser.c` `compile_quote()`) has no case for its own rational-literal desugar `(exact-rational n d)` (a different name than native's `make-rational`), so `'1/3` still quotes to a 3-element list on the VM; and `compile_quasiquote()`'s "Atom: number" arm ignores the `is_bignum` flag entirely. Both are documented in the new test files, which deliberately avoid asserting those two cases so the VM-parity corpus program stays a true native/VM differential.

## Test plan

- [x] `tests/core/quoted_bignum_literals_test.esk` — 29/29 on native JIT (`quoted_bignum_literals_jit_smoke`) and AOT (`quoted_bignum_literals_aot_smoke`): the task repro, negative bignum, quoted bignum-rational (cross-checked against the evaluated spelling and via `equal?`/`eqv?`), quasiquote with a bignum atom and alongside an `unquote`, quoted vector literal + `(vector ...)`-procedure wrapper, quoted dotted pair, and negative controls (small int/inexact/small-rational literals, a genuine `(make-rational x 3)` call unaffected).
- [x] `tests/vm_parity/corpus/80_quoted_bignum_literals.esk` — byte-identical across native `-r`, VM source and VM ESKB (`scripts/run_vm_parity.sh`: 241 passed, 0 failed).
- [x] Ledger: `SW-163` added (`.icc/ledger/entries/SW-163.yaml`, `.icc/ledger/order.txt`), aggregate regenerated (`scripts/gen_silent_wrong_ledger.py`, 146 entries), `scripts/check_ledger_integrity.py` clean.
- [x] `ctest --test-dir build -R 'quote|bignum|rational|vm_parity|homoiconic'` — 4/4 green.
- [x] `./scripts/run_all_tests.sh` — 100% suite pass rate.
- [x] `scripts/gen_api_docs.py --check` and `scripts/gen_language_surface.py --check` — both clean.


tsotchke:

PR #646, opened 2026-09-10
fix(ad): exactness is decided by the carrier at run time, not by a static proof of the differentiand

tsotchke:
## Summary

- The exact tier (ESH-0394) that routes `derivative`/`gradient`/`hessian` at an exact point through the same Taylor-tower pass `derivative-n`/`taylor` already run unconditionally used to gate entry on a static AST whitelist (`towerSafeExpr`) over the differentiand's body and its point expression. Any shape the whitelist could not prove pure fell back to the always-inexact jet path, or — for a function-call-expression differentiand passed to `taylor`/`derivative-n` — hit a hard compile failure instead of a wrong answer.
- Root cause: exactness was being decided from the differentiand's source text at compile time, when it is a property of the carrier at run time — `adExactTowerGate` already reads the point's runtime tag to choose the route, and the tower's own arithmetic already keeps a value exact through composition by ordinary R7RS contagion.
- Fix: removed the static whitelist entirely. Eligibility is now purely structural (can the function resolve to the shape the tower pass's synthetic op requires); a general-expression fallback lets any expression that evaluates to a closure serve as a differentiand; the point is evaluated exactly once and threaded into the exact arm instead of re-evaluated, removing the need for a purity proof.
- Two regressions found and fixed during hardening: a multi-argument named function could reach a closure-capture-loading path meant for lambda captures (malformed IR); a differentiand returning a non-number (`R -> R^n`) could be silently substituted with a wrong scalar by the tower and crash downstream. Both are now guarded.
- Also closes the same class of defect for a differentiand whose body branches on `if`, which depends on #644 (SW-158, comparisons/branch selection inside a differentiand acting on the carrier's primal) already merged on this branch's base.

Filed as ledger SW-159/160/161 (SILENT-WRONG) and SW-162 (LOUD-ERROR).

Depends on `fix/ad-exact-rational-and-nesting` (#630); this branch is rebased onto its head (through #644 / SW-158).

## Test plan

- [x] `tests/ad/exactness_runtime_property_test.esk` — 67 checks (JIT + AOT + CTest), covering the four original repro shapes, composition depth 1..6, constants reached through define/let/letrec/a parameter object, point arguments from car/vector-ref/hash-ref/a loop variable/a function return, a differentiand given as a lambda/a variable/a call expression/a let-bound closure/`(compose ...)`, the `if`-conditional shape, and negative controls
- [x] `tests/vm_parity/corpus/79_exact_tier_runtime_property.esk` — byte-identical across native/vm-src/vm-eskb
- [x] Full CTest suite green apart from pre-existing, unrelated failures (confirmed present on the unmodified base): `abi_header_inventory_*` baseline drift and one `eskt_tensor_engine_parity` serialization parity failure
- [x] `scripts/run_all_tests.sh` — 847/847 across all 46 suites
- [x] `scripts/run_vm_parity.sh` — 239/239
- [x] The nine `examples/mathematics_navier_stokes_*.esk` programs from `feat/ns-examples-strengthen` still print `RESULT: ALL PASS`
- [x] `gen_api_docs.py --check` and `gen_language_surface.py --check` clean


tsotchke:

PR #647, opened 2026-09-10
fix(numeric): exact square and n-th roots over the exact tower; tensors never silently zero an exact element

tsotchke:
## Summary

- **MS-05 / SW-167**: `(sqrt 16)`, `(sqrt 1/4)`, `(expt 4 1/2)` etc. now return exact results per R7RS 6.2.6 when the exact root exists, on both the native and VM engines. New shared primitive `eshkol_bignum_iroot` (exact n-th integer root via Newton's method) backs `eshkol_exact_sqrt_tagged` / `eshkol_exact_rational_pow_tagged`. Fixing this surfaced and closed an adjacent, unrelated pre-existing VM defect: `sqrt`/`log`'s inexact fallback read the heap-blind `as_number()` instead of `as_number_vm()`, so `(sqrt 1/2)` silently returned `0`.
- **MS-04 / SW-166**: `(tensor (vector 1/2 1/3))` no longer silently stores `#(0 0)`. Two independent root causes closed: `coll_leaf_double` (the single-collection-argument `(tensor X)` path) hardcoded `0.0` for any non-double heap value, and `TensorCodegen::extractAsDouble` (the chokepoint for `tensor`'s multi-arg form, `make-tensor`, `tensor-set!`, `vector->tensor`) was a separate, incomplete duplicate of the correct `ArithmeticCodegen::extractAsDouble` that converted a rational's heap pointer address to a double via `SIToFP` — the second bug was found while writing this fix's own regression tests, not by the original ticket.
- Ledger entries `SW-166`/`SW-167` added (`.icc/ledger/entries/`), aggregate regenerated, integrity check green.
- Docs updated: `docs/COMPLETE_LANGUAGE_SPECIFICATION.md` §§14.4.2–14.4.4, `docs/breakdown/EXACT_ARITHMETIC.md` §11.9, `docs/reference/tensors/creation.md`.

Depends on #634 (`fix/exact-rational-expt-and-vector-literals`) — this branch is based on it and should merge after it.

## Test plan

- [x] `tests/core/exact_roots_test.esk` — 47/47 PASS on native JIT, AOT, **and the VM** (all three engines)
- [x] `tests/core/tensor_exact_elements_test.esk` — 22/22 PASS on JIT/AOT (VM-covered subset verified separately; `vector->tensor` is not implemented on the VM at all, a pre-existing unrelated gap)
- [x] `tests/vm_parity/corpus/80_exact_sqrt_and_expt.esk` and `81_tensor_exact_elements.esk` — output byte-identical between native `-r` and `eshkol-vm-standalone-test`
- [x] CTest wired: `exact_roots_{jit,aot,vm}_smoke`, `tensor_exact_elements_{jit,aot}_smoke`
- [x] `scripts/run_all_tests.sh`: 848/848 individual tests, 46/46 suites, 100%
- [x] `ctest --test-dir build -R 'exact|rational|sqrt|tensor|vm_parity'`: 33/34 — the one failure (`eskt_tensor_engine_parity`) is pre-existing and unrelated, confirmed by reverting this change's files to `HEAD` and reproducing the identical failure
- [x] `gen_api_docs.py --check`, `gen_language_surface.py --check`, `check_surface_counts.py --no-trace`: all clean


tsotchke:

PR #648, opened 2026-09-10
fix(vm): quoted and quasiquoted rational and bignum literals are exact on the bytecode VM

tsotchke:
## Summary

- SW-168: the bytecode VM's own quote/quasiquote lowering had two gaps native already closed for the same class of defect (SW-163, #645).
  - `lib/backend/vm_parser.c` `compile_quote()` had no case for the VM reader's own rational-literal desugar — `1/3` syntax reads to the list node `(exact-rational num denom)` (a different desugar name than native's `make-rational`) — so `'1/3` quoted to a 3-element list instead of the rational value.
  - `lib/backend/vm_compiler.c` `compile_quasiquote()`'s numeric-atom arm ignored the `is_bignum` flag entirely, so a quasiquoted bignum atom silently lost its exactness.
- Fixed by giving both functions a dedicated `exact-rational` case (guarded to both operands being literal `N_NUMBER` nodes, the only shape the desugar produces, mirroring SW-163's `isQuotedRationalLiteralDesugar()` whitelist), and by having `compile_quasiquote()`'s numeric-atom arm delegate to `compile_quote()` instead of duplicating a stale subset of its discrimination.

**Depends on #645** (`fix/quote-bignum-literals`) — this branch is cut from its head (`ed5b0fdf`), which fixed the equivalent gap on native codegen.

## Test plan

- [x] `tests/vm/quote_exact_literals_test.esk` (ctest `quote_exact_literals_vm_smoke`) — 35/35, covering `'1/3`, a quoted bignum, a dotted pair mixing rational/bignum elements, a quasiquote template mixing `unquote` with rational/bignum atoms, nested quoted vectors, `equal?` between quoted and evaluated forms, and negative controls (inexact literals stay inexact quoted and quasiquoted, a verbatim `|1/3|` symbol stays a symbol, a genuine `(exact-rational x 3)` list with a variable operand stays list data).
- [x] `tests/vm_parity/corpus/81_vm_quote_exact_literals.esk` matches byte-for-byte across native `-r`, VM source and VM ESKB.
- [x] `scripts/run_vm_parity.sh`: 243 passed, 0 failed.
- [x] `ctest --test-dir build -R 'vm|quote|bignum|rational'`: 28/28.
- [x] `./scripts/run_all_tests.sh` (foreground): 46/46 suites, 848/848 individual tests.
- [x] `gen_api_docs.py --check` and `gen_language_surface.py --check` clean.
- [x] `scripts/check_ledger_integrity.py --no-trace`: PASS (SW-168 split entry added, aggregate regenerated).


tsotchke:

PR #649, opened 2026-09-10
fix(ad): differentiation passes nest at any depth and any order; a foreign-level carrier is a coefficient, never a constant or an error

tsotchke:
Nested differentiation was a table of special cases built on one value series
plus one first-order companion, so it covered exactly the shapes where at most
one enclosing level was live, at first order. Past that ceiling the same
mathematics was correct, a loud refusal, or a silent zero depending only on
which spelling was used:

```scheme
(define (h r) (* r r r r))
(define (g t1 t2) (derivative-n h (+ 1 t1 t2) 2))

;; d/dt1 d/dt2 [12(1+t1+t2)^2] at 0 = 24
(derivative-n (lambda (t1) (derivative-n (lambda (t2) (g t1 t2)) 0 1)) 0 1)
;; => 0, silently, exit 0

(taylor (lambda (t) (* t (derivative-n h (+ 1 t) 2))) 0 2)
;; => raised "unsupported nested differentiation (both passes order >= 2)"
```

`docs/design/AD_TAYLOR_TOWER.md` and `docs/guide/AUTOMATIC_DIFFERENTIATION.md`
section 11 both state that nesting is safe at any depth. Filed as ledger
**SW-154**.

## Root cause

The ceiling was the representation, not the mathematics.
`eshkol_ad_nested_seed`'s `CARRY_TWR` branch built the inner tower's companion
from `nest_coeff(outer, 1)` and dropped the outer tower's own companion and
`carry_epoch` — the second enclosing level — without a word
(`lib/core/runtime_taylor.c:1818-1830` on the base branch). `foreign_first_order()`
raised as soon as a foreign level carried any coefficient above index 1.

## Design

Every pass owns a **level** (its epoch). A carrier at level *L* of order *K* is
the truncated series in *L*'s perturbation whose **coefficients are themselves
carriers of the enclosing levels** — the "jet over a jet" the design has named
since section 1.

- `ESH_TAYLOR_COEFF_CARRIER`: a tower whose `c[]` is `eshkol_tagged_value_t`,
  each entry any tagged number **including** another tower whose epoch names a
  strictly enclosing level.
- A foreign level is a **coefficient-valued constant** of the active level —
  the whole tower, structure and level tag intact, in `c[0]`. Nothing is
  flattened, so nothing is lost, so nothing has to raise.
- The `trc_*` recurrences are the `tr_*` algebra written over "whatever the
  coefficients are": every scalar operation dispatches back into the tower
  kernel when a coefficient is itself a carrier. Depth and per-level order are
  bounded only by the levels a program opens.
- Seeding opens a level over whatever the point already is; extraction returns
  `k!·c[k]`, again a carrier of the enclosing level.
- 8-jet boundary: a pass that would seed a jet but meets a tower level runs as
  an order-1 level. A single enclosing jet direction still rides the section 8
  companion (no allocation per coefficient op — the shape a forward-over-forward
  Hessian-vector product takes in a loop); two directions become levels of their
  own.
- Codegen mirrors the tower-context depth into the runtime
  (`eshkol_ad_tower_enter/leave`): its own `__ad_tower_active` has **internal**
  linkage in the AOT build, one private copy per module, so the runtime could
  never read it. It also passes whether a reverse tape is live, so the ESH-0093
  mixed-mode protocol keeps the jet slots it owns.

Perturbation confusion is impossible by construction rather than by convention:
two series combine only when their level tags are equal, and a tag is never
dropped or reused. The Siskind–Pearlmutter cases are negative controls and
answer 1, not 2.

Exactness now survives a nest — a nested level's coefficients are the same exact
numeric tower every other coefficient is:

```scheme
(derivative-n (lambda (a) (derivative-n (lambda (b) (* a a a b b b)) 1/5 2)) 1/3 2)
;; => 12/5, exact
```

Retired as unreachable: `eshkol_ad_tower_carry_result`,
`eshkol_ad_nested_capture_unsupported` (and their WASM env stubs),
`esh_taylor_t.carry_epoch`.

## Tests

- `tests/ad/nested_towers_matrix_test.esk` — 164 checks: the depth-2 operator ×
  order matrix (`derivative`/`derivative-n`/`taylor` as outer and inner, orders
  1..3 on each side), the depth-3 operator cube, the full 1..3 order sweep with
  three live levels, nesting through the evaluation point, exact seeds plus
  exactness assertions, two enclosing 8-jet levels around a tower pass, and the
  Siskind–Pearlmutter confusion controls.
- `tests/ad/nested_operator_matrix_test.esk` gains the both-passes-order-≥2
  cells that used to raise.
- Both on the JIT and AOT lanes.

`ctest -R 'ad|taylor|derivative|gradient|hessian|nested|conditional|vm_parity'`
and `scripts/run_all_tests.sh` are green. The three `abi_header_inventory_*`
failures are pre-existing on this base: the ratchet produces byte-identical
findings at `4757bdf0` and at this commit, about
`tests/ffi/sqlite_column_roundtrip_test.c`, which this change does not touch.
The `eskt_tensor_engine_parity` failure is likewise pre-existing and unrelated —
the writer emits an `ESKM`-magic record the parity script still reads as `ESKT`
v1.

Cost at the deepest shape measured — depth 3 with order 4 on every level, 5³ =
125 coefficients — is 0.26 ms per nested differentiation.

The bytecode VM has no tower and already raises on a nested `derivative`
(`lib/backend/vm_native.c` case 393, ESH-0369); that documented raise is
unchanged.

## Dependency

Stacked on **#630** (`fix/ad-exact-rational-and-nesting`), whose branch also
carries the SW-158 conditional fix at `4757bdf0`. This branch is rebased onto
that commit and should merge after #630.


tsotchke:

PR #650, opened 2026-09-10
fix(runtime): exact rational temporaries are reclaimed like integers; the heap limit is a fail-closed contract

tsotchke:
**Contains #634.** This branch was cut from `fix/exact-rational-expt-and-vector-literals` because it builds on that PR's exact-tower fixes. Master has since moved, so the branch was rebased onto it, and the rebase carried #634's three commits along — SW-152 (`expt` exact for a rational base and negative exponents), SW-153 (`#(...)` literals keep exact elements) and the VM bignum-rational reader/printer fix. They appear here as their own commits, ahead of the SW-164/SW-165 work.

So #650 is self-contained and needs nothing merged before it. If #634 is merged first, rebase this branch again and those three commits will drop out as already applied; if #650 is merged first, #634 becomes empty and should be closed rather than merged.

## SW-164 — exact rational temporaries are reclaimed like integers

Exact rational arithmetic grew resident memory in proportion to the **work** an operation did rather than the **values** it produced, while the identical loop over bignum **integers** stayed flat. That asymmetry had four causes, in four different layers, and all four had to close before it did. A fifth, found alongside them, cost the same reclamation to every loop written with `cond`, exact arithmetic or not.

**1. Reduction ran on the result.** `bn_gcd` reduced each result by running Euclid over the full-width numerator and denominator; the step count grows with their digit count and every step bump-allocates a quotient, a remainder and two limb buffers. Reduction now follows Knuth (TAOCP 4.5.1) — reduce first, multiply second — so every GCD is taken on operands no larger than the inputs. For a running exact sum the only GCDs left are between the large running denominator and the small term denominator; for a multiply by a coprime value (the identity included) the GCD disappears. Values are unchanged: both forms produce the unique lowest-terms representation with a positive denominator.

**2. A primitive had no reclamation boundary.** The arena reclaims only at a scope or region boundary and a numeric primitive had none inside it, so whatever scratch remained was retained for the life of the arena. New arena primitive `arena_scope_end_retaining()` ends the innermost scope *retaining* a named set of header-prefixed objects; `rat_scoped()` brackets every exact-rational operation with it, retaining only the result and its numerator and denominator. An operation now allocates its result and nothing else.

**3. The per-iteration loop scope gave up instead of promoting.** ESH-0214b reclaimed an iteration only when nothing flowing into the next one pointed into it, and otherwise **committed** — retained the whole iteration, as if the feature were off. That is the common case, not the rare one: any loop that accumulates builds its accumulator inside the iteration. A loop now opens a LOOP scope at entry, and an escaping back edge evacuates the loop-carried values out of the span, rewinds to the loop's entry mark, and copies them back — the arena and a scratch arena forming a semispace, so resident size is bounded by the live set rather than the iteration count.

Two things keep that honest:

- Promotion is gated the way a copying collector gates a minor collection — only once the span has grown to a multiple of the last measured live set — so a loop that grows its accumulator **by accretion** (consing onto a list), where the span *is* the live set, is never turned from linear into quadratic. It falls back to the previous retention, which is the right trade there.
- **What may be moved is deliberately narrow.** Copying an object is half of moving it; the other half is rewriting every other reference to the original, and the evacuator can rewrite only what it reaches from the roots it was handed. `loopBodyIterScopeSafe` rules out references the body creates by storing into pre-existing Eshkol structure, but says nothing about the C runtime's own roots — the AD tape's `nodes[]` array being the one that matters, where a moved node yields a plausible wrong gradient rather than an error. Rather than enumerate runtime-side roots, a list that would go stale the first time one was added, promotion admits only values that cannot be held that way: immediates and the exact tower's heap payloads (bignum, rational), which are pure values. Anything else, and any loop running while a tape is recording, retains the span.

The evacuator itself is the existing region evacuator with its ownership question abstracted (`EvacState::owns`): one Cheney copier, one forwarding map, one per-subtype interior walk, so the two callers cannot drift the way ESH-0214d showed they can.

**4. The scope's own admission test rejected almost every `cond`.** The parser stores a cond CLAUSE in a call node whose `func` slot holds the clause's **test**, not a callee, and whose variables are the body — but the analysis walked clauses as ordinary expressions, so each one reached the generic call arm, asked "is this a callee I can analyze?", found a computed callee in a test like `(unit k)`, and returned false. An `else` clause failed the same way, as an unknown function named `"else"`. Since an unrecognized callee rejects the **whole** loop, very nearly every `cond` silently forfeited its loop's per-iteration reclamation — while the identical loop written with nested `if`, or with `and`/`or`, was always flat, which is what identifies the cond path rather than the allocation as the cause. Clauses are now taken apart structurally, and `else` is treated as the keyword it is in that position.

**5. The exact tower allocated outside the loop's reclamation domain, and a literal was allocated per evaluation.** `arithmetic_codegen.cpp` and the `numerator` / `denominator` / `make-rational` / `rationalize` sites emitted their allocations against the `__global_arena` module global — the shared current-arena *slot*, which `with-region` hijacks but which the ESH-0214e nursery does not touch, since it redirects the thread-local memory context. Exact arithmetic therefore landed where a loop's own per-iteration reset could not see it; wrapping the identical body in an explicit `with-region` reclaimed all of it, which is what isolated this cause. It now routes through `eshkol_current_arena()`, as every other loop temporary does.

Separately, the reader desugars `1/2` into a `(make-rational 1 2)` call, so a literal was a heap allocation per evaluation **and**, the constructor not being on the iteration scope's pure-builtin list, disqualified its whole enclosing loop from reclamation. Literals with integer operands are now materialized once into a module-level slot from `eshkol_literal_arena()`, an arena on which no scope is ever pushed and which is never reset — picking the right arena is not enough, because a constant materialized lazily on a loop's first iteration lands above that loop's mark and dies on the first rewind.

## SW-165 — the heap limit is a fail-closed contract

Crossing the heap ceiling printed `Heap limit exceeded` once per arena block for the rest of the run — including on the **default** ceiling nobody asked for — and then exited 0, because `eshkol_track_allocation()` did the diagnosing itself and the interrupt it requested is only ever acted on for timeouts. A sub-megabyte ceiling printed as `0MB > 0MB`, and a malformed value was silently discarded, so an operator who set a bound believed one was in force when it was not.

Accounting now only accounts. Enforcement belongs to `create_arena_block()`, the one site that can carry it out, which reports the breach **once**, in bytes, and exits nonzero without completing. With no ceiling requested the default is an accounting reference and says nothing. A malformed `ESHKOL_MAX_HEAP` / `ESHKOL_MAX_STACK` / `ESHKOL_MAX_TENSOR_ELEMS` / `ESHKOL_MAX_STRING_LEN` now names itself, the offending value and the accepted grammar before falling back to its default, instead of falling back in silence. Reporting rather than refusing is deliberate — `ESHKOL_STACK_SIZE` already reported, and the three non-size variables in the family fall back on a bad value too, so refusing for four of the seven would leave the family answering one mistake three different ways.

## Tests

`tests/memory/bignum_rational_flat_rss_test.sh` — five cases across JIT and AOT: exact H_40000 under a 200MB / 2s ceiling, 100000 exact-rational multiplications under 300MB, and `math_acceptance_exact_rational_memory` gated as a **ratio** (2×, 64MB floor) against a control running the identical loop shapes over machine integers, so it asserts the exact/inexact asymmetry — the defect — rather than an absolute figure on one machine. Every fixture checks its own **answer** as well as its shape (closed forms, a lowest-terms round trip, equality after 100000 multiplications by one), so a reclamation bug that frees something live fails as a wrong result rather than passing as a memory win.

Proven red by building this tree with the changed sources reverted to the branch point: the gate failed there on peak RSS by two orders of magnitude over the ceiling.

`tests/memory/heap_limit_fail_closed_test.sh` — the contract in both directions, with the no-ceiling and ample-ceiling cases as negative controls: without them, a runtime that failed every run would pass the fail-closed case.

The acceptance fixture and its control also carry a `cond`-clause loop whose test allocates, against the same loop written with nested `if`.

`tests/numeric/exact_rational_reduction_identities_test.esk` — 19 algebraic identities pinning the canonical form (`x - x`, `x / x`, `x * 1`, sign normalization, re-association, literal against computed), guarding the reduction rewrite and the literal hoisting.

## Verification

- `./scripts/run_all_tests.sh` green (46 suites).
- Both flat-RSS gates green; the 17 `r_*.esk` / `r2_*.esk` exact-tower probes flat at baseline.
- `scripts/check_ledger_integrity.py` PASS (147 entries); `gen_api_docs.py --check` and `gen_language_surface.py --check` clean.
- `scripts/abi_header_inventory.py check`: `arena_scope_end_retaining()` copies a whole header-prefixed object, so it needs the object's total footprint and its allocation base; both are now expressed as `ESHKOL_GET_OBJECT_TOTAL_SIZE` / `ESHKOL_GET_OBJECT_BASE` beside the rest of the `ESHKOL_GET_*` family, and the primitive calls them instead of recomputing the offset — the disposition the ratchet asks for. Its baseline is raised for exactly those four (class, file) pairs, with the reason recorded in the artifact. **The ratchet is separately red at this branch point** for `tests/ffi/sqlite_column_roundtrip_test.c` and `tests/fuzz/eskm_model_fuzz_probe.cpp`, which this branch neither touches nor re-baselines.

Ledger: `.icc/ledger/entries/SW-164.yaml`, `SW-165.yaml`, aggregate regenerated.


tsotchke:

tsotchke:

tsotchke:
Carried onto the v1.3.5 candidate by #675 (merged into #612). SW-154's carrier rewrite is not carried and its ledger entry now reads guarded; it is targeted at v1.4.

PR #653, opened 2026-09-10
fix: element-wise arithmetic dispatches on both operands, and names the failing site

tsotchke:
Two defects from one report, both in the arithmetic path.

## 1. A vector against a scalar crashed the process (ledger LE-18)

```scheme
(display (* (vector 1 2) 2))
```

died with `[Eshkol] fatal signal: SIGSEGV (segmentation fault) at address 0xa`,
exit 139, no diagnostic and no source location — and the runtime's handler then
blamed the wrong thing ("most likely a stack overflow"). So did
`(+ (vector 1 2) 1)` at `0x9`, `(* (vector 1.0 2.0) 2)`, the same forms reached
through a function, and `(tensor-mul (vector 1.0 2.0) 2.0)`. The reverse
operand order, `(* 2 (vector 1 2))`, raised a clean catchable type error and
exited 1.

The addresses identify the defect: they are the SCALAR ITSELF plus the 8-byte
vector length header. `TensorCodegen::tensorArithmeticInternal` chose its kernel
from the LEFT operand alone (`isVector(arg1)`), and `schemeVectorArithmetic`,
the kernel it selects, reads operand 2 as `[len:i64][tagged_value elems...]`
with no check at all — it int-to-ptrs the raw payload and GEPs +8. Only the
right operand was ever type-checked, because the not-a-vector branch routes
both operands through the shared operand check, which does validate. One
operator had two dispositions, selected by which operand happened to be written
first, and the wrong one was a signal no `guard` can intercept.

The same one-sided test also made a MIXED carrier pair **silently wrong** rather
than fatal:

```scheme
(* (vector 1.0 2.0) (tensor 3.0 4.0))   ; before: #(4967213448 0), exit 0
                                        ; after:  #(3 8)
```

**Fix — carrier-first dispatch.** Both operand positions are classified before
either is dereferenced. The Scheme-vector kernel runs only when BOTH operands
are Scheme vectors; every other combination goes to the tensor path, where the
shared operand check validates each side independently — coercing a numeric
vector or list to a rank-1 tensor and raising the identical catchable
`Type error in tensor-mul: expected tensor, got integer` for a scalar in either
position.

That is the documented contract. `docs/reference/tensors/operations.md`: "All
binary elementwise ops require two tensors of matching shape", with scalar
broadcast spelled as its own operator (`tensor-scale`), and this exact shape
already documented as an error. A vector and a rank-1 tensor are two spellings
of one value, so the mixed pair is the element-wise product.

Two supporting defects the one-sided dispatch had been hiding, both fixed here:

- `buildTaggedValue` inserted a raw `double` into the i64 payload slot.
  IRBuilder does not type-check `insertvalue` in a release build, so the result
  was a structurally malformed tagged value whose later readers emitted invalid
  IR (`Load operand must be a pointer`) and failed module verification far from
  the cause. The payload is now coerced bit-identically at the one place that
  builds it.
- A raw double operand was tagged INT64, so the diagnostic called a float an
  integer. It now keeps its DOUBLE tag (`got double`).

## 2. The diagnostic named the wrong line (ledger LE-19)

The report also showed a vector/scalar failure attributed to a function body
that had already returned successfully. Reduced:

```scheme
(define a 2)
(define b 3)
(display (* a b))              ; line 3 — succeeds
(display (* a (vector 1 2)))   ; line 5 — fails
```

```
before   outline_two_sites.esk:3:15: Type error in tensor-mul: …
after    outline_two_sites.esk:5:16: Type error in tensor-mul: …
```

`+ - * /` share one out-lined numeric-tower dispatch helper per module,
emitted at the first site of the operator and called by every other. Its error
branches recorded the location with a COMPILE-TIME constant, so every
arithmetic type error in the program was reported at whichever site emitted the
helper. The exact column made the wrong line credible: a reader was sent to a
working expression and told it was the failing one. Correct only when an
operator has a single live site, which is why it survived — most small
reproducers do.

**Fix.** The location travels with the call: the helper takes
`{file, line, column}` as three extra arguments each call site fills in from its
own position, and a `CodegenContext` source-location override routes every
error-location emitter inside the helper body at those parameters. The override
records the owning `llvm::Function`, so an emitter can never reference an
Argument of another function; `getOrEmitBinaryOutline` saves and restores it so
one helper emitting another keeps its own; and a call site inside a helper
forwards the enclosing helper's parameters, so the position survives arbitrary
nesting.

## Measured, before and after

| form | before | after (JIT / AOT / VM) |
|---|---|---|
| `(* (vector 1 2) 2)` | SIGSEGV `0xa`, exit 139 | `Type error in tensor-mul: expected tensor, got integer`, exit 1 |
| `(+ (vector 1 2) 1)` | SIGSEGV `0x9`, exit 139 | `Type error in tensor-add: …`, exit 1 |
| `(* (vector 1.0 2.0) 2)` | SIGSEGV `0xa`, exit 139 | `Type error in tensor-mul: …`, exit 1 |
| through a function | warning, then SIGSEGV `0xa` | `Type error in tensor-mul: …`, exit 1 |
| `(* 2 (vector 1 2))` | type error, exit 1 | unchanged |
| `(tensor-mul (vector 1.0 2.0) 2.0)` | SIGSEGV `0x4000000000000008` | `… expected tensor, got double`, exit 1 |
| `(* (vector 1.0 2.0) (tensor 3.0 4.0))` | `#(4967213448 0)`, exit 0 | `#(3 8)` |
| `(* (vector 1 2) (vector 3 4))` | `#(3 8)` | unchanged |

The VM was never affected — it raised `*: expected numeric operands` catchably
throughout — so this was also a substrate divergence in which the fast path was
the unsafe one. All three engines now agree on outcome and exit status.

## Gates

- `tests/features/vector_scalar_arithmetic_dispatch_test.esk` — 30 checks: all
  six reported forms, both operand orders for all four operators, the
  through-a-function shape, the mixed vector/tensor product, and the
  element-wise CONTROLS that must keep working.
- `tests/vm_parity/corpus/78_vector_scalar_arith_dispatch.esk` — the rows native
  and the VM agree on, across all four parity axes.
- `tests/vm_parity/fatal/vector_scalar_arith.esk` — the unguarded form must fail
  closed on both substrates: nonzero exit, a diagnostic, nothing printed past
  the failing form.
- `tests/diagnostics/vector_scalar_arith_crash_le18/` and
  `tests/diagnostics/arith_error_span_outlined_helper/` — exit status, the exact
  runtime `file:line:col`, the crash signature forbidden from returning, and the
  two working `*` sites forbidden from being blamed.

Asserting a RUNTIME span needed a new capability: the golden-corpus gate
discarded the run's stderr and could only pin COMPILE-time spans, so
`scripts/check_diagnostic_corpus.py` gains `runtime_diagnostics` (contains +
span, matched against the runtime `file:line:col: <message>` format) and
`run_stderr_forbidden`, with six red/green self-test fixtures — including this
exact drift shape, and one that rejects a runtime assertion on a case that never
runs. This continues ADR-0010 gap A11 (diagnostic quality), which previously
covered compile-time diagnostics only.

Two new LOUD-ERROR ledger entries (LE-18 crash, LE-19 structural), both filed
`closed` with measured before/after values; no new SILENT-WRONG entries.

## Verification on this branch

- `scripts/run_all_tests.sh` — 46/46 suites, 849/849 tests, 100%
- `scripts/run_vm_parity.sh` — 243 passed, 0 failed (audit + corpus + oos + fatal)
- `scripts/check_diagnostic_corpus.py` — 17/17 cases; `--self-test` PASS
- `scripts/check_ledger_integrity.py` — PASS, 143 entries
- `gen_language_surface.py --check` and `gen_api_docs.py --check` — clean

## Docs

`CHANGELOG.md` (Unreleased/Fixed), `docs/KNOWN_ISSUES.md` (Resolved in
v1.3.5-evolve), `docs/COMPLETE_LANGUAGE_SPECIFICATION.md` 4.1 (the element-wise
contract is binary, in both operand orders) and
`docs/reference/tensors/operations.md` (the same rule for the generic
operators). The latter also corrects a stale line claiming
`(tensor-pow t 2.0)` errors — it broadcasts, as `tensors/INDEX.md` already
said.


tsotchke:

tsotchke:

tsotchke:
Carried onto the v1.3.5 candidate by #677 (merged into #612); the shape-mismatch entry took ledger id LE-22 because the candidate's LE-20 records a different defect.

PR #654, opened 2026-09-10
fix(numeric,ad): exact-tower wave: quoted exact literals, runtime exactness, exact roots and rational tensors, VM exact quoting, nested towers

tsotchke:
Consolidates nine branches' worth of exact-tower work into one CI run. The stack
was #630 and #634 on master, with five fixes stacked above them; each rebase of
that stack cost a full matrix run per branch. This branch is `origin/master`
plus a `--no-ff` merge of each base branch plus a cherry-pick of each stacked
branch's own commits, so the whole group is verified once.

**Consolidates:** #645, #646, #647, #648, #649 — on top of #630 and #634, whose
commits are included here (merged, not rewritten) until those two land.

### What each fix does

- **#634 — `expt` is exact for a rational base and negative exponents (SW-152).**
  `(expt 2/3 -2)` and friends left the exact tower for `pow()`; the exact
  integer-exponent path now covers a rational base and a negative exponent.
- **#634 — `#(...)` literals keep exact elements (SW-153).** Every flat `#(...)`
  literal was parsed as a tensor node, so an exact rational or bignum element
  was force-converted through a double — `#(1/3)` is now a real Scheme vector
  holding the exact `1/3`, while a rectangular nest of plain numeric literals
  keeps its pinned tensor-literal contract.
- **#634 — bignum and bignum-rational literals on the bytecode VM (#640).**
  Read, serialize and print exactly on the VM, matching native.
- **#630 — exact rationals reach the derivative carrier.** An exact rational
  point reaches the carrier, a vanishing tangent keeps the seed's exactness, and
  `derivative`/`derivative-n`/`taylor` nest safely.
- **#630 — comparisons and branch selection inside a differentiand (#644).**
  A comparison inside a differentiand acts on the carrier's primal instead of
  its tagged representation, so a branch selects the same arm it would outside
  the operator.
- **#645 — quoted bignum and bignum-rational literals are the same values as
  evaluated ones (SW-163).** `codegenQuotedAST()` had no case for
  `ESHKOL_BIGNUM_LITERAL` and returned a tagged null, so
  `(car '(123456789012345678901234567890 1))` was `()`, `(exact? '123…)` was
  `#f`, and `(number->string '123…)` printed `"0"`. The `n/d` rational-literal
  desugar also quoted as three-element list data instead of a rational.
- **#646 — exactness is decided by the carrier at run time (SW-159..162).** The
  exact tier ran a static AST whitelist over the differentiand's body *and* its
  point expression and fell back to the always-inexact jet path the instant
  either merely looked unprovable — a composed call a few frames deep, a
  top-level `define`d constant, a point built from `(car …)`/`(vector-ref …)`/a
  loop variable, or a function-call/`(compose …)` differentiand (the last a hard
  compile failure for `taylor`/`derivative-n`). The route is now decided by the
  point's carrier at run time; the static check left is purely structural.
- **#647 — exact square and n-th roots over the exact tower (SW-167).**
  `(sqrt 16)`, `(sqrt 1/4)` and `(expt 4 1/2)` all returned inexact results
  where R7RS 6.2.6 requires the exact one. New shared
  `eshkol_bignum_iroot` computes `floor(a^(1/n))` in exact bignum arithmetic,
  mirrored at the VM's own representation.
- **#647 — tensors never silently zero an exact element (SW-166).**
  `(tensor (vector 1/2 1/3))` stored `#(0 0)` with no diagnostic from one root
  cause, and heap-address garbage from a second, independent one in
  `TensorCodegen::extractAsDouble`. Both closed, on native and the VM. A tensor
  stays f64-backed; an exact element is now correctly *rounded*, never zeroed.
- **#648 — quoted and quasiquoted rational and bignum literals on the VM
  (SW-168).** The VM's own quoting path had the same gap #645 closed natively.
- **#649 — differentiation passes nest at any depth and any order (SW-154).**
  The previous carrier composed one value series with one first-order companion,
  so the matrix had a ceiling: both-passes-order ≥ 2 raised, and two enclosing
  levels at once answered a silent `0`. Every pass now owns a level and a
  foreign level is a *coefficient* of the active one, so depth and per-level
  order are unbounded — and `derivative-n`/`taylor` keep their exactness through
  a nest.

### Conflicts resolved in consolidation

- `.icc/ledger/order.txt` — union of both sides, HEAD's ids first (5 times).
- `.icc/silent-wrong-ledger.yaml` — regenerated from the entries
  (`scripts/gen_silent_wrong_ledger.py`, 5 times); integrity check PASS,
  150 entries.
- `docs/api/INDEX.md`, `docs/api/README.md` — regenerated
  (`scripts/gen_api_docs.py`); `--check` clean.
- `CMakeLists.txt` — additive union of disjoint CTest blocks: #645's
  `quoted_bignum_literals_*` alongside #647's `exact_roots_*` and
  `tensor_exact_elements_*`; #646's `exactness_runtime_property_*` alongside
  #649's `nested_towers_matrix_*`. All six lanes are registered.
- `docs/reference/ad/operators.md`, `docs/guide/AUTOMATIC_DIFFERENTIATION.md` —
  prose written against different premises. #646 removed the static whitelist;
  #649 removed the "an exact seed cannot survive a nest" limitation. Both are
  applied, with the stale halves of each replaced rather than dropped.

The last commit is a separate docs pass for the three claims that were true of
each branch alone and false of the union — the `#(1/3)`/`(tensor 1/3)` note in
the AD operator reference, the nesting limitation in the AD guide, and the two
`(tensor 1/3)` prints `#(0)` citations in the mechanization design note. Each
corrected value was read off the built compiler, not inferred.

### Verification (Release, tests ON, this exact tree)

- `cmake --build` clean; VM prelude cache regenerated — byte-identical, no diff.
- `check_vm_prelude_cache_builtins.py` PASS (733/733 builtins present).
- `gen_api_docs.py --check`, `gen_language_surface.py --check` clean.
- `check_ledger_integrity.py` PASS.
- `ctest -R "exact|rational|taylor|quote|bignum|symbolic|ns_|enclosure|vm_parity"`
  — **47/47 passed**.
- `scripts/run_all_tests.sh` — **849/849 individual tests, 46/46 suites, 0 failures**.
- Every `.esk` test the group adds (22 files) run under JIT, AOT and the
  bytecode VM: all exit 0 on JIT and AOT. Seven exit non-zero on the VM alone,
  each a pre-existing VM feature gap the tests document and for which no VM
  CTest lane is wired — the five AD tests (the AD operators are not implemented
  on the VM), `tensor_exact_elements_test` (`vector->tensor` is not implemented
  on the VM), and `vector_literal_exactness_test` (the VM's reader never
  promoted `#(...)` to a tensor, so its three "is still a tensor" pins are
  native-only by construction). The VM-covered subsets are pinned instead by the
  eight new `tests/vm_parity/corpus/` programs, all of which are byte-identical
  between engines.


tsotchke:

tsotchke:

tsotchke:
Carried onto the v1.3.5 candidate by #675 (merged into #612). SW-154's carrier rewrite is not carried and its ledger entry now reads guarded; it is targeted at v1.4.

PR #671, opened 2026-09-11
fix(codegen): variadic builtins as first-class values keep their arity

tsotchke:
Targets `fix/builtins-are-first-class-values` (#651), which owns the
first-class builtin table this builds on; #651 was still open when this was
opened, so it stacks on that branch rather than on master.

## What was wrong

A reference to a variadic builtin did not answer what the same call in
operator position answers, and said nothing about it. On native, both `-O0`
and `-O2`:

| expression | before | correct |
| --- | --- | --- |
| `(map list (list 1 2 3))` | `((1 . 0) (2 . 0) (3 . 0))` | `((1) (2) (3))` |
| `(map list (list 1 2) (list 3 4))` | `((1 3 . 0) (2 4 . 0))` | `((1 3) (2 4))` |
| `(length (car (map list (list 1 2))))` | raises `cdr: argument is not a pair` | `1` |
| `(map vector (list 1 2) (list 3 4))` | `(#(1) #(2))` | `(#(1 3) #(2 4))` |
| `(apply vector (list 1 2 3))` | `#(1)` | `#(1 2 3)` |
| `(apply string-append (list "a" "b" "c"))` | `8589934593` | `"abc"` |
| `(define f string-append) (f "a" "b" "c")` | `"ab"` | `"abc"` |
| `(define f vector) (f 1 2 3)` | `#(1)` | `#(1 2 3)` |

Walking one of those improper tails printed
`Attempted to get pointer from non-pointer cell (type=1)` per element.
`(map (lambda (x) (list x)) xs)`, `(apply list xs)` and
`(define f list) (f 7)` were unaffected — the three forms that already went
through a rest-list-shaped path, which is what made the defect look like a
`list` quirk rather than a class.

On the VM the mirror-image gap was loud rather than silent: `list`,
`vector` and `string` are matched by head symbol in call position before any
binding lookup, so the bare names had no value at all and `(map list xs)`
died with `undefined variable 'list'`.

## Root cause

Three sites, one class, and each of them a second implementation of a
builtin standing beside the authoritative call-position lowering — the same
failure mode LE-16 closed for the fixed-arity surface.

1. `resolveLambdaFunction` is the route `map` resolves its procedure
   through, and `map` calls that procedure with as many arguments as it has
   lists. It carried a hand-written per-arity factory for `list` that built
   `(cons a0 (cons a1 TERMINATOR))` with the terminator packed by
   `packInt64ToTaggedValue(0)` — the tagged **integer** 0, not nil — so every
   list it produced was improper. Every other variadic builtin had no row at
   that site and fell through to a fixed-arity closure.
2. `codegenVariable`'s value-position route hard-coded exactly two names,
   `list` and `values`, as variadic closures. Every other variadic builtin
   materialised at the table's single fixed arity, and the closure ABI then
   passed only that many arguments.
3. `CallApplyCodegen::applyClosure` spread the argument list into
   `arg_count` separate parameters without ever reading
   `CLOSURE_FLAG_VARIADIC`, so even a correctly flagged variadic closure was
   called with the wrong ABI.

## The fix

The first-class builtin table now says which rows are variadic and, for
each, how the builtin computes its answer from a **rest list** — because
that is the only shape the closure ABI has for "any number of arguments":

* `Identity` — the rest list is the answer (`list`, `values`)
* `RestUnary` — a unary builtin is definitionally `(apply <name> lst)`
  (`vector` is `list->vector`, `string` is `list->string`)
* `LeftFold` — fold the list with the binary form (`string-append`, `min`,
  `max`, `gcd`, `lcm`, `vector-append`, `bytevector-append`)
* `None` — variadic, but a value reference still materialises at the row's
  fixed arity; recorded rather than silently assumed

`resolveLambdaFunction` wraps a variadic row at the **call site's** arity via
`createInlineBuiltinWrapper`, whose body is `codegenCall` itself, and the
hand-written `list` factory is deleted. `codegenVariable` asks the table
instead of naming two builtins. `applyClosure` reads the flag and hands a
zero-fixed-parameter variadic closure its list whole. No builtin acquires a
second implementation: every wrapper body is generated from the
call-position lowering, so a first-class reference cannot disagree with a
direct call by construction.

A latent defect surfaced with it and is fixed here:
`createInlineBuiltinWrapper` keyed its wrapper symbol on the **name** alone,
so the second arity a builtin was wrapped at silently reused the first
arity's function — which LLVM's verifier rejects outright
(`Incorrect number of arguments passed to called function`). The arity is
now part of the symbol.

The VM prelude gains `(define (list . args) args)`,
`(define (vector . args) (list->vector args))` and
`(define (string . chars) (list->string chars))`, giving the bare names the
procedure the call position already implements without touching the
head-symbol fast paths, and `vm_prelude_cache.h` is regenerated.

Non-variadic rows are deliberately untouched at the `map` site: their arity
is a property of the builtin, not of the call site, and wrapping
`(substring s)` because someone mapped `substring` over one list would turn
a wrong answer into a failed compilation.

The zero-argument case of a `LeftFold` row raises an explicit arity error
rather than an invented value: R7RS gives some of these an identity element
(`(string-append)` is `""`, `(gcd)` is `0`) and others none (`(max)` is an
error), and the identity is not derivable from the binary lowering, so
inventing one would be a new second implementation of exactly the kind this
PR removes. Reaching it needs `(apply <variadic-builtin> '())`, which
answered garbage before.

## Evidence

* `tests/core/variadic_builtins_first_class_test.esk` — 28 checks over
  `map` / `apply` / a define-alias / a user higher-order call for `list`,
  `vector`, `string-append`, `append`, `+`, `max`. The `max` cases put the
  largest element **last** so a two-argument wrapper cannot pass by luck.
  0 failures on native JIT, AOT `-O0`, AOT `-O2` and the VM.
* `tests/vm_parity/corpus/80_variadic_builtin_first_class.esk` — twelve of
  the same calls, native and VM output byte-identical (verified by `diff`).
* New ctest lanes: `variadic_builtins_first_class_runtime_smoke`,
  `variadic_builtins_first_class_aot_o0_smoke`,
  `variadic_builtins_first_class_vm_smoke`,
  `variadic_builtin_first_class_vm_parity`.
* `ctest -R "builtin|first_class|apply|map|vm_parity|prelude|variadic"` —
  65/65 pass, including `vm_prelude_cache_is_current`.
* `scripts/run_all_tests.sh` — all suites pass, 100%.
* `scripts/run_value_position_sweep.py` — 637 comparisons matching call
  position, 0 new findings, baseline unchanged.
* `python3 scripts/check_wasm_imports.py --build-dir build` — OK.

One pre-existing test was corrected rather than accommodated: the generated
first-class audit's `string` row called `(string "abc")`, passing a string
where R7RS `string` takes characters, and pinned the call position's garbage
against the value position's garbage. Its fixture argument is now a
character, and the regenerated chunk is a one-line diff.

Ledger: `SW-173` (bucket SILENT-WRONG, status fixed), aggregate regenerated,
`check_ledger_integrity.py` and `gate_no_silent_wrong.py` both PASS.


tsotchke:
Superseded: carried onto the v1.3.5 integrated candidate by #674 (merged into #612), where the candidate's closure dispatcher already owns the variadic rest-list decision. Closing; reopened if #612 is abandoned.

PR #694, opened 2026-09-15
Fix function argument order in list operations

LJGz:
Why: a reader who types the tutorial examples into the native read-eval-print loop gets a type error. The functions were fine; the examples were not.

Verified against:
- lib/core/list/transform.esk  (define (take lst n) ...)
- lib/core/list/sort.esk       (define (sort lst less?) ...)
- docs/ESHKOL_QUICK_REFERENCE.md

Not fixed here: LE-04 (native vs bytecode virtual machine sort order). Docs now describe the native/source contract.


tsotchke:
Thank you. Verified by running both forms: the corrected examples print the documented results, and the previous ones raise a type error. Merged; this ships in v1.3.5-evolve, and a standing check that executes every tutorial example is going in alongside it.

PR #698, opened 2026-09-17
fix(model-io): preserve scalar and empty ESKM tensors across engines

Gabriel-Kahen:
ESKM v1 already represents rank-zero scalars and zero-element tensors, but the VM loader refused both historical fixtures. Native and VM observation guards also rejected scalar metadata. This change makes loading, inspecting and rewriting those values preserve their original shape and exact checkpoint bytes on native JIT/AOT and VM source/bytecode.

The VM uses an I/O-local adapter with owned storage, checked strides and the existing region lifetime machinery. A private shared guard permits scalar observation without changing general arithmetic or destination validation. Whole-file preflight, the v1 wire format, public signatures and the general tensor constructors remain unchanged. The handoff and contributor roadmap now distinguish landed work from outstanding acceptance, and the v2 proposal remains Proposed.

Scope: VM checkpoint admission/materialization; native/VM scalar observation; existing C and four-engine compatibility gates; serialization documentation; regeneration of the required browser VM bundle and its size statistic. This implements the D1–D5 conventions from #614 for review; it does not claim prior maintainer acceptance.

Validation on Linux x86-64, fresh Release build, LLVM 21.1.8 / Clang 22.1.6, based on upstream `0901264c` (optional BLAS/GPU/quantum/agent FFI disabled):

- PASS: all six immutable valid fixtures across 4 producers × 4 consumers: 96 exact-byte pairs, plus the existing constructed-model matrix.
- PASS: 20 public tensor cases, 16 function/region lifetime cases, and 16 negative controls. Wrong scalar metadata/payload expectations fail on all four engines.
- PASS: 8/8 focused CTests: model_io_test, vm_model_io_fail_closed_test, atomic_checkpoint_file_test, tensor_operand_classification_runtime_smoke, generated_artifact_validator_test, vm_prelude_cache_is_current, eskm_tensor_engine_parity and eskm_tensor_oracle_self_test.
- PASS: regenerated browser bundle with Emscripten 4.0.22; 9/9 existing REPL transcript cases, 3/3 math/AD/tensor smoke cases, and site-release verifier. Native `tensor-length` assertions are restored and both fixture-matrix runs still pass.
- PASS: fixture integrity and its payload-drift self-test; generated API docs; public API documentation; coverage inventory; shell syntax; git diff --check.
- Regression evidence: the new positive producer test fails on the VM translation unit built from unchanged `0901264c` at scalar and empty model loading (exit 1), and passes on the changed VM. The baseline VM links the same native runtime archive; it is not a separately rebuilt full historical toolchain.

Primary commands:

```sh
bash scripts/run_eskm_v1_model_load_parity.sh build/eshkol-run build/eshkol-vm-standalone-test
bash scripts/run_eskm_v1_model_load_parity.sh --self-test build/eshkol-run build/eshkol-vm-standalone-test
ctest --test-dir build --output-on-failure -j1 -R '^(model_io_test|vm_model_io_fail_closed_test|atomic_checkpoint_file_test|tensor_operand_classification_runtime_smoke|generated_artifact_validator_test|vm_prelude_cache_is_current|eskm_tensor_engine_parity|eskm_tensor_oracle_self_test)$'
python3 scripts/gen_api_docs.py --check
python3 scripts/check_public_api_docs.py --no-trace
python3 scripts/check_test_coverage.py
```

Limits: browser checkpoint I/O reaches the existing documented `eshkol_capability_require` aborting stub, so no WASM checkpoint-parity claim is made. Rank-zero transfer across VM worker heaps remains unsupported by the existing worker-copy path. Neither limit is expanded by this PR.

Not run: macOS/Windows, the full repository suite, new sanitizer/resource campaigns, or GitHub CI at this head. V2 implementation, stronger save durability, and general scalar tensor algebra are separate work.


tsotchke:
Thank you, Gabriel. This work is included in the v1.3.5-evolve release: every file from this pull request is on the release integration branch, and the ESKM checkpoint tests pass there on the native and VM engines. It reaches master through the release pull request #628, so this pull request is closed as included rather than merged separately.

PR #699, opened 2026-09-17
feat(model-io): stage internal ESKM v2 preflight validation

Gabriel-Kahen:
The proposed ESKM v2 envelope now has an executable, private validation boundary. Given an immutable buffer, explicit limits, and caller workspace, the C17 validator checks the entire container and returns validated ranges/counts or a deterministic status and byte offset. Failure clears the result; no tensor construction, I/O, callbacks, recursion, or heap allocation occurs.

The parser validates the fixed header and CRC, optional TLV framing, opaque annotations with unique keys, and every v1-shaped tensor record. Ordered unsigned shape multiplication preserves scalar and empty-shape wire semantics. Duplicate detection uses iterative heapsort in a size-asserted 16 KiB workspace. Callers can lower the provisional limits but cannot raise them; zero means zero.

The parser is compiled into private tests only. Public native/VM I/O, browser artifacts, and the immutable v1 corpus remain unchanged. The format decision stays **Proposed**: this prepares the wire-validation portion of GK-SER-05a and does not complete that milestone or approve public integration. Backend admission, allocation accounting, and transactional materialization remain gates for reader integration. The experimental reference records this distinction explicitly.

Validation adds three independently constructed exact-byte fixtures with metadata, raw binary64 bits, CRCs, hashes, and provenance; 504 C assertions; a C++ consumer; and negative controls for wrong metadata/status/offset and CRC-correct metadata drift. The standalone runner needs no LLVM build and audits parser object imports. CMake and focused Linux/macOS CI entries require all three test gates individually.

### Validation

Base: `0901264ccef7bfad307a1965188bf7658e68cc32`  
Tested head: `293d89523e5847c74f1ef2d74d115d645e2e0151`

On x86-64 Linux:

- **PASS** `CC=gcc CXX=g++ scripts/run_eskm_v2_preflight_tests.sh` — GCC 16.1.1, 504 checks, C++ consumer, three goldens and negative controls, allocation-import audit.
- **PASS** `CC=clang CXX=clang++ ESKM_V2_SANITIZE=1 scripts/run_eskm_v2_preflight_tests.sh` — Clang 22.1.6 with AddressSanitizer/UndefinedBehaviorSanitizer; same gates.
- **PASS** all eight focused CTests: `eskm_v2_preflight_test`, `eskm_v2_preflight_cpp_test`, `eskm_v2_fixture_oracle_test`, `model_io_test`, `vm_model_io_fail_closed_test`, `tensor_validation_test`, `eskm_tensor_engine_parity`, `eskm_tensor_oracle_self_test`. Release build used Clang 22.1.6 / LLVM 21.1.8, with BLAS, GPU, quantum and agent FFI disabled.
- **PASS** `python3 scripts/check_eskm_v1_fixtures.py` and `--self-test`; `python3 scripts/check_public_api_docs.py --no-trace`; `python3 scripts/check_test_coverage.py`; `make api-docs-check`; `git diff --check`.
- **PASS** independent parser and integration/test reviews; strict GCC/Clang compilation and Clang static analysis. `python3 -O scripts/check_eskm_v2_fixtures.py` correctly refuses to skip assertions.

The workspace is exactly 16,384 bytes on supported builds. Clang `-O2 -fstack-usage` reports a longest visible parser frame chain of 328 bytes (120 + 104 + 104) on this host, excluding library/ABI overhead; that measurement is not a portable stack guarantee. Input storage belongs to the caller, so the 256 MiB buffer ceiling is not a total-process memory bound.

macOS, Windows and 32-bit execution were not run locally. No public v2 producer/consumer matrix is claimed; there are no public v2 entry points in this PR. This branch starts directly from master and is independent of #698.


tsotchke:
Thank you, Gabriel. This work is included in the v1.3.5-evolve release: every file from this pull request is on the release integration branch, and the ESKM checkpoint tests pass there on the native and VM engines. It reaches master through the release pull request #628, so this pull request is closed as included rather than merged separately.

